PR #8, opened 2015-01-21
Controller for 4 channel DDS used in several ion storage labs. 

jboulder:


fallen:
In general, to fix PEP8 issues I recommend scanning your python files using the following command:
```python3 -m flake8 your_file.py```
It will show all PEP8 issues.

**Comment by jboulder** [comment]
Thank you for the comments Yan. I've pushed changes to jboulder/artiq. The only one I don't understand is the first. 

Do you really need to assign the classname like this?
Cannot you do something like obj.__class__.__name__ if obj is an instance of class Novatech409B?

Also, please make changes as required to integrate error/status reporting. I don't yet know about this infrastructure.

**Comment by jboulder** [comment]
Also, I found that Spyder has this built in...


   - use a text editor that enforces PEP8
      - Spyder :: Preferences :: Editor :: Code Introspection ::
         - check pyflakes
         - check pep8

**Comment by fallen** [comment]
about the first comment, I had troubles due to the underscore meaning "underline" with markdown markup language.
I meant: ```obj.__class__.__name__``` to retrieve the class name of an instance.

**Comment by fallen** [comment]
I quite recently started using Pycharm which also warns about PEP8

**Comment by jboulder** [comment]
Thank you for the many suggestions. I believe I've included them all. Please pull.

File: artiq/devices/novatech409B/novatech409B.py [hunk 1]
```diff
@@ -0,0 +1,362 @@
+#!/usr/bin/env python3
```
jordens: delete

artiq/devices/novatech409B/novatech409B.py
```diff
@@ -0,0 +1,362 @@
...
+class Novatech409B:
+    """controller for Novatech 409B 4-channel DDS
+
+    This class is an interface with the Novatech Model 409B 4-
+    channel DDS box. The interface is a serial interface.
+
+    :param int comport: COM port number on Windows
+    :param int debug: debug level
+    """
+    # NOTE: the in-line documentation is in Sphinx format
```
jordens: delete

artiq/devices/novatech409B/novatech409B.py
```diff
@@ -0,0 +1,362 @@
...
+    # http://sphinx-doc.org/domains.html
+    def __init__(self, comport=1, debug=1, simulate_hw=False):
+        # some private members
+        self.__comport = comport
+        self.__debug = debug
+        self.__className = "Novatech409B"
+        self.simulate_hw = simulate_hw  # true if disconnected from hw
+        self.serial_rw_delay = 0.001  # is time between reads and writes
+
+        # setup logging
```
jordens: do a simplified version of it in main() or -- even better -- elevate to generic for alll drivers

artiq/devices/novatech409B/novatech409B.py
```diff
@@ -0,0 +1,362 @@
...
+    channel DDS box. The interface is a serial interface.
+
+    :param int comport: COM port number on Windows
+    :param int debug: debug level
+    """
+    # NOTE: the in-line documentation is in Sphinx format
+    # http://sphinx-doc.org/domains.html
+    def __init__(self, comport=1, debug=1, simulate_hw=False):
+        # some private members
+        self.__comport = comport
```
jordens: useless

artiq/devices/novatech409B/novatech409B.py
```diff
@@ -0,0 +1,362 @@
...
+        self.__debug = debug
```
jordens: use logging

artiq/devices/novatech409B/novatech409B.py
```diff
@@ -0,0 +1,362 @@
...
+        self.__className = "Novatech409B"
```
jordens: delete as yann requested

artiq/devices/novatech409B/novatech409B.py
```diff
@@ -0,0 +1,362 @@
...
+                xonxoff=0,
+                timeout=0.05)
+        self.setup()
+
+    def __del__(self):
+        if not self.simulate_hw:
+            self.__ser.close()
+            time.sleep(1)
+
+    def echo(self, s):
```
jordens: if this is really needed, elevate to all drivers

artiq/devices/novatech409B/novatech409B.py
```diff
@@ -0,0 +1,362 @@
...
+                serial_port_id,
+                baudrate=19200,
+                bytesize=8,
+                parity="N",
+                stopbits=1,
+                xonxoff=0,
+                timeout=0.05)
+        self.setup()
+
+    def __del__(self):
```
jordens: this is broken. __ser may already be deleted. what is the sleep for?

artiq/devices/novatech409B/novatech409B.py
```diff
@@ -0,0 +1,362 @@
...
+                # note that pySerial starts counting serial ports at 0
+                serial_port_id = int(self.__comport)-1
+            elif self.__platform is "Linux":
+                # just use the device string as
+                # passed (e.g. "/dev/ttyUSB0")
+                serial_port_id = "/dev/ttyUSB0"
+            else:
+                self.debug_message("__init__", "unknown platform", level=0)
+                sys.exit()
+            self.__ser = serial.Serial(
```
jordens: remove the options that just repeat a default

artiq/devices/novatech409B/novatech409B.py
```diff
@@ -0,0 +1,362 @@
...
+
+        # setup logging
+        FORMAT = "%(asctime)-15s %(message)s"
+        logging.basicConfig(format=FORMAT)
+        self.logger = logging.getLogger("artiq.driver.novatech409B")
+        self.logger.setLevel(logging.DEBUG)
+        self.debug_message("__init__", "", level=3)
+
+        # establish serial connection --- platform dependent
+        if not self.simulate_hw:
```
jordens: please factor out the simulation code into its own class (see lda)

artiq/devices/novatech409B/novatech409B.py
```diff
@@ -0,0 +1,362 @@
...
+            self.__ser.close()
+            time.sleep(1)
+
+    def echo(self, s):
+        ss = "novatech409B.echo() :: " + s
+        self.debug_message("echo", ss)
+        return ss
+
+    def debug_message(self, func_name, msg, level=2):
```
jordens: delete

artiq/devices/novatech409B/novatech409B.py
```diff
@@ -0,0 +1,362 @@
...
+            except serial.SerialException as e:
+                self.debug_message("ser_send", e, level=0)
+
+            # check for error from device
+            # expected response (no error) is myStr\r\nOK\r\n
+            # after convert to python3 need to specify type of return
+            # to be bytes
+            if ~ignore_unusual_response:
+                if result != expected_response:
+                    print("ERROR :: novatech409B.ser_send() "
```
jordens: remove all print() s, use logging

artiq/devices/novatech409B/novatech409B.py
```diff
@@ -0,0 +1,362 @@
...
+                        "response was {}".format(result))
+                    return (result, expected_response)
+                    sys.exit()
```
jordens: raise an exception if this is really needed. why would you want to exit the program here?

artiq/devices/novatech409B/novatech409B.py
```diff
@@ -0,0 +1,362 @@
...
+        important to change all the outputs to new values
+        simultaneously."""
+        self.debug_message("set_simultaneous_update", "", level=2)
+        if my_bool:
+            self.ser_send("I m")
+        else:
+            self.ser_send("I a")
+
+    def set_freq(self, ch_no, freq):
+        """set_freq(ch_no,freq):
```
jordens: this docstring does not add anything.
use units (see lda).

artiq/devices/novatech409B/novatech409B.py
```diff
@@ -0,0 +1,362 @@
...
+
+        :param int channel_num: 0 to 3
+        :param float phase: phase angle in cycles [0,1]
+        :returns: None
+        """
+        self.debug_message("set_phase",
+                          str(channel_num) + "," + str(phase),
+                          level=3)
+        if channel_num < 0 or channel_num > 3:
+            print("ERROR :: novatech409B.set_phase() channel_num Error")
```
jordens: why not exit the program here like set_freq()?

artiq/devices/novatech409B/novatech409B.py
```diff
@@ -0,0 +1,362 @@
...
+        """ sweep phase of all DDSs, phase continuous
+
+        Sweep frequency in a phase continuous fashion.
+
+        :param float f0: starting frequency (MHz)
+        :param float f1: ending frequency (MHz)
+        :param float t: sweep duration (seconds)
+        :returns: None
+        """
+        s = str(f0) + "," + str(f1) + "," + str(t)
```
jordens: "".format()

artiq/devices/novatech409B/novatech409B.py
```diff
@@ -0,0 +1,362 @@
...
+
+        :param float frac: 0 to 1 (full attenuation to no attenuation)
+        """
+        self.debug_message("scaleOutput", str(frac), level=2)
+        for ch_no in range(4):
+            self.output_scale(ch_no, frac)
+        # send command necessary to update all channels at the same time
+        self.ser_send("I p")
+
+    def output_on_off(self, ch_no, my_bool):
```
jordens: rename my_bool to something meaningful

doc/manual/drivers_reference.rst
```diff
@@ -37,6 +42,8 @@ When writing a new driver, choose a free TCP port and add it to this list.
 +--------------------------+--------------+
 | Master (control)         | 3251         |
 +--------------------------+--------------+
+| Novatech 409B            | 3254         |
```
jordens: sort by port or alphabetically

artiq/devices/novatech409B/novatech409B_client.py
```diff
@@ -0,0 +1,131 @@
...
+                                          help="save to EEPROM")
+    parser_eeprom.add_argument("save-to-eeprom",
+                        action="store_true",
+                        help="saves current state into EEPROM "
+                        "and sets valid flag; state used as default"
+                        " upon next power up or reset")
+    return parser
+
+
+def _get_args():
```
jordens: inline this, if you want to be really concise, say `args = define_parser().parse_args()`

artiq/devices/novatech409B/novatech409B.py
```diff
@@ -0,0 +1,362 @@
...
+
+    def output_scale(self, ch_no, frac):
+        """changes amplitude of a DDS
+
+        :param int ch_no: DDS channel 0, 1, 2 or 3
+        :param float frac: 0 to 1 (full attenuation to no attenuation)
+        :returns: None
+        """
+        self.set_simultaneous_update(False)
+        dac_ch_no = int(math.floor(frac*1024))
```
jordens: `dac_ch_no`?

artiq/devices/novatech409B/novatech409B.py
```diff
@@ -0,0 +1,362 @@
...
+        else:
+            #turn off output
+            self.output_scale(ch_no, 0.0)
+
+    def output_on_off_all(self, my_bool):
+        """:param bool my_bool: is True (on) or False (off)
+
+        turns on or off the all the DDSs
+        """
+        self.debug_message("output_on_off_all", "", level=2)
```
jordens: would you not expect `self.set_simultaneous_update(True)` here?

artiq/devices/novatech409B/novatech409B.py
```diff
@@ -0,0 +1,362 @@
...
+            self.ser_send("I a")
+
+    def set_freq(self, ch_no, freq):
+        """set_freq(ch_no,freq):
+        Set ch_no to frequency freq MHz"""
+        self.debug_message("set_freq", str(ch_no)+","+str(freq), level=3)
+        if ch_no < 0 or ch_no > 3:
+            print("ERROR :: novatech409B.set_freq() ch_no Error")
+            sys.exit()
+        if freq < 0.0 or freq > 171.1276031:
```
jordens: magic number?

artiq/devices/novatech409B/novatech409B.py
```diff
@@ -0,0 +1,362 @@
...
+        self.ser_send("R", ignore_unusual_response=True)
+        time.sleep(1)
+        self.setup()
+
+    def setup(self):
+        """initial setup of 409B
+
+        Setup the Novatech 409B with the following defaults.
+        * command echo off ("E d")
+        * external clock ("") 10 MHz sinusoid -1 to +7 dBm
```
jordens: is that done here?

artiq/devices/novatech409B/novatech409B.py
```diff
@@ -0,0 +1,362 @@
...
+                time.sleep(self.serial_rw_delay)
+                result = self.__ser.read(1028)
+            except serial.SerialException as e:
+                self.debug_message("ser_send", e, level=0)
+
+            # check for error from device
+            # expected response (no error) is myStr\r\nOK\r\n
+            # after convert to python3 need to specify type of return
+            # to be bytes
+            if ~ignore_unusual_response:
```
jordens: use `not`

artiq/devices/novatech409B/novatech409B.py
```diff
@@ -0,0 +1,362 @@
...
+        :param str myStr: a character string to send to device
+        :returns: None
+        """
+        self.debug_message("ser_send", myStr, level=3)
+        s = myStr + "\r\n"
+        expected_response = b"OK\r\n"
+        result = b""
+        if self.simulate_hw is False:
+            try:
+                self.__ser.flush()
```
jordens: why? i doubt this is useful or needed. would you not want to flush after `write()`?

jordens:
* i would advocate renaming `novatech409B.py` to `driver.py` as the directory already has the name. also make entry_points for the `main()`s and optionally rename+move their files to `artiq.frontend` (like lda).

* it would be much less code if you deleted all the debugging that just holds the function name. you can do that in the rpc client before calling the decoded method (if it is not already in there).

artiq/devices/novatech409B/novatech409B_client.py
sbourdeauducq: ``define_parser`` is called ``get_argparser`` elsewhere in ARTIQ. Having it as a separate function enables sphinx to extract the argument info.

jordens: that's fine. but my comment is about inlliniing `_get_args()`, not `define_parser()`.

doc/manual/drivers_reference.rst
```diff
@@ -24,6 +24,11 @@ Each driver is run in a separate "controller" that exposes a RPC interface (base
 .. automodule:: artiq.devices.lda.driver
     :members:
 
+:mod:`artiq.devices.novatech409B` module
+---------------------------------
+
+.. automodule:: artiq.devices.novatech409B.novatech409B
+    :members:
```
jordens: rebase and use `sphinx-argparse`.

artiq/devices/novatech409B/novatech409B.py
jboulder: What is "it" here? I don't know what you mean by "elevate to generic".

artiq/devices/novatech409B/novatech409B.py
jboulder: I've fixed this bug. self.__ser is now only deleted in the destructor if it was successfully created.

artiq/devices/novatech409B/novatech409B.py
jboulder: Simulation here means that there is no attempt at serial communication. Creation of a new class is a good idea for something complex but here I think it would be cumbersome.

artiq/devices/novatech409B/novatech409B.py
jboulder: I'd like to keep this for debugging purposes.

artiq/devices/novatech409B/novatech409B.py
jboulder: I didn't know artiq had a unit checking system. Cool. :) But I don't understand how they're to be used. Is the idea that all quantities with units that are passed around in the artiq ecosystem (e.g. to/from parameter database) are Quantities? Please add to the documentation a writeup and brief use case. I'll gladly modify my client and driver to use them.

artiq/devices/novatech409B/novatech409B_client.py
jboulder: I'm happy with this. This user interface doesn't need to be optimized for speed. At least that's why I guess you may suggest inlining.

artiq/devices/novatech409B/novatech409B.py
jboulder: Initially, the device is configured to be reconfigured immediately upon receiving a freq, phase or amplitude update instruction. This is the expected behavior of a device. The alternate mode is when several freq, phase or amplitude changes are made sequentially but applied in parallel.

artiq/devices/novatech409B/novatech409B.py
jboulder: This is the maximum frequency of the DDS. 

It would be nice if the artiq units class supported range checking.

jboulder:
Robert suggested: 
>i would advocate renaming novatech409B.py to driver.py as the directory 
>already has the name. 
I tried this but encountered a problem with package name resolution. How do client.py and controller.py know which of the many driver.py's to import? And, it makes it confusing in an editor window when there are a bunch of identically named driver.py files. I've encountered this trying to compare lda, novatech409B and pdq2. 

>also make entry_points for the main()s and optionally 
>rename+move their files to artiq.frontend (like lda).
I'm of the understanding that artiq's interaction with drivers will be via RPC. What is the benefit of moving things to artiq.frontend? 

>it would be much less code if you deleted all the debugging that just 
>holds the function name. you can do that in the rpc client before calling 
>the decoded method (if it is not already in there).

Agreed. I'll fix this when adding error logging.

artiq/devices/novatech409B/novatech409B.py
jordens: Seems like all you need to do is overload `ser_send()` or even better, warp `__ser.write()` in your class and then overload the wrapper.

artiq/devices/novatech409B/novatech409B.py
jordens: `debug_message()`? you can do all that functionality (and more) much better by setting up logging.

artiq/devices/novatech409B/novatech409B.py
jordens: The problem is that several methods do call `set_simultaneous_update()` as a hidden side effect. It is much better to just expose the functionality of the device as faithfully as possible without adding side effects. Then, at a higher level -- e.g. in the experiment code -- you can abstract to compound methods.

More specifically, if I understand this correctly, the correct terminology is transactions. You are in fact opening a transaction with `I m` and committing that transaction with `I p`. And `I a` is implicitly opening a transaction and committing it on every command. You can expose that functionality much more idiomatically.

artiq/devices/novatech409B/novatech409B.py
jordens: You should write the frequency so that it is obvious how it comes about.
`f0 = 5e6*2**n/5**m` or thereabouts.
Does the device check the range? Does it depend on the external clock settings? If yes, you can simply delete all that code.

artiq/devices/novatech409B/novatech409B.py
jordens: A line comment here refers to the line just above the comment, `# setup logging` in this case. What Yann wrote.

artiq/devices/novatech409B/novatech409B.py
jordens: https://docs.python.org/3/reference/datamodel.html#object.__del__
You don't need `__del__` here. If the instance is collected, its attributes collected if they are not otherwise referenced. `__ser.close()` will be called anyway. What problem were you trying to solve?

doc/manual/writing_a_driver.rst
```diff
@@ -106,6 +113,34 @@ The controller's code would contain something similar to this: ::
...
         simple_server_loop(Hello(), args.bind, args.port)
 
+The driver's server should be saved in ~/artiq-dev/artiq/artiq/devices/driver_name/server.py.
+
+__init__.py
+-----------
+This file must exist for porper integration with ARTIQ. It can be empty.
+~/artiq-dev/artiq/artiq/devices/driver_name/__init__.py
+
+Device Manager Udev Rules
```
jordens: For the serial ports `adduser user dialout` and `serial.serial_for_url()` is much simpler, more powerful, and also works under windows.

artiq/devices/novatech409B/novatech409B.py
sbourdeauducq: I'm actually considering dumbing it down since it doesn't play nice with hdf5, numpy, json, etc.

doc/manual/writing_a_driver.rst
```diff
@@ -106,6 +113,34 @@ The controller's code would contain something similar to this: ::
         args = _get_args()
         simple_server_loop(Hello(), args.bind, args.port)
 
+The driver's server should be saved in ~/artiq-dev/artiq/artiq/devices/driver_name/server.py.
+
+__init__.py
+-----------
+This file must exist for porper integration with ARTIQ. It can be empty.
```
sbourdeauducq: proper

doc/manual/writing_a_driver.rst
```diff
@@ -106,6 +113,34 @@ The controller's code would contain something similar to this: ::
...
+~/artiq-dev/artiq/artiq/devices/driver_name/__init__.py
+
+Device Manager Udev Rules
+-------------------------
+On Linux systems udev is the device manager. It manages device nodes in /dev. ARTIQ drivers
+often interface with instruments that have USB-Serial interfaces. Normally these devices
+are assigned device nodes on an ad hock basis by the kernel (e.g. /dev/ttyUSB3); the
```
sbourdeauducq: ad hoc

artiq/devices/novatech409B/dmesg.txt
```diff
@@ -0,0 +1,9 @@
+What is reported when the Novatech409B is connected to a USB port on Xubuntu?
```
sbourdeauducq: delete or make part of the documentation (not the source).

artiq/devices/novatech409B/controller.py
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/python3
+
+# Copyright (c) 2014 Joe Britton, Sebastien Bourdeauducq
+
+import argparse
+from artiq.protocols.pc_rpc import simple_server_loop
+import importlib
+import novatech409B
+importlib.reload(novatech409B)
```
sbourdeauducq: Why?

jboulder:
Please take a look at these proposed changes.

doc/manual/writing_a_driver.rst
jboulder: If we all agree on including a template for driver development at artiq/devices/example_artiq_device please add an example serial interface using serial_for_url().

artiq/devices/novatech409B/novatech409B.py
jboulder: Do you plan to use Quantities for passing data to driver and the parameter database?

artiq/devices/novatech409B/novatech409B.py
jboulder: I agree Robert. I'll add this to my todo list.

artiq/devices/novatech409B/novatech409B.py
jboulder: The device does not check the range. The number of low-level details which can influence this number are too numerous given my aspiration for this driver. If the device is configured to used it's internal reference servo'd to an external reference via a PLL this is the appropriate constant. I'll let future users worry about other use cases.

doc/manual/writing_a_driver.rst
jboulder: Robert, I've added __setup_serial(self, serial_port) 

Please add to the example how to use serial_for_url() as I've not used it before.

artiq/devices/novatech409B/controller.py
jboulder: When using an interactive python session reload() is necessary to force reloading of python modules that may have been edited. Python isn't smart enough to automatically reload modified modules.

jboulder:
What else needs to be done to finalize this pull request? It looks like this branch is starting to diverge from the m-labs root. e.g. writing_a_driver.rst

jordens:
Yes. It is becoming unwieldy.

* `rebase --interactive` squashing all the changes into functional patches
* split into branches (example, documentation, novatech)
* do pull requests for each

artiq/devices/novatech409B/novatech409B.py
jordens: Then you the title "used in several ion storage labs" is overly ambitious.

If you can not determine this magic number at runtime, make it a class attribute so that others can just subclass and override.

doc/manual/writing_a_driver.rst
jordens: You should.

artiq/devices/novatech409B/controller.py
jordens: Not due to a lack of smartness. Please remove the reload().

artiq/devices/novatech409B/novatech409B.py
jboulder: Good suggestion.

sbourdeauducq:
Novatech fixed and merged.

Please send updated files/patch for the generic driver and doc edits.

PR #10, opened 2015-02-04
fix __init__.py in pdq2 and lda

jboulder:


sbourdeauducq:
Experiments never use the lda driver directly (instead they use a controller entry in the DDB) so exporting the driver at the top level does not improve the public API.

I'm planning to export in devices/*/__init__.py only the parts of drivers that need to be executed locally and used directly by the experiment; for example, the high-level PDQ or PXI DAC card APIs that need to define kernels that generate timing pulses.

jordens:
Apart from the fact that the stack of commits is messed up:

* `pdq2`: Putting actual code (not just namespace mangling) into `__init__.py` is irritating because then you never know in which file to look for a class, especially if there are other files next to `__init__.py`. If there are no other modules in the same directoty, rename `module/__init__.py` to `module.py` and delete the directory.
* Cleaning up and flattening the API is a good idea whether public or not.

PR #13, opened 2015-03-21
Improve installation instructions

whitequark:


doc/manual/installing.rst
```diff
@@ -93,17 +93,19 @@ These steps are required to generate bitstream (``.bit``) files, build the MiSoC
         $ git clone --recursive https://github.com/m-labs/misoc
         $ export MSCDIR=~/artiq-dev/misoc # append this line to .bashrc
 
-* Build and flash the bitstream and BIOS by running `from the MiSoC top-level directory` ::
+* Download and install ARTIQ: ::
+
+        $ cd ~/artiq-dev
+        $ git clone https://github.com/m-labs/artiq
+        $ python3 setup.py develop --user
```
whitequark: The next step will fail unless ARTIQ is installed, so move it earlier.

doc/manual/installing.rst
```diff
@@ -133,15 +135,15 @@ Installing the host-side software
         $ cd ~/artiq-dev/llvm-or1k
         $ mkdir build
         $ cd ~/artiq-dev/llvm-or1k/build
-        $ ../configure --prefix=/usr/local/llvm-or1k
-        $ make ENABLE_OPTIMIZED=1 -j4
-        $ sudo -E make install ENABLE_OPTIMIZED=1
+        $ cmake .. -DCMAKE_INSTALL_PREFIX=/usr/local/llvm-or1k -DLLVM_TARGETS_TO_BUILD=OR1K -DCMAKE_BUILD_TYPE=Debug -DBUILD_SHARED_LIBS=ON
```
whitequark: * autoconf is deprecated in LLVM. 
 * why build all targets (the default)? this just wastes time.
 * debug builds finish much quicker for an insignificant (for our purposes) performance loss. also, debug symbols help when you get an assert.
 * shared libs make the install (and any installed dependency as well) take almost an order of magnitude less space. llvm uses rpath so no LD_LIBRARY_PATH alterations are necessary. you also don't have to rebuild llvmlite if you update llvm.

doc/manual/installing.rst
```diff
@@ -179,8 +181,8 @@ Installing the host-side software
         $ cd ~/artiq-dev/artiq/doc/manual
         $ make html
 
-Xubuntu 14.04 specific instructions
------------------------------------
+Ubuntu 14.04 specific instructions
```
whitequark: the package sources are mostly same in ubuntu and xubuntu, so the instructions apply to either.

doc/manual/installing.rst
fallen: Which step is failing if ARTIQ is not installed?

doc/manual/installing.rst
fallen: I think the conda package right now takes for granted that llvmlite is linked statically with llvm.
We can use shared library, but we would need to do a conda package for llvm then and ship it as an llvmlite dependency.

fallen: or maybe we just keep the llvmlite conda package statically linked ...

doc/manual/installing.rst
whitequark: `./make.py -X ~/artiq-dev/artiq/soc -t artiq_ppro all` ← this imports some modules from artiq and fails

doc/manual/installing.rst
whitequark: i think llvmlite package should depend on llvm either way, so that updates to our llvm fork would propagate to package consumers. it doesn't matter which way we build it; if it's static, llvmlite should be also rebuilt, if shared, then no

PR #319, opened 2016-03-06
Problem: ARTIQ setup via conda or src is complex

sjmackenzie:
Solution: Provide a single command to download, compile and install
dependencies.

### DIFF
--- a/nix/README.rst
+++ b/nix/README.rst
@@ -1,0 +1,22 @@
+Install ARTIQ via the Nix Package Manager
+===========================
+
+Nix does not support windows.
+
+* Install the nix package manager
+
+  * many linux distros already have a package for the `nix package manager <http://nixos.org/nix/>`_
+
+    * for example: $ apt-get install nix
+
+  * if you would like to install via sh (please be sure you `understand <https://www.seancassidy.me/dont-pipe-to-your-shell.html>`_ the dangers involved when curl piping to sh. Also ensure you have read the contents of the script and feel comfortable executing it. Otherwise there is the `manual <http://nixos.org/nix/manual/>`_)
+
+    * $ curl https://nixos.org/nix/install | sh
+
+    * $ source ~/.nix-profile/etc/profile.d/nix.sh
+
+* $ git clone github.com/m-labs/artiq
+* $ cd artiq/nix
+* $ nix-env -i -f default.nix
+
+The above command will setup your entire environment.
--- a/nix/artiq.nix
+++ b/nix/artiq.nix
@@ -1,0 +1,125 @@
+{ stdenv, fetchFromGitHub, fetchsvn, python35Packages, qt5Full, llvm-or1k, llvmlite, python35}:
+
+let
+
+levenshtein = python35Packages.buildPythonPackage rec {
+  name = "levenshtein";
+  src = fetchFromGitHub {
+    owner = "ztane";
+    repo = "python-Levenshtein";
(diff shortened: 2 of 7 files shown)


jordens:
As a first step you should talk to the Windows users.

We had discussed this at great length on IRC a while ago already.

If you ever want to coexist with the rest of a usual Python distribution (that you don't want to maintain, package, and build on multiple architectures yourself) or if you are looking at Windows at all, NIX is just a hip but naive hack that simply doesn't scale to the problems we have. Like docker.

sjmackenzie:
Then close the pull request.

jordens:
Let's leave it for a while. There could be differing views. And if these NIX recipes don't go stale, I'd be the last to reject them.

sjmackenzie:
They are called Nix Expressions.

jordens:
But will you maintain them?

sjmackenzie:
Irritate me with silly hipster comments and it won't happen.

sbourdeauducq:
I've been giving Nixos a try, and I don't think it's fair to call nix a "naive hack". There is definitely a lot of thought that went into it. You can run a complete, usable Linux distro out of nix, and things actually work properly. The main issue I have with it is the lack of documentation, or documentation that has gone out of date.

Compared to conda, nix is a lot faster, more powerful, and less buggy. It also has functionalities equivalent to docker, which you started using.

It reportedly works with the Linux subsystem for Windows, but I have not tried it.

jordens:
Don't make a straman out of that comment. If it is better, if it works on Windows as well, it it leads to less maintenance, if it gives anywhere near the completeness of a Python distribution like conda, if it is well documented, and if it means an end to the complaining, I am clearly wrong.
And I still don't think Docker is a solution to the problems we have with conda. But on top of a packaging system it is a good solution to other problems IME.

sbourdeauducq:
> If it is better,

Quality is definitely higher.

> if it works on Windows as well

I'll test it when I get a chance. That would only work on recent Windows with that Linux subsystem, though (or Cygwin, but Cygwin is buggy and generally crappy).

> it it leads to less maintenance,
> if it means an end to the complaining

It probably does.

> if it gives anywhere near the completeness of a Python distribution like conda,

What is considered "complete" for scientific users? The usual suspects (numpy, scipy, matplotlib, pandas, pyqtgraph, ipython, jupyter) are working just fine.

I think you might also be overestimating the reach of conda; in my experience, many users install conda only because they want to install ARTIQ.

> if it is well documented

Documentation is lacking as I mentioned, in many cases you have to understand some of the nix language and RTFS. But it doesn't sound very complicated to produce a "nix cheat cheat for scientific/conda users" that describes in simple terms the basic operations of managing environments and installing/removing the usual packages. Also, I'd rather learn some nix basics than deal with obscure conda bugs.

marmeladapk:
> It reportedly works with the Linux subsystem for Windows, but I have not tried it.

Which appeared in Windows 10 last year, so Windows 7 wouldn't be supported.

sbourdeauducq:
Things don't look too good with WSL at the moment. Setting things up requires many steps, and nix tickles some WSL bug that need working around - these two problems might get better as there are plans to add NixOS to the MS Store. A bigger problem is the GUI; this needs yet another step to manually install and start a X server on the native Windows side. Then it is slow, with a lot of flickering, and crippled with bugs; there are various rendering issues, and application freezes and crashes. I'm particularly worried that ARTIQ applet embedding will not work.

PR #606, opened 2016-11-05
compiler: Fix break/continue targets in loop else blocks

klickverbot:
Previously, they would still target the respective labels in the
just-exited loop.


whitequark:
All lit tests pass and code looks good to me, thanks.

PR #608, opened 2016-11-07
compiler: Clarify recv_rpc value names and documentation [nfc]

klickverbot:
Previously, the phi emitted for the pointer parameter to recv_rpc was –
rather confusingly – called "size", and the pseudo-code in the comment
had bit-rotted.


whitequark:
Thanks!

PR #610, opened 2016-11-08
[PoC] compiler: Emit all-kernel_invariant objects as LLVM constants

klickverbot:
This enables constant propagation optimisations, as verified by
the included test case. This is only a first stop-gap measure, though;
we should support optimisation based on kernel invariants on a more
fine-grained level.

### DIFF
--- a/artiq/compiler/transforms/llvm_ir_generator.py
+++ b/artiq/compiler/transforms/llvm_ir_generator.py
@@ -1387,6 +1387,7 @@ class LLVMIRGenerator:
         def _quote_attributes():
             llglobal = None
             llfields = []
+            all_attrs_constant = True
             for attr in typ.attributes:
                 if attr == "__objectid__":
                     objectid = self.embedding_map.store_object(value)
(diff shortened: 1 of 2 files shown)


artiq/compiler/transforms/llvm_ir_generator.py
```diff
@@ -1407,13 +1408,17 @@ def _quote_attributes():
                                          not types.is_c_function(typ.attributes[attr]))
                     if is_class_function:
                         attrvalue = self.embedding_map.specialize_function(typ.instance, attrvalue)
+                    if all_attrs_constant:
+                        inst = typ.instance if types.is_constructor(typ) else typ
```
klickverbot: @whitequark: This might be dodgy – I wasn't sure about the exact semantics of `TConstructor`.

whitequark: It is. E.g. in `class Foo:`, TConstructor is the type of the expression `Foo` whereas TInstance is the type of the expression `Foo()`. I'll merge this manually while fixing this.

whitequark:
Merged manually as described above. `kernel_invariants` only currently applies to instance attributes, not class attributes; something like `class_kernel_invariants` would be applicable to class attributes. We can implement this separately if it is desired.

PR #666, opened 2017-02-03
Add tooltips to experiment arguments

cjbe:
This adds optional tooltips to experiment arguments in the Artiq Dashboard and Browser. 

This is useful as the ArgumentEditor gets unwieldy for long argument names - this allows one to use short argument names and add extra context as a tooltip.

### DIFF
--- a/artiq/dashboard/experiments.py
+++ b/artiq/dashboard/experiments.py
@@ -73,6 +73,8 @@ class _ArgumentEditor(QtWidgets.QTreeWidget):
 
             entry = procdesc_to_entry(argument["desc"])(argument)
             widget_item = QtWidgets.QTreeWidgetItem([name])
+            if argument["tooltip"]:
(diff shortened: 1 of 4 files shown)


artiq/dashboard/experiments.py
```diff
@@ -73,6 +73,8 @@ def __init__(self, manager, dock, expurl):
 
             entry = procdesc_to_entry(argument["desc"])(argument)
             widget_item = QtWidgets.QTreeWidgetItem([name])
+            if argument["tooltip"]:
```
sbourdeauducq: Why not ``argument.get("tooltip", None)`` like in the browser above?

cjbe: The inconsistency is from when I was trying to make the browser able to open old results files without the tooltip entry in expid - I decided this was unnecessary. 

I will correct this inconsistency.

sbourdeauducq:
Thanks.

Issue #648, opened 2016-12-19
Absolute imports of repository modules


cjbe:
Is it possible for Artiq to add its experiment repository to the python path, in order to allow absolute imports of experiment modules?

An example use case is a repository with the following structure:
```
repository/
    scans/
        laser_scan.py
    servos/
        laser_servo.py
```
* laser_scan.py contains an experiment that scans a laser detuning
* laser_servo.py contains an experiment that performs a laser_scan, fits the data, and adjusts some experimental parameters

It would be very useful to be able to `import repository.scans.laser_scan` in laser_servo.py

Currently the only way of doing something similar is
1) moving the common code to a package outside the repository (which is sometimes a good thing, but here laser_scan.py is a gnarly volatile experiment, which we want version controlled with the rest of the experiments)
2) having everything in one module, or doing relative imports between neighboring modules, or (shudder) having experiments fiddle with the python path.

sbourdeauducq:
Sounds reasonable. Can you submit a patch (need to take care of master, artiq_run, and documentation)?

sbourdeauducq:
@cjbe In fact, this is a problem with ``artiq_run``, unless we add a ``--repository`` option to it, or require that the user sets ``PYTHONPATH``. But this is quite unfriendly. Maybe it is better to agree that experiments should do relative imports in this case?

jordens:
Can't the workers adjust `sys.path`?

sbourdeauducq:
Yes but that's not the problem: ``artiq_run`` doesn't know about the repository and therefore cannot do it.

cjbe:
@sbourdeauducq I agree this is problematic for artiq_run. My opinion is that artiq_run is a lower level tool that will be mostly used for debugging rather than by the average user, so requiring the user to set PYTHONPATH is OK.

cjbe:
I will tidy up my code for this and submit a PR in a few days.

sbourdeauducq:
@cjbe Do you still want this?

cjbe:
I still think it would be very useful, but have not come up with a tidy implementation. Happy to close this issue if you want.

sbourdeauducq:
Yes. I'm not sure if there is any tidy solution that doesn't introduce a problem at least as bad as the initial one.

charlesbaynham:
To revisit a long-dead issue, is there a better solution to this these days? I agree with @cjbe that absolute imports would be really useful. 

[...]

PR #668, opened 2017-02-04
Add experiment repository to experiment PYTHONPATH

cjbe:
This implements #648.

If the git backend is used the repository is checked out as 'repository' into a clean temporary directory, which is added to the worker PYTHONPATH. If the filesystem backend is used, the parent of the repository directory is added to the worker PYTHONPATH. So long as the repository is a valid python module it can then be imported as 'repository' (using git backend) or whatever the repository directory name is (using filesystem backend). 

A potential nasty is adding things to the PYTHONPATH unintentionally when using the filesystem backend, but this can be avoided by putting the repository directory in an (otherwise) empty directory. In any case, I assume all serious work will be done out of a git repository.

At the moment artiq_browser does not know about the experiment repository, hence absolute imports in experiments will not work unless the user manually appends to PYTHONPATH. I don't see a neat solution to this - one option is a 'repository' command line argument that is added to the worker PYTHONPATH.

If everyone is happy with this implementation I will document it.

### DIFF
--- a/artiq/master/experiments.py
+++ b/artiq/master/experiments.py
@@ -64,7 +64,7 @@ class _RepoScanner:
                         exc_info=not isinstance(exc, WorkerInternalException))
                     # restart worker
                     await self.worker.close()
-                    self.worker = Worker(self.worker_handlers)
+                    self.worker = self._create_worker()
             if de.is_dir():
                 subentries = await self._scan(
                     root, os.path.join(subdir, de.name))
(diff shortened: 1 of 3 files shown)


sbourdeauducq:
Problems with artiq_run, and now with artiq_browser and potential misbehavior when the repository directory is not in an empty directory... are you sure relative imports are not better?

sbourdeauducq:
@cjbe Do you insist on this or can we close this pull request?

cjbe:
Lets close it - I still think absolute imports are really useful, but this is not the right way of doing it.

PR #699, opened 2017-03-30
language,gui: combine LinearScan, RandomScan into RangeScan. Closes #679

### LINKED ISSUE #679: Link Scannable LinearScan and RandomScan ranges
It would be useful for the LinearScan and RandomScan min/max/nsteps to be linked, so that when one changes from LinearScan to RandomScan the scan parameters are copied across.

This seems to match typical use cases better than the current behavior (RandomScan and LinearScan parameters independent, so that if you set up your scan nicely, then want to change mode, you have to set the parameters again).

cjbe:


artiq/language/scan.py
```diff
@@ -59,52 +59,28 @@ def describe(self):
...
-        self.seed = seed
-        self.sequence = list(LinearScan(start, stop, npoints))
-        if seed is None:
-            rf = random.random
-        else:
-            rf = Random(seed).random
-        random.shuffle(self.sequence, rf)
+        if randomize:
+            if seed is not None:
+                random.seed(seed)
```
sbourdeauducq: Even if this has no practical consequences here, I'd like to avoid touching global state when that can be avoided, which is why I was using ``random.Random`` in the original code. Your seeding code would break if there are two threads with different seeds executing this at the same time.

artiq/gui/entries.py
```diff
@@ -255,15 +258,22 @@ def update_npoints(value):
             if npoints.value() != value:
                 npoints.setValue(value)
 
+        def update_randomize(value):
+            state["randomize"] = value
+            if randomize.isChecked() != value:
```
sbourdeauducq: Is this test necessary?

artiq/language/scan.py
cjbe: I did it this way as the original implementation was broken, but I will fix this.

sbourdeauducq:
Thanks

sbourdeauducq:
@cjbe You forgot to update the examples. I fixed them.

PR #725, opened 2017-04-26
master: record time run() is called

cjbe:
The wall time when build() is called is currently recorded as the results key 'start_time'. 
This adds an additional key 'run_time', the time when run() is called.

This is useful for reconstructing experiment time-lines as the build time is often very different from the run time.


jordens:
I am generally a bit worried about the `localtime()` usage (same for `start_time`). I would prefer `time()` (but still use local timezone conversions for the naming of the directories and files).
Otherwise LGTM.

jbqubit:
Good idea! Thanks. :)

PR #734, opened 2017-05-13
firmware: fix revision of compiler_builtins for satman

cjbe:


whitequark:
Why is this necessary exactly?

cjbe:
Satman does not compile using the master branch of compiler_builtins. This pins it back to a working version -the same fix was applied to runtime in 52a046d

PR #808, opened 2017-08-02
master: rotate logs at midnight, rather than on log size

cjbe:
Currently the master rotates logs based on file size, with numeric postfixes indicating the log age - this means that it is hard to determine which log file contains the output from a given experiment.

This patch changes the log rotation to occur at midnight. This means that after the log rolls over there is an easy to understand and immutable relationship between the experiment and the log file containing its output.


cjbe:
@sbourdeauducq TimedRotatingFileHandler uses time.localtime() to determine the date string in the same way as we do, so the notion of 'midnight' is consistent.

jordens:
thanks

PR #809, opened 2017-08-03
master: use epoch time for timestamps (closes #726)

### LINKED ISSUE #726: use time() for timestamps
`start_time` and `run_time` should be epoch and not `localtime()` to avoid timezone, DST, and leap second issues. directories and filenames can remain in local time.

cjbe:
Use time.time() for experiment run_time and start_time - the results directory names are still based on local time.

### DIFF
--- a/artiq/master/worker_impl.py
+++ b/artiq/master/worker_impl.py
@@ -242,7 +242,7 @@ def main():
             obj = get_object()
             action = obj["action"]
             if action == "build":
-                start_time = time.localtime()
+                start_time = time.time()
                 rid = obj["rid"]
                 expid = obj["expid"]
                 if obj["wd"] is not None:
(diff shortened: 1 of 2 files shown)


artiq/master/worker_impl.py
```diff
@@ -286,8 +287,8 @@ def main():
                     dataset_mgr.write_hdf5(f)
                     f["artiq_version"] = artiq_version
                     f["rid"] = rid
-                    f["start_time"] = int(time.mktime(start_time))
-                    f["run_time"] = int(time.mktime(run_time))
+                    f["start_time"] = int(start_time)
+                    f["run_time"] = int(run_time)
```
jordens: Let's make these doubles while we are at it. I see no reason to truncate it here.

sbourdeauducq:
Can't we use epoch time everywhere?

jordens:
Lab books are kept in local time and being able to correlate lab book entries with data is high value. Otherwise we could just use run id.

sbourdeauducq:
Yes, having epoch time in files and local time everywhere else (logs, directories) can be reasonable for manual use. But it makes it a bit tricky to programmatically locate experiment files based on time.

cjbe:
Using local time makes it much easier to correlate with other lab timestamps (i.e. a wristwatch!). As long as we have run_time in epoch time we can unambiguously recreate the experiment timeline.

jordens:
@sbourdeauducq To make locating files by runid or epoch easier we should log those tuples `(runid,start_time,filename)` to a file.

sbourdeauducq:
OK.

jordens:
thanks. i amended the log entry and resolved the conflict.

PR #816, opened 2017-08-22
add test for artiq_rpctool

mntng:


artiq/test/test_rpctool.py
```diff
@@ -0,0 +1,38 @@
...
+        proc = asyncio.create_subprocess_shell(
+                            sys.executable + " -m artiq.frontend.artiq_rpctool ::1 3249 call message", 
+                            stdout = asyncio.subprocess.PIPE,
+                            stderr = asyncio.subprocess.PIPE)
+        result = await proc
+        (value, err) = await result.communicate()
+        self.assertEqual(value.decode('ascii').rstrip(), '4')
+        await result.wait()
+
+    async def start(self):
```
sbourdeauducq: This should be called ``do_test`` or similar, not ``start``.

artiq/test/test_rpctool.py
```diff
@@ -0,0 +1,38 @@
+from artiq.protocols.pc_rpc import Server
+import asyncio
```
sbourdeauducq: Put system imports before artiq imports.

artiq/test/test_rpctool.py
```diff
@@ -0,0 +1,38 @@
...
+import sys
+import unittest
+
+class Target:
+    def message(self):
+        return 4
+
+class Rpctool_test(unittest.TestCase):
```
sbourdeauducq: Use CamelCase to name classes (see: https://www.python.org/dev/peps/pep-0008/).

artiq/test/test_rpctool.py
```diff
@@ -0,0 +1,38 @@
...
+    async def get_message(self):
```
sbourdeauducq: This is not just getting the message, it is also checking its value. Rename that function, or just merge it into the ``do_test`` function below.

artiq/test/test_rpctool.py
```diff
@@ -0,0 +1,38 @@
...
+        proc = asyncio.create_subprocess_shell(
+                            sys.executable + " -m artiq.frontend.artiq_rpctool ::1 3249 call message", 
+                            stdout = asyncio.subprocess.PIPE,
+                            stderr = asyncio.subprocess.PIPE)
+        result = await proc
+        (value, err) = await result.communicate()
```
sbourdeauducq: No checking on ``err``?

artiq/test/test_rpctool.py
```diff
@@ -0,0 +1,38 @@
+from artiq.protocols.pc_rpc import Server
+import asyncio
+import sys
+import unittest
+
+class Target:
+    def message(self):
+        return 4
```
sbourdeauducq: I suggest using a more uncommon value, e.g. 385710948. Or a string.

artiq/test/test_rpctool.py
```diff
@@ -0,0 +1,38 @@
...
+        await result.wait()
+
+    async def start(self):
+        server = Server({"target": Target()}, None, True, True)
+        await server.start("::1", 3249)
+        await self.get_message()
+        await server.stop()
+        
+    def test_rpc(self):
+        if sys.platform == "win32":
```
sbourdeauducq: Other code in ARTIQ (e.g. ``artiq_master.py``) tests this with ``if os.name == "nt"``. Please do the same for consistency, plus I'm not even sure if ``win32`` really works reliably (what about 64-bit systems)?

PR #818, opened 2017-08-24
add test_rpctool.py

mntng:


artiq/test/test_rpctool.py
```diff
@@ -0,0 +1,38 @@
+import os, sys
+import asyncio
+import unittest
+from artiq.protocols.pc_rpc import Server
```
sbourdeauducq: Separate from system imports with a blank line.

artiq/test/test_rpctool.py
```diff
@@ -0,0 +1,38 @@
...
+class Target:
+    def output_value(self):
+        return 4125380
+
+class rpctool_test(unittest.TestCase):
+    async def check_value(self):
+        proc = asyncio.create_subprocess_shell(
+                            sys.executable + " -m artiq.frontend.artiq_rpctool ::1 3249 call output_value", 
+                            stdout = asyncio.subprocess.PIPE,
+                            stderr = None)
```
sbourdeauducq: ``None`` is the default. This line can be removed.

artiq/test/test_rpctool.py
```diff
@@ -0,0 +1,38 @@
...
+from artiq.protocols.pc_rpc import Server
+
+class Target:
+    def output_value(self):
+        return 4125380
+
+class rpctool_test(unittest.TestCase):
+    async def check_value(self):
+        proc = asyncio.create_subprocess_shell(
+                            sys.executable + " -m artiq.frontend.artiq_rpctool ::1 3249 call output_value",
```
sbourdeauducq: This will fail if ``sys.executable`` contains a space character. Better pass the arguments with a list and use ``create_subprocess_exec`` as done elsewhere.

sbourdeauducq: Also, 3249 is already used by another ARTIQ service. Better pick another port, e.g. 7777 as used in test_pc_rpc.py.

artiq/test/test_rpctool.py
```diff
@@ -0,0 +1,38 @@
...
+    def output_value(self):
+        return 4125380
+
+class rpctool_test(unittest.TestCase):
+    async def check_value(self):
+        proc = asyncio.create_subprocess_shell(
+                            sys.executable + " -m artiq.frontend.artiq_rpctool ::1 3249 call output_value", 
+                            stdout = asyncio.subprocess.PIPE,
+                            stderr = None)
+        result = await proc
```
sbourdeauducq: You could just do ``proc = await asyncio.create_subprocess_exec(...) ... await proc.communicate()...``

artiq/test/test_rpctool.py
```diff
@@ -0,0 +1,38 @@
+import os, sys
+import asyncio
+import unittest
+from artiq.protocols.pc_rpc import Server
+
+class Target:
+    def output_value(self):
+        return 4125380
+
+class rpctool_test(unittest.TestCase):
```
sbourdeauducq: Underscore separation is the naming convention for functions, not classes. How about ``TestRPCTool``?

artiq/test/test_rpctool.py
```diff
@@ -0,0 +1,38 @@
...
+            loop = asyncio.ProactorEventLoop()
+            asyncio.set_event_loop(loop)
+        else:
+            loop = asyncio.get_event_loop()
+        try:
+            loop.run_until_complete(self.do_test())
+        finally:
+            loop.close()
+
+
```
sbourdeauducq: Only one blank line is enough.

artiq/test/test_rpctool.py
mntng: I get this when using` subprocess_exec`: `[WinError 2] The system cannot find the file specified`, which happened the same previously when I use `Popen` without adding `shell=True`.
Also, only string arguments are accepted. Can I simply replace `sys.executable` by `python`?

artiq/test/test_rpctool.py
mntng: I got this:  generator object has no attribute "communicate()"

artiq/test/test_rpctool.py
sbourdeauducq: What is the complete code you were using? It should be with multiple arguments, something like ``create_subprocess_exec(sys.executable, "-m", "artiq.frontend.artiq_rpctool", ...)`` and not a single string.

artiq/test/test_rpctool.py
sbourdeauducq: Forgot ``await``?

sbourdeauducq:
@mntng No, ``sys.executable`` cannot be replaced with ``"python"``. There can be multiple Python interpreters available on the system, and on a number of them (e.g. Debian) ``"python"`` refers to Python 2.7, which would break the test.

There are a few remaining minor style issues but otherwise this looks good to me.

artiq/test/test_rpctool.py
```diff
@@ -0,0 +1,36 @@
...
+    async def check_value(self):
+        proc = await asyncio.create_subprocess_exec(
+                            sys.executable, "-m", "artiq.frontend.artiq_rpctool", "::1", "7777", "call", "output_value",
+                            stdout = asyncio.subprocess.PIPE)
+        (value, err) = await proc.communicate()
+        self.assertEqual(value.decode('ascii').rstrip(), '4125380')
+        await proc.wait()
+
+    async def do_test(self):
+        server = Server({"target": Target()}, None, True, True)
```
sbourdeauducq: Are the default values ``description=None, builtin_terminate=False, allow_parallel=False`` not working? If they are working then please use them.

artiq/test/test_rpctool.py
```diff
@@ -0,0 +1,36 @@
+import os, sys
+import asyncio
+import unittest
+
+from artiq.protocols.pc_rpc import Server
+
```
sbourdeauducq: 2 blank lines to separate.

artiq/test/test_rpctool.py
```diff
@@ -0,0 +1,36 @@
...
+class Target:
+    def output_value(self):
+        return 4125380
+
```
sbourdeauducq: ditto

artiq/test/test_rpctool.py
```diff
@@ -0,0 +1,36 @@
+import os, sys
```
sbourdeauducq: import os
import sys
(see PEP8)

PR #831, opened 2017-09-20
set maximal divider value for SPI write and read clk

mntng:


artiq/coredevice/spi.py
```diff
@@ -155,6 +155,8 @@ def set_config_mu(self, flags=0, write_div=6, read_div=6):
           by one RTIO clock cycle.
         :param read_div: Ditto for the read clock.
         """
+        if write_div > 256 or read_div > 256:
```
jordens: The max valid is 257 at that point. And you will want to check for >=2 as well.

jordens:
Oh. And please, please please, clean up your git history.

PR #855, opened 2017-11-25
Label masters with friendly name, store dashboard config per master (closes #848)

### LINKED ISSUE #848: Label masters with a name to distinguish dashboards
It would be useful for Artiq masters to have a configurable name, and have the dashboards show the logical name of the master they are connected to, rather than just an IP address. 

At the moment is it rather confusing when connecting to multiple masters with similar addresses. This is particularly necessary if one is wants to run multiple masters on one host (as we are planning).

cjbe:
This PR implements two features which make Artiq more pleasant to use when connecting to multiple masters:
1) The dashboard config is (by default) stored separately for each different master, keyed by server address and port. This means that applets and experiment arguments don't mix between different masters.
2) The master can be labelled with a friendly name. If this is present, the dashboard displays this rather than the server address - this is particularly useful when connect to two masters on the same server (distinguished only by port).

### DIFF
--- a/artiq/frontend/artiq_dashboard.py
+++ b/artiq/frontend/artiq_dashboard.py
@@ -218,8 +218,11 @@ def main():
     if d_log0 is not None:
         main_window.tabifyDockWidget(d_schedule, d_log0)
 
+    server_description = server_name
+    if server_name != args.server:
+        server_description += " ({})".format(args.server)
     logging.info("ARTIQ dashboard %s connected to %s",
-                 artiq_version, server_name)
+                 artiq_version, server_description)
 
     # run
     main_window.show()
--- a/artiq/frontend/artiq_dashboard.py
+++ b/artiq/frontend/artiq_dashboard.py
@@ -21,9 +21,6 @@ from artiq.dashboard import (experiments, shortcuts, explorer,
 
 
 def get_argparser():
-    default_db_file = os.path.join(get_user_config_dir(),
-                                   "artiq_dashboard.pyon")
-
     parser = argparse.ArgumentParser(description="ARTIQ Dashboard")
     parser.add_argument(
         "-s", "--server", default="::1",
@@ -38,9 +35,8 @@ def get_argparser():
         "--port-broadcast", default=1067, type=int,
         help="TCP port to connect to for broadcasts")
     parser.add_argument(
-        "--db-file", default=default_db_file,
-        help="database file for local GUI settings "
-             "(default: %(default)s)")
+        "--db-file", default=None,
+        help="database file for local GUI settings")
     verbosity_args(parser)
     return parser
 
@@ -92,6 +88,12 @@ def main():
     # initialize application
     args = get_argparser().parse_args()
     widget_log_handler = log.init_log(args, "dashboard")
+
+    if args.db_file is None:
+        args.db_file = os.path.join(get_user_config_dir(),
+                           "artiq_dashboard_{server}_{port}.pyon".format(
+                            server=args.server.replace(":","."),
+                            port=args.port_notify))
 
     app = QtWidgets.QApplication(["ARTIQ Dashboard"])
     loop = QEventLoop(app)
--- a/artiq/frontend/artiq_dashboard.py
+++ b/artiq/frontend/artiq_dashboard.py
@@ -12,7 +12,7 @@ from quamash import QEventLoop
 from artiq import __artiq_dir__ as artiq_dir, __version__ as artiq_version
 from artiq.tools import (atexit_register_coroutine, verbosity_args,
                          get_user_config_dir)
-from artiq.protocols.pc_rpc import AsyncioClient
+from artiq.protocols.pc_rpc import AsyncioClient, Client
 from artiq.protocols.broadcast import Receiver
 from artiq.gui.models import ModelSubscriber
 from artiq.gui import state, log
@@ -110,6 +110,11 @@ def main():
         atexit.register(client.close_rpc)
         rpc_clients[target] = client
 
+    config = Client(args.server, args.port_control, "master_config")
+    server_name = config.get_name()
+    if server_name is None:
+        server_name = args.server
+
     disconnect_reported = False
     def report_disconnect():
         nonlocal disconnect_reported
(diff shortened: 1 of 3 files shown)


artiq/frontend/artiq_dashboard.py
```diff
@@ -108,6 +110,11 @@ def main():
         atexit.register(client.close_rpc)
         rpc_clients[target] = client
 
+    config = Client(args.server, args.port_control, "master_config")
```
sbourdeauducq: This needs freeing of resources.

artiq/frontend/artiq_master.py
```diff
@@ -48,6 +48,8 @@ def get_argparser():
 
     log_args(parser)
 
+    parser.add_argument("--name", help="friendly name")
```
sbourdeauducq: This help text is a bit terse.

artiq/frontend/artiq_dashboard.py
```diff
@@ -93,6 +89,12 @@ def main():
     args = get_argparser().parse_args()
     widget_log_handler = log.init_log(args, "dashboard")
 
+    if args.db_file is None:
+        args.db_file = os.path.join(get_user_config_dir(),
+                           "artiq_dashboard_{server}_{port}.pyon".format(
+                            server=args.server.replace(":","."),
```
sbourdeauducq: Is ``:`` actually forbidden in filenames on any platform? FWIW Linux uses it in sysfs.

artiq/frontend/artiq_master.py
```diff
@@ -94,6 +96,12 @@ def ccb_issue(service, *args, **kwargs):
     scheduler.start()
     atexit_register_coroutine(scheduler.stop)
 
+    class MasterConfig:
+        def get_name(self):
+            return args.name
```
sbourdeauducq: For the purposes of limiting the complexity of the ``main`` function, which is getting quite heavy already, I would put the class definition outside.

sbourdeauducq: The name can be easily passed to the constructor.

artiq/frontend/artiq_master.py
```diff
@@ -112,7 +120,8 @@ def ccb_issue(service, *args, **kwargs):
         "master_device_db": device_db,
         "master_dataset_db": dataset_db,
         "master_schedule": scheduler,
-        "master_experiment_db": experiment_db
+        "master_experiment_db": experiment_db,
+        "master_config": config
```
sbourdeauducq: Trivial style issue, but I think it should be first in that list, since this is the first thing that gets accessed.

artiq/frontend/artiq_dashboard.py
```diff
@@ -209,7 +216,7 @@ def report_disconnect():
         main_window.tabifyDockWidget(d_schedule, d_log0)
 
     logging.info("ARTIQ dashboard %s connected to %s",
-                 artiq_version, args.server)
+                 artiq_version, server_name)
```
sbourdeauducq: What happens if the master was not run with ``--name``?

artiq/frontend/artiq_dashboard.py
```diff
@@ -38,9 +35,8 @@ def get_argparser():
         "--port-broadcast", default=1067, type=int,
         help="TCP port to connect to for broadcasts")
     parser.add_argument(
-        "--db-file", default=default_db_file,
-        help="database file for local GUI settings "
-             "(default: %(default)s)")
+        "--db-file", default=None,
+        help="database file for local GUI settings")
```
sbourdeauducq: I don't like losing the path to the file. Maybe add "by default in ... and dependent on master hostname".

cjbe: OK

artiq/frontend/artiq_dashboard.py
cjbe: Its forbidden on Windows, alas.

artiq/frontend/artiq_dashboard.py
cjbe: OK

artiq/frontend/artiq_dashboard.py
cjbe: `server_name` defaults to `args.server`, so if `--name` is not provided the behaviour is as before.

artiq/frontend/artiq_master.py
cjbe: OK

artiq/frontend/artiq_dashboard.py
sbourdeauducq: May be better to have both pieces of information in the log.

cjbe:
@sbourdeauducq I think I have now addressed all of your points

artiq/frontend/artiq_dashboard.py
```diff
@@ -108,6 +112,12 @@ def main():
         atexit.register(client.close_rpc)
         rpc_clients[target] = client
 
+    config = Client(args.server, args.port_control, "master_config")
+    atexit.register(config.close_rpc)
+    server_name = config.get_name()
+    if server_name is None:
+        server_name = args.server
```
sbourdeauducq: It's a bit tortuous with the (imprecise) test ``if server_name != args.server`` below. I would keep the initial value of this variable, and then do ``MainWindow(args.server if server_name is None else server_name)`` and test again ``server_name is None`` to generate the log message.

artiq/frontend/artiq_dashboard.py
```diff
@@ -108,6 +112,12 @@ def main():
         atexit.register(client.close_rpc)
         rpc_clients[target] = client
 
+    config = Client(args.server, args.port_control, "master_config")
+    atexit.register(config.close_rpc)
```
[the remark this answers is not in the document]
sbourdeauducq: (Use try/finally)

artiq/frontend/artiq_dashboard.py
cjbe: Yep - that is much neater.

artiq/frontend/artiq_dashboard.py
```diff
@@ -113,10 +113,10 @@ def main():
...
 
     config = Client(args.server, args.port_control, "master_config")
-    atexit.register(config.close_rpc)
-    server_name = config.get_name()
-    if server_name is None:
-        server_name = args.server
+    try:
+        server_name = config.get_name()
+    finally:
+        config.close_rpc()
```
cjbe: I assume this is what you were looking for? Client() tidies up after itself if it has an exception, so the only logic that needs guarding seems to be the RPC.

sbourdeauducq: Yes.

PR #875, opened 2017-12-15
dashboard: Restore minimized experiments when trying to open them again

klickverbot:
When the user tried to open an experiment from the explorer that
already existed, previously "nothing would happen" (focus change
without the window being restored).


jordens:
LGTM. Thanks

sbourdeauducq:
Does the isMinimized test do anything? (other than hint what this code is for)

PR #889, opened 2018-01-11
firmware: make read leveling robust for KUS SDRAM

cjbe:
Increases the initial delay step into the valid read window from 8 -> 16 as with the original delay I was not getting out of the noisy transition window, as evidenced by seeing read delay windows of only 8 LSB ~10% of the time, leading to failing memory tests.

See discussion on https://irclog.whitequark.org/m-labs/2018-01-11


sbourdeauducq:
Thanks!

PR #901, opened 2018-01-19
Add register dump on HMC830 lock timeout.

hartytp:


hartytp:
Not tested on hardware yet (sorry!)

PR #922, opened 2018-02-16
Hmc7043

hartytp:
Update HMC7043 setup code. Verified with a scope that the DAC clocks are 600MHz for a 1.2GHz input and that the SYSREF is 9.375MHz, with an adjustable phase.

Still to do:
- Check which outputs we actually need to enable
- Check input/output termination settings.


artiq/firmware/libboard_artiq/hmc830_7043.rs
```diff
@@ -113,7 +113,8 @@ mod hmc830 {
                     if addr == 0x0d || addr == 0x0e { continue; }
                     error!("[0x{:02x}] = 0x{:04x}", addr, read(addr));
                 }
-                return Err("HMC830 lock timeout");
+            //    return Err("HMC830 lock timeout");
```
sbourdeauducq: Patches should do a single thing, so disabling the HMC830 should not be here.

sbourdeauducq:
Works here too. Thanks!

sbourdeauducq:
Also I think the delay adjustment code should be broken out into separate functions, so we can easily implement automatic scans (which will be used to determine the fixed values to put into the code).

artiq/firmware/libboard_artiq/hmc830_7043.rs
hartytp: Sorry, that wasn't supposed to be there.

hartytp:
I agree the delay should be broken out.

In the long run, I also think it might be worthwhile putting all the configuration settings like divider values in a single place, rather than scattered hardcoded numbers through the codebase. At the moment, changing the DAC clock frequency will mean digging through the code to remember which numbers need to be changed (RTIO frequency, DAC clock divider, DRTIO transceiver settings, etc)

PR #969, opened 2018-03-24
Zotino

hartytp:


artiq/coredevice/ad53xx.py
```diff
@@ -0,0 +1,328 @@
...
+      :const:`AD53XX_CMD_GAIN`
+    :return: The 24-bit word to be written to the DAC, aligned as the 24 MSB of
+      a 32-bit integer, ready to be transferred directly by the SPI core.
+    """
+    return (AD53XX_CMD_SPECIAL | AD53XX_SPECIAL_READ | op |
+           (((channel & 0x3f) + 8) << 7)) << 8
+
+
+@portable
+def voltage_to_mu(voltage, offset_dacs=8192, vref=5.):
```
whitequark: I think that calling both units of time and voltage `mu` is going to cause confusion at some point.

artiq/coredevice/ad53xx.py
```diff
@@ -0,0 +1,328 @@
...
+    :param core_device: Core device name (default: "core")
+    """
+    kernel_invariants = {"bus", "ldac", "clr", "chip_select", "div_write",
+                         "div_read", "vref", "core"}
+
+    def __init__(self, dmgr, spi_device, ldac_device, clr_device,
+                 chip_select=1, div_write=4, div_read=8, vref=5.,
+                 offset_dacs=8192, core="core"):
+        self.bus = dmgr.get(spi_device)
+        self.ldac = dmgr.get(ldac_device)
```
sbourdeauducq: @hartytp Did you eventually deem it unnecessary to support configurations without LDAC?

artiq/coredevice/zotino.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+
+    def __init__(self, dmgr, spi_device, ldac_device, clr_device,
+                 div_write=4, div_read=8, vref=5., core="core"):
+        AD53xx.__init__(self, dmgr=dmgr, spi_device=spi_device,
+                        ldac_device=ldac_device, clr_device=clr_device,
+                        chip_select=_SPI_CS_DAC, div_write=div_write,
+                        div_read=div_read, core=core)
+
+    @ kernel
```
sbourdeauducq: no space

artiq/coredevice/zotino.py
```diff
@@ -0,0 +1,53 @@
...
+
+Output event replacement is not supported and issuing commands at the same
+time is an error.
+"""
+
+from artiq.language.core import kernel
+from artiq.coredevice import spi2 as spi
+from artiq.coredevice.ad53xx import SPI_AD53XX_CONFIG, AD53xx
+
+_SPI_DAC_CONFIG = SPI_AD53XX_CONFIG
```
sbourdeauducq: I don't think this redefinition helps.

sbourdeauducq:
There should be an entry in RELEASE_NOTES about the rename to ``ad53xx`` and methods that have changed - written for users of the old driver that need to switch to the new.

artiq/coredevice/ad53xx.py
hartytp: I take your point, but we currently use mu for all units even freq phi and amp for dds. So this is just being consistent (which is the least confusing thing to do imo).

artiq/coredevice/ad53xx.py
hartytp: True that works, but it seems a but heavyweight for me unless there is a concrete use case for it - I'm not aware of one. My approach is to start simple and add that kind of feature when a need is expressed.

If you want to add this during merge I don't object.

artiq/coredevice/zotino.py
hartytp: Thanks! Will fix

artiq/coredevice/zotino.py
hartytp: Yes let's scrap it. Remnant f old version of code. Will fix.

hartytp:
Yes. Will do once we agree that the changes are ok.

artiq/coredevice/ad53xx.py
sbourdeauducq: Just need to check that e.g. NIST don't need it... @dhslichter ?

hartytp: Assuming no major objections, would you mind finishing this off from here? This is now all working on the hw and I'd like to prioritize analogy tests on Zotino.

sbourdeauducq: Ok.

hartytp: @sbourdeauducq thinking about this more: using this without ldac is probably a rare use case, so let's not clutter up the driver with it if we can avoid it. What about having a general dummy ttl that the device manager can provide for this kind of case?

jordens:
In general commit messages should be as concise and readable as possible. That means including the area where the work was done. "fix core device addr." is not decipherable. And if at all possible, that commit and the change it reverts should not appear in the history.
We'll squash this so it doesn't appear.

artiq/coredevice/ad53xx.py
jordens: IMO `mu` is ok for an opaque quantity of machine units independent of what the quantity describes if the corresponding SI unit is always clear from the context and it's not used as a variable name (`mu` itself is not ok, but `voltage_{to_}_mu` is fine).

jordens:
Is "Test Zotino on hw" a TODO or was that done?
Also next time:
- [ ] RELEASE_NOTES
- [ ] update other examples and device_dbs
- [ ] documentation

hartytp:
Testing done.

Other points ack.

jordens:
We should probably write up a check list for pull requests in the contribution guidelines.

hartytp:
> We should probably write up a check list for pull requests in the contribution guidelines.

Sounds like a good idea.

FWIW, that wasn't intended as merge-ready code, but more of a design sketch showing the interface and timings I wanted. Thanks a lot @jordens and @sbourdeauducq for finishing it off and turning it into high-quality code!

artiq/coredevice/ad53xx.py
dhslichter: @sbourdeauducq @hartytp we have some DACs that we use without LDAC (it's tied low on the board).  We like having this option because then you free a TTL line for other uses.

jordens: That option was subsequently added back in.
5ca59467fdfca01e3b0754b1cc6a1d7fbf6ec036

dhslichter: Thanks @jordens!

PR #972, opened 2018-03-27
Sampler rename v_ref to v_span, fix default value in adc_mu_to_volt

sotirova:


jordens:
Hmm. A little comment, explanation, rationale would be helpful.
Why? What are you trying to "fix"? Is it broken?

* The ADC differential span is 10V, yes, BUT
* The Sampler input span (range) is 20V, 2V, 0.2V, 0.02V in the different gain settings. It is never 10V.
* Compared to that ambiguous variable (span), the reference voltage stays constant.
* If you want to, and if it improves your understanding of the code, you can edit `v_ref=5` and rewrite the `adc_mu_to_volt()` to apply the 0.5 AFE gain.

hartytp:
@jordens The rational here is:
1. `sampler.Sampler.v_ref` is hard-coded to `10.` https://github.com/m-labs/artiq/blob/5ca59467fdfca01e3b0754b1cc6a1d7fbf6ec036/artiq/coredevice/sampler.py#L55 but the `v_ref` parameter in `sampler.adc_mu_to_volt` defaults to `5.`. This inconsistency means that calling `sampler.adc_mu_to_volt(data, gain)` does not give the correct voltage. I would prefer to see the `vref` parameter default in `sampler.adc_mu_to_volt` set to match `sampler.Sampler.v_ref`.
2. I find it confusing that `sampler.Sampler.v_ref = 10.` when the reference voltage is 5V not 10V. Renaming this to span was an attempt to correct that inconsistency. But, I take your point that the suggested alternative may not be any better.

Since the reference is fixed on Sampler and is not likely to change, maybe a better option would be to completely remove all references to `v_ref` and just have a fixed value of `10.` in the `adc_mu_to_volt` function (no need for `sampler.Sampler` to store that value) but, I'm open to suggestions.

hartytp:
NB `span` here was intended to denote the ADC differential input span, not Sampler's input span. I don't think that's as confusing as all that, but maybe the comments could have been clearer.

Anyway, unless you object I think the best solution is:
1. Remove `sampler.Sampler.v_ref`
2. Remove the `v_ref` input value to `sampler.adc_mu_to_volt`
3. Hard-code an appropriate conversion factor into `sampler.adc_mu_to_volt`.

jordens:
Fixed magic values have horrible semantics.

What about my suggestion above?

hartytp:
1. `Sampler` already has a fixed value of `v_ref` (there is no `v_ref` argument in the constructor).
2. I don't see the need for an input parameter reference voltage to the `adc_mu_to_volt` when the reference voltage is not easily changeable.

Other than that, it's fine. If you feel strongly we can do it your way..

jordens:
Ok. Then let's remove the attribute and the argument. But keep it as a variable in the function and comment it.

hartytp:
@jordens that sounds good to me. @sotirova can you have a go at doing that, please (remember to revert the past commit first!)?

FWIW, if someone wants it back as a parameter later on (e.g. if for some reason we start using variants with different reference voltages) I'd like to make sure that `v_ref` always refers to the actual reference voltage and is used consistently rather than the current usage where `sampler.Sampler.v_ref=10.` but `sampler.adc_mu_to_volt` `v_ref` defaults to `5.`

sotirova:
@hartytp   @jordens  Thanks for the comments! Will try to change it asap!

jordens:
NB: Use `git rebase --interactive` to squash, reword and clean up your changes. `git revert` is probably not what you are looking for.

hartytp:
Yes, this should be closed.

PR #988, opened 2018-04-27
Remove Sampler v_ref parameter.

hartytp:
Replaces #972

Context:
- Currently `adc_mu_to_volt` takes a parameter `v_ref`
- This parameter isn't necessary, since there are no plans to make the reference voltage switchable on Sampler 
- The way the code is written is a bit confusing, since passing `v_ref=5.` (the default value) does not give the correct mu->input voltage conversion, since `adc_mu_to_volt` does not take account of the gain of the differential input stage. 
- `Sampler` internally uses `v_ref=10.`, which is confusing, since (a) this isn't the reference voltage used in the PCB/schematic (b) it's not the default value, so passing the output of `sample` dumbly into `adc_mu_to_volt` gives the wrong answer.

@jordens apologies, I haven't got HW set up to test this atm, but I wanted to get this off my to do list. Do you mind checking it/fixing it for me? Thanks!

### DIFF
--- a/artiq/coredevice/sampler.py
+++ b/artiq/coredevice/sampler.py
@@ -16,17 +16,17 @@ SPI_CS_PGIA = 1  # separate SPI bus, CS used as RCLK
 
 
 @portable
-def adc_mu_to_volt(data, gain=0, v_ref=5.):
+def adc_mu_to_volt(data, gain=0):
     """Convert ADC data in machine units to Volts.
 
     :param data: 16 bit signed ADC word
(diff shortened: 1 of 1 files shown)


jordens:
* Please close #972 if you don't need that anymore (or mention there that it's abandoned).
* Could you change the commit message to "sampler: remove v_ref parameter" to be in line with the majority of the other commits?
* Otherwise looks good. I don't have the hardware right now either but will test in two weeks when I'm back and have it.

hartytp:
will do.

PR #1049, opened 2018-06-05
Sayma RTM: hold hmc7043 in reset/mute state during init.

hartytp:
Tested on my Sayma. Since implementing this fix, I haven't seen any more runtime crashes (although, serwb still seems unreliable on my board).

This requires the HMC7043 RESET line to be connected in hardware. My board had been reworked to short this to GND. NB connecting the HMC7043 RESET line breaks the Si5324 because they share a reset line.

### DIFF
--- a/artiq/firmware/libboard_artiq/hmc830_7043.rs
+++ b/artiq/firmware/libboard_artiq/hmc830_7043.rs
@@ -146,6 +146,7 @@ pub mod hmc7043 {
     const DAC_CLK_DIV: u32 = 2;
     const FPGA_CLK_DIV: u32 = 8;
     const SYSREF_DIV: u32 = 128;
+    const HMC_SYSREF_DIV: u32 = SYSREF_DIV*8; // Must be <= 4MHz
 
     // enabled, divider, analog phase shift, digital phase shift
     const OUTPUT_CONFIG: [(bool, u32, u8, u8); 14] = [
@@ -243,7 +244,6 @@ pub mod hmc7043 {
         spi_setup();
         info!("loading configuration...");
 
-        write(0x3, 0x10);  // Disable SYSREF timer
         write(0xA, 0x06);  // Disable the REFSYNCIN input
         write(0xB, 0x07);  // Enable the CLKIN input as LVPECL
         write(0x50, 0x1f); // Disable GPO pin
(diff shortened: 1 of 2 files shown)


artiq/gateware/targets/sayma_rtm.py
```diff
@@ -19,8 +19,11 @@
 from artiq import __version__ as artiq_version
 
 
-class CRG(Module):
+class CRG(Module, AutoCSR):
     def __init__(self, platform):
+
```
sbourdeauducq: Remove the empty line.

artiq/gateware/targets/sayma_rtm.py
```diff
@@ -110,6 +113,8 @@ def __init__(self, platform):
         csr_devices = []
 
         self.submodules.crg = CRG(platform)
+        csr_devices.append("crg")
```
sbourdeauducq: Why not use the GPIO out core instead of adding this to the CRG? CRG is normally used to clock the local design only.

sbourdeauducq:
This will need a hardware fix (or at least a review) as well, once the RTM FPGA is loaded by the AMC FPGA.
Maybe add a pull-up on the reset line? Though Xilinx FPGAs sometimes have weak pull-ups when unconfigured (depending on the state of certain pins) so maybe nothing needs to be changed.

artiq/firmware/libboard_artiq/hmc830_7043.rs
```diff
@@ -224,10 +225,17 @@ pub mod hmc7043 {
         Ok(())
     }
 
-    pub fn shutdown() -> Result<(), &'static str> {
+    pub fn enable() -> Result<(), &'static str> {
+        info!("enabling hmc7043");
+
+        unsafe {
+        csr::crg::hmc7043_rst_write(0);
```
sbourdeauducq: Indent

sbourdeauducq: And there is no time for the HMC7043 to wreak havoc between the release of its reset line and the SPI write that mutes it?

artiq/gateware/targets/sayma_rtm.py
hartytp: Lack of familiarity with misoc/to the man with a hammer everything is a nail.

hartytp: Do you mind changing this when merging?

hartytp:
> This will need a hardware fix (or at least a review) as well, once the RTM FPGA is loaded by the AMC FPGA.
> Maybe add a pull-up on the reset line?

hmmm...yes. That's a good idea. Let's do that.

gkasprow:
We can add negator that would invert the RESET line. It was foreseen in
Sayma rev2

PR #1054, opened 2018-06-05
Added 'unset' method to I2C switch

marmeladapk:


jordens:
IMHO it would be better to seize the opportunity to change the API into a single method `enable(bitmask)` (`set(channel)`: `enable(1 << channel)` for backwards compat). That's less code for more feature and additionally represents the underlying protocol.

PR #1069, opened 2018-06-11
Sayma: disable unused HMC7043 outputs

hartytp:
We've seen several cases where noise on the HMC7043 outputs can disrupt the AMC FPGA in subtle ways -- even if there is no active logic clocked from them.

This PR disables all HMC7043 outputs that are not currently used to prevent them from causing problems.


jbqubit:
Good idea @hartytp

PR #1082, opened 2018-06-26
Update installing.rst to reflect changes in OpenOCD rules file

apatura-iris:
The file 99-openocd.rules as downloaded from githubusercontent.com seems to be outdated and does not work on Ubuntu 16.04 (access error). The version that ships with OpenOCD has an additional ``TAG+="uaccess"`` in the rules file and works fine. Thus I suggest to use the file that is bundled with OpenOCD.


apatura-iris:
Good point. I hope the added comment clarifies this.

PR #1105, opened 2018-07-11
[NFC] Sayma: hmc7043 add explanation of input buffer configuration, and do …

hartytp:
…not needlessly reprogram CLKIN config with its default value.

The comment `Enable the CLKIN input as LVPECL` was also potentially misleading, given that "LVPECL mode" in the data sheet means "DC-coupled LVPECL", which we don't want to use.


sbourdeauducq:
Did you test this?

hartytp:
Building atm, but should not have any functional change (the register write I removed only set the default value).

sbourdeauducq:
Sure, but HMC chips have a track record of registers not being set to their documented default values upon reset, among other examples of awful behavior.

hartytp:
AKC. There is no hurry to merge this, just wanted to change the docs while it's fresh in my mind. I'll let you know when I've tested it.

Issue #1113, opened 2018-07-19
TTL[In]Out methods using *_previous_timestamp are incompatible with DMA


klickverbot:
(Failing) test case here: https://github.com/m-labs/artiq/pull/1112

This is the loopback test, but with the output events (including the TTL counter sensitivity settings) generated via DMA. `count()` after the sequence returns incorrect values.

Another test case with debug output, where exactly the same sequence without DMA works without issues: https://gist.github.com/klickverbot/b274840ce8e2372d406767a41e20140f.

klickverbot:
The timeline cursor seems to be advanced correctly: https://gist.github.com/klickverbot/c50e6e89a0752ae809ba5dc1caa4436c

sbourdeauducq:
What about the analyser, anything suspicious there? The raw records should contain the gate events.

klickverbot:
Oh, the problem here is much more pedestrian: DMA of course doesn't advance `TTLInOut.i_previous_timestamp`, so `count()` still uses the timeout from when the sequence was recorded (i.e. `gate_*` was actually called). Same would go for output events and `.sync()`, although that is much less frequently used.

The fact that three experienced devs looked at this and failed to notice the issue, obvious as it may be in retrospect, probably indicates that we have an usability issue here. However, I'm not sure what to do about this – "attribute writeback" for DMA would be rather nontrivial to implement, and a questionable design in the first place (especially across kernels).

An example illustrating this in the DMA docs will certainly help to some extent, but any other ideas on how to address this?

jordens:
Maybe invalidate `i_previous_timestamp` in `count()` (and elsewhere) to prevent it being used wrong in the DMA case and describe how to manually compute it and set it when doing `count()` ing a DMA gate. Or change the TTL api to enforce returning the gate closure time from `gate()` and require using that in `count()` (and elsewhere).

klickverbot:
Requiring a timestamp argument to `count` is perhaps the cleanest way to solve this. I was going to add this anyway to be able to write correct user code without fiddling with `i_previous_timestamp` or directly using `rtio_input_timestamp`, but we could remove the no-timestamp version with the next major release.

We should probably ditch `sync()` entirely as well.

jordens:
I'd like to have `sync_mu(t_mu)` as a method of the core device though. It's a frequent pattern.

klickverbot:
Indeed, and currently a rather clunky one (manually writing out the loop), unless I've missed something.

[...]

Issue #1007, opened 2018-05-21
llvm/compiler: FP related slowness


jordens:
Fails unittests: http://buildbot.m-labs.hk/builders/artiq/builds/2364/steps/python_unittest_2/logs/stdio

Presumably some constant propagation and folding is not happening anymore. The `test_clock_generator_loopback` (`test_rtio.CoredeviceTest`) and the `1*MHz` line.
First failing build (AFAICT) was http://buildbot.m-labs.hk/builders/artiq/builds/2344/steps/python_unittest_2/logs/stdio (rust update).

The `test_host_to_device` test has become slower **after** this one.

whitequark:
Yes. The root cause is I removed the unconditionally_dereferenceable LLVM patch because it was unsound while upgrading to LLVM 6.0. I am reimplementing that in the ARTIQ compiler now.

sbourdeauducq:
Any update?

whitequark:
Yes. I tried to fix this purely in the ARTIQ compiler, and it didn't work. Specifically, hoisting invariant loads requires inlining, which requires devirtualization, which is quite hard to implement due to Python's semantics. (We had devirtualization to support compiler-assisted interleaving, but it broke a while ago, and I wasn't successful in fixing it).

A proper solution here would probably be to completely change the semantics of ARTIQ Python to make it more amenable for analysis, but that's not going to fly in the 3.x branch, so now I'm looking at the LLVM side of things. A loadable pass is a nightmare to debug, so I think I'll add an LLVM and llvmlite patches to conda to add a pass to LLVM globally.

sbourdeauducq:
@whitequark ping

sbourdeauducq:
After this gets fixed, we should add a device unittest for this. The clock generator loopback test is not a good one, and I made it pass in d2c8e62cb71fea888af1d71f3d9e953840b6cdb4 to stop polluting the CI results, which had been going on for way too long @whitequark.

jordens:
also c.f. #686 for tracking various aspects of coredevice performance

sbourdeauducq:
@whitequark can you give us an update?

sbourdeauducq:
With all the time that this painful issue is taking, it seems that meanwhile a number of people have been using development versions that have the problem. How much of an impact does it actually have? Does it make sense to delay releases further? @cjbe @dhslichter @jordens

jordens:
[...]

PR #1115, opened 2018-07-24
ttl: Stop tracking function call cursor positions

klickverbot:
The motivation behind this change is that the current API doesn't play well with DMA as detailed in #1113 – RTIO events generated by DMA can't set the timestamp attributes.

The only part of the change strictly necessary to fix this are the optional timestamp arguments, so that code using DMA can manually specify these. However, as pointed out in #1113, the current API is unnecessarily error-prone, especially as DMA is deployed more and more widely (and possibly hidden behind layers of abstraction).

Hence, I removed `TTLOut.sync` altogether, since it seems to be used only very rarely (there are zero uses in `artiq`, as well as all our experiment code). The input methods, on the other hand, are used much more often, so 67fe87a makes them default to `now_mu()` when no timestamps are passed explicitly. Requiring explicit timestamps might well be the cleaner design, but this seems to be a sensible choice in terms of backwards compatibility for now.

There are also two commits adding convenience functions/return values as suggested by @jordens.

Since this is a very central piece of the realtime API, I'd appreciate comments on the design – I'll update the tests once a design has been agreed on.

### DIFF
--- a/artiq/coredevice/core.py
+++ b/artiq/coredevice/core.py
@@ -151,6 +151,15 @@ class Core:
 
     @kernel
     def get_rtio_counter_mu(self):
+        """Retrieve the current value of the hardware RTIO timeline counter.
+
+        As the timing of kernel code executed on the CPU is inherently
+        non-deterministic, the return value is by necessity only a lower bound
+        for the actual value of the hardware register at the instant when
+        execution resumes in the caller.
+
+        For a more detailed description of these concepts, see :doc:`/rtio`.
+        """
         return rtio_get_counter()
 
     @kernel
--- a/artiq/coredevice/core.py
+++ b/artiq/coredevice/core.py
@@ -154,6 +154,17 @@ class Core:
         return rtio_get_counter()
 
     @kernel
+    def wait_until_mu(self, cursor_mu):
+        """Block execution until the hardware RTIO counter reaches the given
+        value (see :meth:`get_rtio_counter_mu`).
+
+        If the hardware counter has already passed the given time, the function
+        returns immediately.
+        """
+        while self.get_rtio_counter_mu() < cursor_mu:
+            pass
+
+    @kernel
     def get_drtio_link_status(self, linkno):
         """Returns whether the specified DRTIO link is up.
 
--- a/artiq/coredevice/core.py
+++ b/artiq/coredevice/core.py
@@ -134,7 +134,7 @@ class Core:
 
     @portable
     def seconds_to_mu(self, seconds):
-        """Converts seconds to the corresponding number of machine units
+        """Convert seconds to the corresponding number of machine units
         (RTIO cycles).
 
         :param seconds: time (in seconds) to convert.
@@ -143,7 +143,7 @@ class Core:
 
     @portable
     def mu_to_seconds(self, mu):
-        """Converts machine units (RTIO cycles) to seconds.
+        """Convert machine units (RTIO cycles) to seconds.
 
         :param mu: cycle count to convert.
         """
@@ -166,7 +166,7 @@ class Core:
 
     @kernel
     def get_drtio_link_status(self, linkno):
-        """Returns whether the specified DRTIO link is up.
+        """Return whether the specified DRTIO link is up.
 
         This is particularly useful in startup kernels to delay
         startup until certain DRTIO links are up."""
--- a/artiq/coredevice/ttl.py
+++ b/artiq/coredevice/ttl.py
@@ -35,9 +35,6 @@ class TTLOut:
         self.core = dmgr.get(core_device)
         self.channel = channel
 
-        # in RTIO cycles
-        self.o_previous_timestamp = numpy.int64(0)
-
     @kernel
     def output(self):
         pass
@@ -45,14 +42,6 @@ class TTLOut:
     @kernel
     def set_o(self, o):
         rtio_output(now_mu(), self.channel, 0, 1 if o else 0)
-        self.o_previous_timestamp = now_mu()
-
-    @kernel
-    def sync(self):
-        """Busy-wait until all programmed level switches have been
-        effected."""
-        while self.core.get_rtio_counter_mu() < self.o_previous_timestamp:
-            pass
 
     @kernel
     def on(self):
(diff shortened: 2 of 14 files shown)


artiq/coredevice/ttl.py
```diff
@@ -206,88 +182,124 @@ def pulse(self, duration):
...
         self._set_sensitivity(3)
         delay(duration)
         self._set_sensitivity(0)
+        return now_mu()
 
     @kernel
-    def count(self):
-        """Poll the RTIO input during all the previously programmed gate
-        openings, and returns the number of registered events.
+    def count(self, up_to_timestamp_mu=-1):
```
jordens: I suspect that default may need to be `int64(-1)`.

artiq/coredevice/ttl.py
```diff
@@ -206,88 +182,124 @@ def pulse(self, duration):
...
+        """Poll the RTIO input up to the specified timestamp, and returns the
+        number of registered events.
 
-        This function does not interact with the time cursor."""
+        If no timestamp is specified, the current cursor position is used. This
+        unnecessarily reduces the slack if the last gate closed further in the
+        past; prefer specifying a timestamp.
+
+        This function does not modify the timeline cursor."""
```
jordens: Maybe add an example here for the frequent use case: `channel.count(channel.gate_rising(10*us))`.

And if this resolves the original issue with DMA, maybe an example how to code that as well.

artiq/coredevice/ttl.py
jordens: Bikeshed names and API: `count_mu(until_mu)` and additionally `count(until)` in seconds?

artiq/coredevice/ttl.py
```diff
@@ -206,88 +182,124 @@ def pulse(self, duration):
...
         count = 0
-        while rtio_input_timestamp(self.i_previous_timestamp, self.channel) >= 0:
+        while rtio_input_timestamp(up_to_timestamp_mu, self.channel) >= 0:
             count += 1
         return count
 
     @kernel
-    def timestamp_mu(self):
+    def timestamp_mu(self, up_to_timestamp_mu=-1):
         """Poll the RTIO input and returns an event timestamp (in machine
```
jordens: While we are here: Maybe `returns the first event timestamp` to clarify which and add a sentence saying that at most one input event is consumed.

artiq/coredevice/ttl.py
```diff
@@ -426,10 +435,3 @@ def set(self, frequency):
     def stop(self):
         """Stop the toggling of the clock and set the output level to 0."""
         self.set_mu(0)
-
-    @kernel
-    def sync(self):
```
jordens: IMHO this merits an entry in `RELEASE_NOTES.md` for all the brekaing changes.

artiq/coredevice/core.py
```diff
@@ -153,6 +153,17 @@ def mu_to_seconds(self, mu):
     def get_rtio_counter_mu(self):
         return rtio_get_counter()
 
+    @kernel
+    def wait_until_rtio_counter_mu(self, cursor_mu):
```
jordens: Bikeshed: Maybe `wait_until_mu(timestamp_mu)` and maybe add `wait_until(timestamp)` in seconds.

jordens:
Thanks. Looking good.

* [ ] Release notes
* [ ] Examples (if there is anything to be modified there)

artiq/coredevice/ttl.py
klickverbot: Regarding the cursors, is there any place where non-`mu` units are used? I'd like to avoid creating potential for confusion between durations and absolute timestamps.

jordens: Oh right. Yes. the absolute timestamp should be `mu`-only.

jordens:
ping @sbourdeauducq on the naming and api changes

artiq/coredevice/core.py
sbourdeauducq: Agreed

sbourdeauducq: Except we should not have ``wait_until``- all absolute timestamps in ARTIQ are in mu.

artiq/coredevice/ttl.py
sbourdeauducq: Should the frequent use case be ``channel.count(channel.gate_rising(10*us))`` or ``channel.gate_rising(10*us); channel.count()``?

Since there is the short form ``channel.count(channel.gate_rising(10*us))`` already, it is not clear to me if having a default value for ``up_to_timestamp`` is useful.

sbourdeauducq:
Fine except for what has already been said.

artiq/coredevice/ttl.py
jordens: I'm ok with dropping the default for `up_to_timestamp`. That would also prevent a few more cases of the original problem of silently assuming that the count end time can be handled automatically (in DMA or with multiple gates).

sbourdeauducq:
@klickverbot Can you finish this before 4.0? (Which should be soon and mostly depends on https://github.com/m-labs/artiq/issues/1007)

sbourdeauducq:
@klickverbot Can you add the timeout compensations to this pull request, and cherry-pick c83e22c11cf229ca20b26d41168249e13773d0ff?

artiq/coredevice/ttl.py
klickverbot: Done.

artiq/coredevice/core.py
```diff
@@ -143,19 +143,39 @@ def seconds_to_mu(self, seconds):
...
+        non-deterministic, the return value is by necessity only a lower bound
+        for the actual value of the hardware register at the instant when
+        execution resumes in the caller.
+
+        For a more detailed description of these concepts, see :doc:`/rtio`.
+        """
         return rtio_get_counter()
 
+    @kernel
+    def wait_until(self, cursor_mu):
```
klickverbot: Should this be `wait_until_mu`? Can't decide which is more regular/least surprising.

sbourdeauducq: > Should this be `wait_until_mu`?

Yes, I think so.

sbourdeauducq:
@klickverbot Thanks, looks great!

PR #1122, opened 2018-08-02
share moninj injection state between dashboards

cjbe:
Previously if one dashboard overrode a channel this was not visible on any other dashboard - the channel appeared to operate under experiment control, and the actual output level was unrelated to the displayed output level.

This PR allows clients to be notified by the core device when the injection state is changed. With the current implementation the modifying client receives a notification of its change up to 200ms after the change, which means that the dashboard moninj panel cannot be used to generate events faster than 200ms.

### DIFF
--- a/artiq/coredevice/comm_moninj.py
+++ b/artiq/coredevice/comm_moninj.py
@@ -56,15 +56,15 @@ class CommMonInj:
         self._writer.write(packet)
 
     def monitor_injection(self, enable, channel, overrd):
-        packet = struct.pack(">bblb", 1, enable, channel, overrd)
+        packet = struct.pack(">bblb", 3, enable, channel, overrd)
         self._writer.write(packet)
 
     def inject(self, channel, override, value):
-        packet = struct.pack(">blbb", 2, channel, override, value)
+        packet = struct.pack(">blbb", 1, channel, override, value)
         self._writer.write(packet)
 
     def get_injection_status(self, channel, override):
-        packet = struct.pack(">blb", 3, channel, override)
+        packet = struct.pack(">blb", 2, channel, override)
         self._writer.write(packet)
 
     async def _receive_cr(self):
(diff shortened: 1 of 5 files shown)


sbourdeauducq:
> With the current implementation the modifying client receives a notification of its change up to 200ms after the change, which means that the dashboard moninj panel cannot be used to generate events faster than 200ms.

This was already a limitation of the previous UDP code iirc.

artiq/coredevice/comm_moninj.py
```diff
@@ -51,16 +51,20 @@ def __init__(self, monitor_cb, injection_status_cb, disconnect_cb=None):
...
         packet = struct.pack(">bblb", 0, enable, channel, probe)
         self._writer.write(packet)
 
+    def monitor_injection(self, enable, channel, overrd):
+        packet = struct.pack(">bblb", 1, enable, channel, overrd)
+        self._writer.write(packet)
+
     def inject(self, channel, override, value):
-        packet = struct.pack(">blbb", 1, channel, override, value)
+        packet = struct.pack(">blbb", 2, channel, override, value)
```
sbourdeauducq: Can we avoid breaking compatibility with the current protocol?

sbourdeauducq: (to handle a dashboard/firmware version mismatch gracefully)

cjbe: OK - permuting the packet numbering would allow new firmware with the old dashboard. New dashboard and old firmware would always break things though (firmware aborts connection on unknown packet type).

sbourdeauducq: OK, that's acceptable.

sbourdeauducq: This also prevents any unexpected injections at dashboard startup.

sbourdeauducq:
Thanks!

PR #1132, opened 2018-08-13
nix: separated buildtime & runtime dependencies

sjmackenzie:


sjmackenzie:
Once (if) you've updated all the code you should remove this [line](https://github.com/m-labs/artiq/pull/1132/files#diff-4d842add0dac68884b9567d925f39b66R125). This'll run your tests automatically upon building of every `artiq` derivation, for now it fails.

PR #1136, opened 2018-08-19
test/coredevice: Add input gate/timestamp test

klickverbot:
Expected to be broken based on local tests.

### DIFF
--- a/artiq/test/coredevice/test_rtio.py
+++ b/artiq/test/coredevice/test_rtio.py
@@ -189,6 +189,52 @@ class LoopbackCount(EnvExperiment):
         self.set_dataset("count", self.loop_in.count())
 
 
+class LoopbackGateTiming(EnvExperiment):
+    def build(self):
+        self.setattr_device("core")
+        self.setattr_device("loop_in")
+        self.setattr_device("loop_out")
+
+    @kernel
+    def run(self):
+        # Make sure there are no leftover events.
+        self.core.reset()
+
+        # Determine loop delay.
+        self.core.break_realtime()
+        with parallel:
+            self.loop_in.gate_rising_mu(10000)
+            with sequential:
+                delay_mu(5000)
+                out_mu = now_mu()
+                self.loop_out.pulse_mu(1000)
+        in_mu = self.loop_in.timestamp_mu()
+        if in_mu < 0:
+            raise PulseNotReceived("No input event, loop connected?")
+        loop_delay_mu = in_mu - out_mu
+
+        # With the exact delay known, make sure tight gate timings work.
+        # In the most common configuration, 24 mu == 24 ns == 3 coarse periods,
+        # which should be plenty of slack.
+        delay_mu(10000)
+
+        gate_start_mu = now_mu()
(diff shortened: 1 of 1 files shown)


artiq/test/coredevice/test_rtio.py
```diff
@@ -189,6 +189,53 @@ def run(self):
...
+    def build(self):
+        self.setattr_device("core")
+        self.setattr_device("loop_in")
+        self.setattr_device("loop_out")
+
+    @kernel
+    def run(self):
+        # Make sure there are no leftover events.
+        self.core.break_realtime()
+        self.loop_in.count()
```
sbourdeauducq: There may be leftover events because of the very bug you are testing against (due to the gateware latency in the timestamp counter to gate path, the timeout - referenced to the timestamp counter - is too small and the kernel thinks the gateware has closed the gate when it has not). Use ``core.reset()`` instead.

artiq/test/coredevice/test_rtio.py
```diff
@@ -189,6 +189,52 @@ def run(self):
...
+        self.setattr_device("loop_in")
+        self.setattr_device("loop_out")
+
+    @kernel
+    def run(self):
+        # Make sure there are no leftover events.
+        self.core.reset()
+
+        # Determine loop delay.
+        self.core.break_realtime()
```
sbourdeauducq: ``reset()`` already does ``break_realtime()``.

artiq/test/coredevice/test_rtio.py
```diff
@@ -189,6 +189,52 @@ def run(self):
...
+        at_mu(out_mu)
+        self.loop_out.pulse_mu(1000)
+
+        in_mu = self.loop_in.timestamp_mu()
+        if in_mu < 0:
+            raise PulseNotReceived("No input event")
+        if not (gate_start_mu < in_mu < gate_end_mu):
+            raise PulseNotReceived("Input event should occur during gate")
+        if not (-2 < (in_mu - out_mu - loop_delay_mu) < 2):
+            raise PulseNotReceived("Loop delay should not change")
```
sbourdeauducq: PulseNotReceived?

sbourdeauducq: What about IncorrectPulseTiming?

artiq/test/coredevice/test_rtio.py
```diff
@@ -189,6 +189,52 @@ def run(self):
...
+
+        out_mu = gate_start_mu - loop_delay_mu
+        at_mu(out_mu)
+        self.loop_out.pulse_mu(1000)
+
+        in_mu = self.loop_in.timestamp_mu()
+        if in_mu < 0:
+            raise PulseNotReceived("No input event")
+        if not (gate_start_mu < in_mu < gate_end_mu):
+            raise PulseNotReceived("Input event should occur during gate")
```
sbourdeauducq: Right now there is an uncompensated latency offset between the gate and the input. So, the received timestamps can be slightly outside the gate. This offset also causes the pulse not to be received above.

PR #1145, opened 2018-08-31
dashboard: fix TTL moninj level display

cjbe:
Previously receiving an injection status callback for the TTL level set the displayed level, even if the channel was not overridden.

### DIFF
--- a/artiq/dashboard/moninj.py
+++ b/artiq/dashboard/moninj.py
@@ -74,6 +74,7 @@ class _TTLWidget(QtWidgets.QFrame):
         self.cur_level = False
         self.cur_oe = False
         self.cur_override = False
+        self.cur_override_level = False
         self.refresh_display()
 
     def enterEvent(self, event):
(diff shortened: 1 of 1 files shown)


sbourdeauducq:
Is it normal for the dashboard to receive such a callback if the channel was not overridden?
I have some trouble understanding what problem that fixes exactly and why this is the correct solution.

cjbe:
The moninj poll loop in the firmware first broadcasts all changed status probes, then all changed injection probes.
This is an issue when the dashboard first connects. On dashboard startup, we subscribe to status updates and injection updates on all TTL channels. At the next firmware poll time we then receive all the probe status updates before the injection status updates. 

Thus we receive callbacks setting the status probe OE and level, then receive callbacks setting the override enable and level. Without this patch, the override level callback changes the displayed output level.

There are several ways of fixing this, but the most robust seems to be ensuring that the override level state is recorded separately, and only displayed when the output override is enabled.

sbourdeauducq:
Thanks

PR #1153, opened 2018-09-17
firmware: Use larger ARP cache

### LINKED ISSUE #1150: ConnectionResetError on TCP Keepalive
On certain experiments I see a significant rate of ConnectionResetErrors (up to one every 10 minutes for some use-cases).

Here is a [packet dump](https://github.com/m-labs/artiq/files/2378883/crash_3.zip) and a [trace core log](https://hastebin.com/igicojokax.php) of an typical error - the error occurs at core log timestamp [449493.322821s] == packet dump timestamp 23:01:29.962904 (packet 98). The core device is 10.255.6.210 and the master is running on 10.255.6.191. The core device connection on 1381 is idle for 3s so a keep-alive is sent. The core device responds to the keepalive, but also records a timeout and kills the connection. It looks like this is because 2.9s before the error (449490.397252) the master ARP entry (.191) is evicted from the ARP cache.

I have several more packet dumps (but without core trace logs). All the errors occur after a keepalive when the Kasli sends a RST/ACK.

I am using a Kasli with gateware and firmware build off c9d8bd15669139b97b42ee830081d2d8ca3eee02
This may be related to #1125

klickverbot:
This works around a problematic interaction between ARP cache
expiry in smoltcp (with its 3 seconds timeout before a discovery
request is sent) and our TCP keepalive settings, where the timeout
is reached before the keepalive had a chance to be sent.

GitHub: Closes #1150.

### DIFF
--- a/artiq/firmware/runtime/main.rs
+++ b/artiq/firmware/runtime/main.rs
@@ -270,9 +270,8 @@ fn startup_ethernet() {
         smoltcp::phy::EthernetTracer::new(net_device, net_trace_fn)
     };
 
-    let mut neighbor_map = [None; 8];
     let neighbor_cache =
-        smoltcp::iface::NeighborCache::new(&mut neighbor_map[..]);
+        smoltcp::iface::NeighborCache::new(alloc::btree_map::BTreeMap::new());
     let mut interface  =
         smoltcp::iface::EthernetInterfaceBuilder::new(net_device)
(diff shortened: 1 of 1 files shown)


whitequark:
This wasn't done before because smoltcp had no support for evicting entries from owned cache, but since https://github.com/m-labs/smoltcp/pull/234, this happens just fine.

PR #1158, opened 2018-09-25
Kasli DRTIO master: do not init rtio core until serdes PLL has locked.

hartytp:
fixes 1155

@jordens if you still have your test setup intact, please could you test this on your setup?

@sbourdeauducq do we need to apply similar fixes anywhere else?

### DIFF
--- a/artiq/firmware/runtime/rtio_mgt.rs
+++ b/artiq/firmware/runtime/rtio_mgt.rs
@@ -1,3 +1,4 @@
+use clock;
 use board_misoc::csr;
 #[cfg(has_rtio_clock_switch)]
 use board_misoc::config;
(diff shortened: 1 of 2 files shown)


hartytp:
NB this will break any DRTIO masters which don't have an RTIO clock multiplier with a `pll_locked` CSR ( just Sayma master AFACIT). @sbourdeauducq do you want me to add a config option, something like `HAS_RTIO_MUL` to feature gate the new firmware?

jordens:
Not right now. In a couple of days.

hartytp:
@jordens thanks, there is no particular hurry for this as far as I'm concerned. Just something I was looking at while doing acceptance testing for drtio switching.

sbourdeauducq:
I'll sort it out. Thanks for hunting down this bug.

sbourdeauducq:
b92350b0

PR #1170, opened 2018-10-08
Urukul: support hw rev v1.3

hartytp:
- The 0x08 (hardware revisions <=1.2) and 0x09 (hardware revision v1.3) are essentially compatible. The only difference being that the v1.3 revision has an extra clock mux
- Clocking options for new hw rev: 100MHz oscillator 0x00; SMA 0x01 (oscillator powered down); MMCX 0x02 (oscillator powered down).
- In the 0x08 CPLD code the clocking was: 0x00 100MHz XO/MMCX (choice by population options); 0x01 SMA (oscillator not disabled by software). 
- We check that the new clocking options aren't used with the old CPLD revision


hartytp:
NB tested in the lab using v1.3 hardware, but I don't currently have access to any <=v1.2 hardware for testing

artiq/coredevice/urukul.py
```diff
@@ -198,8 +201,11 @@ def init(self, blind=False):
             self.cfg_write(self.cfg_reg)
         else:
             proto_rev = urukul_sta_proto_rev(self.sta_read())
-            if proto_rev != STA_PROTO_REV_MATCH:
+            if proto_rev not in STA_PROTO_REV_MATCH:
                 raise ValueError("Urukul proto_rev mismatch")
+            if bool(cfg & CFG_CLK_SEL1) \
+               and proto_rev == STA_PROTO_REV_MATCH[0]:
+                raise ValueError("Unsupported clocking option")
```
jordens: That's the question. Is it worth trying to unify the cpld gateware over v1.2 and v1.3, not icrement the protocol and ignore the missing functionality on v1.2 **or** should we be strict and add the complexity of supporting multiple protocol versions and different gateware for different hardware.
I am leaning towards the latter (and documenting the NOP-ness of clk_sel1 on v1.2/earlier).

jordens:
https://github.com/quartiq/urukul/pull/5 for the CPLD gateware changes.

artiq/coredevice/urukul.py
hartytp: > I am leaning towards the latter (and documenting the NOP-ness of clk_sel1 on v1.2/earlier).

"latter" or "former"?

> I am leaning towards the latter (and documenting the NOP-ness of clk_sel1 on v1.2/earlier).

Where do you want this documented? Just the constructor?

hartytp: OK, let's do that.

I'll update the PR later today

hartytp:
updated
- changed the osc/mmcx cfg bit to match the cpld revision
- sticking with 0x08 cpld revision, no attempt to check for use of unsupported clocking options
- added docs

Will test on hw in the morning

jordens:
Thanks.

jordens:
I checked v1.3 with `clk_sel=0x00`. This would be a nice entry point for CI. The hardware is already set up.

hartytp:
Checked all three clocking options with v1.3 hardware. All work as expected, including oscillator power down.

PR #1176, opened 2018-10-13
Various sync_struct refactors

klickverbot:
These are some refactorings from back when I first looked at sync_struct and related code, lumped together into one PR for CI testing – see individual commits for details.

The only potentially contentious change is the renaming of `Notifier.read`, which I always found difficult to make sense of – Is it a property that `read()`s in new data? Or the previously read data? Etc. Alternative bikeshed colour suggestions welcome.

### DIFF
--- a/artiq/master/databases.py
+++ b/artiq/master/databases.py
@@ -1,7 +1,7 @@
 import asyncio
 import tokenize
 
-from artiq.protocols.sync_struct import Notifier, process_mod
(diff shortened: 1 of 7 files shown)


sbourdeauducq:
LGTM as well but breaks unittests on the buildbot for some reason.

jordens:
No strong preference from me. Your choice.

PR #1178, opened 2018-10-15
Tpz

pathfinder49:
Thorlabs Tcubes fixes:
- flake8
- fix TPZ constructor after move to asyncio
- Tcube fix docummentation in set_channel_enable_state

### DIFF
--- a/artiq/devices/thorlabs_tcube/driver.py
+++ b/artiq/devices/thorlabs_tcube/driver.py
@@ -219,7 +219,7 @@ class _Tcube:
     async def set_channel_enable_state(self, activated):
         """Enable or Disable channel 1.
 
-        :param activated: 1 to enable channel, 2 to disable it.
+        :param activated: 1 to enable channel, 0 to disable it.
         """
 
         if activated:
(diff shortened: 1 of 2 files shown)


hartytp:
We considered both approaches and felt that, in this particular case, putting a 'run_until_complete' in the constructor was cleaner than making the user call some other method after construction but before any voltage setting calls. Is there some issue with our approach that I'm not seeing? Otherwise I guess we set the voltage limit to 0 in the constructor, document the need to query the voltage limit after construction and add a 'run_until_complete' to the controller.

sbourdeauducq:
Yes: ``get_event_loop()`` may create an event loop. If this happens before the user sets an alternate event loop (required e.g. on Windows with ``ProactorEventLoop`` in some cases) we are in trouble.

pathfinder49:
Sorry, didn't mean to push that last bit. Will fix tomorrow.

jordens:
A nit: is there no way to abbreviate the cross references and use relative ones? IIRC you can drop the module and class path if the crossref is local.

pathfinder49:
Thanks for the feedback. Fixed and ready for review.

sbourdeauducq:
Thanks!

PR #1179, opened 2018-10-19
[typo]: fix "addresses" in simple_network_args

drewrisinger:
the "--bind" argument's help misspelled "addresses" as "addresse". Fixed


sbourdeauducq:
Should be singular. Fixed.

PR #1185, opened 2018-10-25
[artiq_client]: Remove custom input validation for built-in argparse

drewrisinger:
# ARTIQ Pull Request

## Description of Changes
Simplify input verification (command-line argument) logic by using built-ins from `argparse`.

### Related Issue
None

## Type of Changes

|   | Type |

### All Pull Requests

### Code Changes

  - NOTE: fixed `flake8` errors from pre-existing code too.
  - Tested by running commands from command line: `artiq_client show fake`, `artiq_client show devices`, `artiq_client set-dataset -p -n`

### Git Logistics

  ```
  [topic]: description. < 50 characters total.
  ```

### Licensing

### DIFF
--- a/artiq/frontend/artiq_client.py
+++ b/artiq/frontend/artiq_client.py
@@ -2,8 +2,8 @@
 """
 Client to send commands to :mod:`artiq_master` and display results locally.
 
-The client can perform actions such as accessing/setting data, scanning devices,
-scheduling experiments, and looking for experiments/devices.
+The client can perform actions such as accessing/setting datasets, 
+scanning devices, scheduling experiments, and looking for experiments/devices.
 """
 
 import argparse
--- a/artiq/frontend/artiq_client.py
+++ b/artiq/frontend/artiq_client.py
@@ -15,7 +15,7 @@ from artiq.protocols.pc_rpc import Client
 from artiq.protocols.sync_struct import Subscriber
 from artiq.protocols.broadcast import Receiver
 from artiq.protocols import pyon
-from artiq.tools import short_format
+from artiq.tools import short_format, parse_arguments
 
 
 def clear_screen():
@@ -46,7 +46,8 @@ def get_argparser():
                                  "scheduling, default: %(default)s)")
     parser_add.add_argument("-t", "--timed", default=None, type=str,
                             help="set a due date for the experiment")
-    parser_add.add_argument("-f", "--flush", default=False, action="store_true",
+    parser_add.add_argument("-f", "--flush", default=False,
+                            action="store_true",
                             help="flush the pipeline before preparing "
                             "the experiment")
     parser_add.add_argument("-R", "--repository", default=False,
@@ -80,10 +81,12 @@ def get_argparser():
                                     help="name of the dataset")
     parser_set_dataset.add_argument("value", metavar="VALUE",
                                     help="value in PYON format")
-    parser_set_dataset.add_argument("-p", "--persist", action="store_true",
-                                    help="make the dataset persistent")
-    parser_set_dataset.add_argument("-n", "--no-persist", action="store_true",
-                                    help="make the dataset non-persistent")
+    # Allow only one of --persist or --no-persist arguments
+    persist_group = parser_set_dataset.add_mutually_exclusive_group()
+    persist_group.add_argument("-p", "--persist", action="store_true",
+                               help="make the dataset persistent")
+    persist_group.add_argument("-n", "--no-persist", action="store_true",
+                               help="make the dataset non-persistent")
 
     parser_del_dataset = subparsers.add_parser(
         "del-dataset", help="delete a dataset")
@@ -93,7 +96,8 @@ def get_argparser():
         "show", help="show schedule, log, devices or datasets")
     parser_show.add_argument(
         "what", metavar="WHAT",
-        help="select object to show: schedule/log/devices/datasets")
+        choices=["schedule", "log", "ccb", "devices", "datasets"],
+        help="select object to show: %(choices)s")
 
     subparsers.add_parser(
         "scan-devices", help="trigger a device database (re)scan")
@@ -114,17 +118,9 @@ def get_argparser():
     return parser
 
 
-def _parse_arguments(arguments):
-    d = {}
-    for argument in arguments:
-        name, value = argument.split("=")
-        d[name] = pyon.decode(value)
-    return d
-
-
 def _action_submit(remote, args):
     try:
-        arguments = _parse_arguments(args.arguments)
+        arguments = parse_arguments(args.arguments)
     except:
         print("Failed to parse run arguments")
         sys.exit(1)
@@ -154,10 +150,6 @@ def _action_delete(remote, args):
 
 
 def _action_set_dataset(remote, args):
-    if args.persist and args.no_persist:
-        print("Options --persist and --no-persist cannot be specified "
-              "at the same time")
-        sys.exit(1)
     persist = None
     if args.persist:
         persist = True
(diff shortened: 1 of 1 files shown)


artiq/frontend/artiq_client.py
```diff
@@ -298,9 +291,6 @@ def main():
             _show_dict(args, "devices", _show_devices)
         elif args.what == "datasets":
             _show_dict(args, "datasets", _show_datasets)
-        else:
```
sbourdeauducq: Better still raise an exception here to help catch program bugs. A simple ``raise ValueError`` will do.

artiq/frontend/artiq_client.py
```diff
@@ -80,10 +81,12 @@ def get_argparser():
                                     help="name of the dataset")
     parser_set_dataset.add_argument("value", metavar="VALUE",
                                     help="value in PYON format")
-    parser_set_dataset.add_argument("-p", "--persist", action="store_true",
-                                    help="make the dataset persistent")
-    parser_set_dataset.add_argument("-n", "--no-persist", action="store_true",
-                                    help="make the dataset non-persistent")
+    # Allow only one of --persist or --no-persist arguments
```
sbourdeauducq: This is already pretty clear from the name of the function, ``add_mutually_exclusive_group``.

drewrisinger: Ok. So it is an extra comment to make things just slightly more explicit.

artiq/frontend/artiq_client.py
drewrisinger: That is taken care of by `choices=[...]` in line 99:
https://github.com/m-labs/artiq/blob/b5a5be4ac8dfb10846ca413a91e8223429839197/artiq/frontend/artiq_client.py#L97-L100
https://docs.python.org/3/library/argparse.html#choices

artiq/frontend/artiq_client.py
klickverbot: I'd categorise this as (mild) code smell – comments like these add cognitive overhead to express something that is already obvious from just reading the code as plain English.

ARTIQ as a whole is definitely underdocumented, and I very much appreciate the effort (cf. some of my own recent commits), but it's a different sort of comments we need.

artiq/frontend/artiq_client.py
sbourdeauducq: Sure. But if ``choices`` does not match this test (e.g. due to a typo) then there is a bug, and not raising an exception here will make it more obscure than it should be. Generally, it is good practice to always use the pattern ``if x=A: elif x=B: ... else: raise ValueError`` (short of a case statement in Python)

sbourdeauducq:
@drewrisinger Can you resolve the two small issues (remove overly verbose comment and use "switch" code pattern?)
BTW there are other places where ``choices`` could be used, e.g. in the targets. Do you want to submit other PRs?

artiq/frontend/artiq_client.py
drewrisinger: Resolved in https://github.com/m-labs/artiq/pull/1185/commits/01e4d70dc1e7b2df035382a3a44a3a064999d096

artiq/frontend/artiq_client.py
drewrisinger: Resolved in https://github.com/m-labs/artiq/pull/1185/commits/01e4d70dc1e7b2df035382a3a44a3a064999d096

artiq/frontend/artiq_client.py
```diff
@@ -278,6 +286,7 @@ def _show_ccb(args):
 
 
 def main():
+    """Submit requests to :mod:`artiq_master`, e.g. schedule, get data."""
```
sbourdeauducq: That's redundant with the module description.

artiq/frontend/artiq_client.py
```diff
@@ -291,6 +300,8 @@ def main():
             _show_dict(args, "devices", _show_devices)
         elif args.what == "datasets":
             _show_dict(args, "datasets", _show_datasets)
+        else:
+            raise ValueError("Unknown object to show; list valid names with -h")
```
sbourdeauducq: Wrong exception message. The user is never supposed to see this; if this code is reached it is a bug in the program and using ``-h`` is not going to hep.

artiq/frontend/artiq_client.py
```diff
@@ -19,13 +25,15 @@
...
 def clear_screen():
+    """Blank the terminal."""
     if os.name == "nt":
         os.system("cls")
     else:
         sys.stdout.write("\x1b[2J\x1b[H")
 
 
 def get_argparser():
+    """Return an argument parser with :mod:`artiq_client` commands."""
```
sbourdeauducq: Obvious.

artiq/frontend/artiq_client.py
```diff
@@ -1,4 +1,10 @@
 #!/usr/bin/env python3
+"""
+Client to send commands to :mod:`artiq_master` and display results locally.
+
+The client can perform actions such as accessing/setting data, scanning devices,
+scheduling experiments, and looking for experiments/devices.
+"""
```
sbourdeauducq: OK

artiq/frontend/artiq_client.py
```diff
@@ -1,4 +1,10 @@
 #!/usr/bin/env python3
+"""
+Client to send commands to :mod:`artiq_master` and display results locally.
+
+The client can perform actions such as accessing/setting data, scanning devices,
```
sbourdeauducq: "datasets" to use consistent terminology.

artiq/frontend/artiq_client.py
drewrisinger: Same error message as was there before
https://github.com/m-labs/artiq/blob/6df4ae934f07e57f6e61be1bafacf91179b40b01/artiq/frontend/artiq_client.py#L301-L303

sbourdeauducq: Sure, without ``choices``.

artiq/frontend/artiq_client.py
drewrisinger: Fixed in https://github.com/m-labs/artiq/pull/1185/commits/c3206d79c7b535bc7c48b5b19844a10077c9ee49

artiq/frontend/artiq_client.py
drewrisinger: This should be caught using argparse's ``choices`` kwarg, so removed entirely. If you don't want that, suggest what error message would be better.
https://github.com/m-labs/artiq/pull/1185/commits/a3711a4e0c1f942f8800f2981a9f7a9652aebfbc

drewrisinger:
I have made all explicitly asked changes. The docstrings that @sbourdeauducq called out above are recommended by `pydocstyle` because they are publicly available functions and so should be documented. 

Also, @sbourdeauducq never said what should happen in the above comments. A statement of a docstring being "obvious" or "redundant" is not saying that it should be removed. Sometimes effective communication requires redundancy to be effective.

sbourdeauducq:
> Also, @sbourdeauducq never said what should happen in the above comments. A statement of a docstring being "obvious" or "redundant" is not saying that it should be removed

They should be removed.

artiq/frontend/artiq_client.py
sbourdeauducq: No error message is fine since users are never supposed to see this error. This is only to give a hint to the developer.

sbourdeauducq: > This should be caught using argparse's `choices` kwarg, so removed entirely.

Yes, if the program is bug-free; as we all know, they always are.

sbourdeauducq: And I already explained this https://github.com/m-labs/artiq/pull/1185#discussion_r230968686

sbourdeauducq:
> The docstrings that @sbourdeauducq called out above are recommended by `pydocstyle` because they are publicly available functions and so should be documented.

Instead of blindly following a rule you found somewhere, you should ask yourself whether the docstring contents actually adds value. Yes, ARTIQ documentation can be vastly improved, but this is not the right way to do it as @klickverbot has already pointed out.

drewrisinger:
Made all requested changes.

artiq/frontend/artiq_client.py
```diff
@@ -114,20 +124,11 @@ def get_argparser():
     return parser
 
 
-def _parse_arguments(arguments):
```
sbourdeauducq: good catch :)

PR #1191, opened 2018-11-08
firmware: Fix dma_record_output_wide

klickverbot:
Tested locally on the actual application, but there don't seem to be any tests for wide-output DMA @sbourdeauducq / @whitequark: Did I miss a place where an integration test could easily be added?


whitequark:
IIRC back when I added it there weren't any consumers of wide DMA data, so no unit test. Sorry about that.

PR #1193, opened 2018-11-14
gui,scan: add CenterScan Scannable variant

jordens:
* parametrized by center/span/step instead of
  start/stop/npoints which is more convenient in some applications
* no scan widget support so far

Signed-off-by: Robert Jördens <rj@quartiq.de>

# ARTIQ Pull Request

## Type of Changes

|   | Type |

### All Pull Requests

- [ ] Update [RELEASE_NOTES.md](../RELEASE_NOTES.md) if there are noteworthy changes, especially if there are changes to existing APIs.
- [x] Close/update issues.

### Code Changes

- [x] Check, test, and update the conda recipes in [conda/](../doc/)

### Documentation Changes

### Git Logistics

  ```
  
  ```

### Licensing

### DIFF
--- a/artiq/gui/entries.py
+++ b/artiq/gui/entries.py
@@ -275,6 +275,78 @@ class _RangeScan(LayoutWidget):
         randomize.setChecked(state["randomize"])
 
 
+class _CenterScan(LayoutWidget):
+    def __init__(self, procdesc, state):
+        LayoutWidget.__init__(self)
+
+        scale = procdesc["scale"]
+
+        def apply_properties(widget):
+            widget.setDecimals(procdesc["ndecimals"])
+            if procdesc["global_min"] is not None:
+                widget.setMinimum(procdesc["global_min"]/scale)
+            else:
+                widget.setMinimum(float("-inf"))
+            if procdesc["global_max"] is not None:
+                widget.setMaximum(procdesc["global_max"]/scale)
+            else:
+                widget.setMaximum(float("inf"))
+            if procdesc["global_step"] is not None:
+                widget.setSingleStep(procdesc["global_step"]/scale)
+            if procdesc["unit"]:
+                widget.setSuffix(" " + procdesc["unit"])
+
+        center = ScientificSpinBox()
+        disable_scroll_wheel(center)
+        apply_properties(center)
+        center.setPrecision()
+        center.setRelativeStep()
+        center.setValue(state["center"])
+        self.addWidget(center, 0, 1)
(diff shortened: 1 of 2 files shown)


jordens:
Since this will break the (major version-ed) dashboard UI state format I'd like to get this into 4.0.

sbourdeauducq:
Looks good.

dhslichter:
@jordens a couple of thoughts: 

1. I think center/span/npoints is a more common use case than center/span/step, although certainly both could be used.  My sentiment arises from the notion that in an experiment one often wants to scan over a span centered around a given frequency, where you either change the span or the center, but in changing the span one (typically) keeps the number of points the same, such that it takes the same amount of time to run the scan.  I think the case of widening the span while keeping the step/resolution constant is less common than widening the span while keeping number of points constant.  Therefore I would suggest it's more suitable to parameterize by number of points because then one needn't change both span and step when changing span and desiring to keep run time constant.  

2. It appear that the current code doesn't scan in order -- `sequence` goes from `center + step` to `center + int(span/(2.*step))*step`, then back to `center` and scanning out to `center - int(span/(2.*step))*step`.  Am I reading this incorrectly?  I'd suggest it should go from `center - int(span/(2.*step))*step` to `center + int(span/(2.*step))*step`, passing through `center` in the middle.

jordens:
There are multiple ways to parametrize a scan. `start, stop` and `center, span` are a conjugate. As are `span, step` and `span, npoints`. If you are in a situation where you are searching for a feature that has a known size (e.g. fourier limited) and is located in a certain span, the parametrization implemented here is the suitable and convenient one. Another use case is when you are generating "nice" publication data where you are interested in an even step size. Also generally when the number of points is irrelevant.
List comprehension in Python has (at least to me) somewhat counterintuitive precedence. It scans center first and then alternates the deviation from the center with increasing distance from center.

dhslichter:
@jordens ack.  Thanks for clearing up the Python list comprehension.  Also, my preference for `npoints` over `step` is a weak one; I usually just do the math in my head anyway with the current scan type.  It feels too much like feature creep in the GUI to have all four kinds of scans (`start,stop` with `npoints` or `step`, and `center, span` with `npoints` or `step`) separately listed with radio buttons, but could one make a clean GUI interface for toggling between `step` and `npoints` within a given `RangeScan` or `CenterScan`?  I might have a think about this if you are not opposed.  If you are strongly against it, forget I mentioned it.

PR #1209, opened 2018-12-11
coredevice, firmware: Add rtio_input_timestamped_data

klickverbot:
This adds a function to retrieve both timestamp and data of an RTIO input event (with an optional timeout), similar to the other `rtio_input_*` syscalls.

Integration tests to follow as part of an RTIO counter phy that makes use of this.


artiq/firmware/ksupport/rtio.rs
```diff
@@ -149,6 +156,38 @@ mod imp {
         }
     }
 
+    pub extern fn input_timestamped_data(timeout: i64, channel: i32) -> TimestampedData {
```
sbourdeauducq: Nitpicking but IIRC you can put ``#[repr(C)]`` on the function and return ``(u64, i32)``. This way it's a tuple in both Rust and Python; and not a structure in Rust and a tuple in Python.

artiq/firmware/ksupport/rtio.rs
```diff
@@ -1,7 +1,14 @@
+#[repr(C)]
+pub struct TimestampedData {
+    timestamp: u64,
```
sbourdeauducq: Should be ``i64`` for ARTIQ-Python consistency.

artiq/firmware/ksupport/rtio.rs
klickverbot: Ah, yes – that's much cleaner.

artiq/firmware/ksupport/rtio.rs
klickverbot: I just went with the `rtio_input_timestamp` return type, but will change.

sbourdeauducq:
Does it work with type aliases? I remember doing this before and not having to define a struct.

artiq/firmware/ksupport/rtio.rs
```diff
@@ -200,14 +230,18 @@ mod imp {
...
-    pub extern fn input_timestamp(_timeout: i64, _channel: i32) -> u64 {
+    pub extern fn input_timestamp(_timeout: i64, _channel: i32) -> i64 {
         unimplemented!("not(has_rtio)")
     }
 
     pub extern fn input_data(_channel: i32) -> i32 {
         unimplemented!("not(has_rtio)")
     }
 
+    pub extern fn input_timestamped_data(_timeout: i64, _channel: i32) -> (i64, i32) {
```
sbourdeauducq: ``#[repr(C)]`` is missing there (though it doesn't work anyway?)

PR #1181, opened 2018-10-22
[tools]: Add --version flag to cmd line args

drewrisinger:
# ARTIQ Pull Request

## Description of Changes

Rename `verbosity_args()` to `add_common_args()`. Add option for `--version`.
Use `process_common_args()` instead of `init_logger()`. Rename `init_logger()` to `_init_logger()` because deprecated.

### Related Issue

Closes #1160

## Type of Changes

|   | Type |
| ------------- | ------------- |
| ✓  | :sparkles: New feature |

### All Pull Requests

- [x] Use correct spelling and grammar.
- [x] Update [RELEASE_NOTES.md](../RELEASE_NOTES.md) if there are noteworthy changes, especially if there are changes to existing APIs.
- [x] Close/update issues.
- [x] Check the copyright situation of your changes and sign off your patches (`git commit --signoff`, see [copyright](../CONTRIBUTING.rst#copyright-and-sign-off)).

### Code Changes

- [x] Run `flake8` to check code style (follow PEP-8 style). `flake8` has issues with parsing Migen/gateware code, ignore as necessary.
- [x] Add and check docstrings and comments

### Documentation Changes

- [x] Check, test, and update the documentation in [doc/](../doc/). Build documentation (`cd doc/manual; make html`) to ensure no errors.

### Git Logistics

- [x] Split your contribution into logically separate changes (`git rebase --interactive`). Merge/squash/fixup commits that just fix or amend previous commits. Remove unintended changes & cleanup. See [tutorial](https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase).
- [x] Write short & meaningful commit messages. Review each commit for messages

PR #1238, opened 2019-01-09
tools: document argument generators and handlers

drewrisinger:
Add docstrings for simple_network_args(), bind_address_from_args(),
and init_logger()

Factored out of #1181

### DIFF
--- a/artiq/tools.py
+++ b/artiq/tools.py
@@ -114,6 +114,23 @@ def verbosity_args(parser):
 
 
 def simple_network_args(parser, default_port):
+    """Add simple network arguments to a command-line argument parser.
+    
+    Added arguments:
+        * `--bind`
+        * `--no-localhost-bind`
+        * `--port`/`-p` or `--port-NAME` (depending on value of `default_port`)
+
+    Args:
+        parser (argparse.ArgumentParser): Parser to add arguments to.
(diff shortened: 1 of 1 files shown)


artiq/tools.py
```diff
@@ -114,6 +114,23 @@ def verbosity_args(parser):
 
 
 def simple_network_args(parser, default_port):
+    """    Add simple network arguments to a command-line argument parser.
```
sbourdeauducq: Why the spaces at the beginning?

drewrisinger: Fixed. Formatting mistake when removing ``\n`` after ``"""``

artiq/tools.py
```diff
@@ -156,11 +173,22 @@ def multiline_log_config(level):
...
     multiline_log_config(
         level=logging.WARNING + args.quiet*10 - args.verbose*10)
 
 
 def bind_address_from_args(args):
+    """Return the bind address from parsed arguments.
+    
+    Returns:
+        None if "*" in args.bind. `args.bind` if `args.no_localhost_bind`. 
+        Otherwise, returns local IP (`127.0.0.1, ::1`) and `args.bind`
```
sbourdeauducq: Translating code verbatim into English is not useful documentation.

PR #1244, opened 2019-01-12
Implement gateware-level input edge counters

klickverbot:
This adds a gateware-level input edge counter, implemented as an RTIO PHY. See release notes/tests/docs for motivation and details.

Currently, only a `SimpleEdgeCounter` is implemented which works at the RTIO clock. Support for a higher event rate on SERDES could be added in the future without modifying the RTIO interface (currently, the LSB of the input word is sampled, with the same restrictions as the for the `TTL*` gates).

Also modifies the Kasli tester target to run loopback tests during CI.

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -14,6 +14,11 @@ ARTIQ-5
   have been renamed to ``ref_time_mu`` for consistency, as they are in machine
   units.
 * :func:`~artiq.tools.verbosity_args` renamed to :func:`~artiq.tools.add_common_args`. New feature: adds an option to print the ARTIQ version.
+* A gateware-level input edge counter has been added, which offers higher
+  throughput and increased flexibility over the usual TTL input PHYs where
+  edge timestamps are not required. See :mod:`artiq.coredevice.edge_counter` for
+  the core device driver and :mod:`artiq.gateware.rtio.phy.edge_counter`/
+  :meth:`artiq.gateware.eem.DIO.add_std` for the gateware components.
 
 
 ARTIQ-4
@@ -22,7 +27,7 @@ ARTIQ-4
 4.0
 ***
 
-* The ``artiq.coredevice.ttl`` drivers no longer track the timestamps of 
+* The ``artiq.coredevice.ttl`` drivers no longer track the timestamps of
   submitted events in software, requiring the user to explicitly specify the
   timeout for ``count()``/``timestamp_mu()``. Support for ``sync()`` has been dropped.
 


RELEASE_NOTES.rst
```diff
@@ -14,6 +14,60 @@ ARTIQ-5
   have been renamed to ``ref_time_mu`` for consistency, as they are in machine
   units.
 * :func:`~artiq.tools.verbosity_args` renamed to :func:`~artiq.tools.add_common_args`. New feature: adds an option to print the ARTIQ version.
+* A gateware-level input edge counter has been implemented (driver:
```
sbourdeauducq: Normally, release notes contain breaking changes. Those examples should be moved to the manual.

artiq/gateware/rtio/phy/edge_counter.py
```diff
@@ -0,0 +1,83 @@
...
+        count_falling = Signal()
+        send_event_stb = Signal()
+        zero_counter_stb = Signal()
+
+        # Read configuration from RTIO output events.
+        self.sync.rio_phy += [
+            If(self.rtlink.o.stb, count_rising.eq(self.rtlink.o.data[0]),
+               count_falling.eq(self.rtlink.o.data[1]),
+               send_event_stb.eq(self.rtlink.o.data[2]),
+               zero_counter_stb.eq(self.rtlink.o.data[3])).Else(
```
jordens: This is a case where we deviate from flake8 formatting. Align `If(` and `).Else(`.

artiq/gateware/rtio/phy/edge_counter.py
```diff
@@ -0,0 +1,83 @@
...
+        current_count = Signal(counter_width)
+
+        count_rising = Signal()
+        count_falling = Signal()
+        send_event_stb = Signal()
+        zero_counter_stb = Signal()
+
+        # Read configuration from RTIO output events.
+        self.sync.rio_phy += [
+            If(self.rtlink.o.stb, count_rising.eq(self.rtlink.o.data[0]),
```
jordens: Line break and 4 space indent for the body after the condition.

artiq/gateware/rtio/phy/edge_counter.py
```diff
@@ -0,0 +1,83 @@
...
+
+        current_count = Signal(counter_width)
+
+        count_rising = Signal()
+        count_falling = Signal()
+        send_event_stb = Signal()
+        zero_counter_stb = Signal()
+
+        # Read configuration from RTIO output events.
+        self.sync.rio_phy += [
```
jordens: Sensitivity should be in the rio domain (see TTL). The others as well AFAICT.
Design for no glitches, no spurious events, and external state kept on rtio reset.
The input state path seems to switch CDs multiple times.

artiq/gateware/rtio/phy/edge_counter.py
```diff
@@ -0,0 +1,83 @@
...
+        self.sync.rio += [
+            self.input_state.eq(input_state),
+            input_state_d.eq(self.input_state)
+        ]
+
+        # Count input edges, saturating at the maximum.
+        new_count_raw = Signal(counter_width + 1)
+        self.comb += new_count_raw.eq(
+            Cat(current_count, C(0, 1)) +
+            (count_rising & (self.input_state & ~input_state_d)) +
```
jordens: I assume vivado is smart enough to transform that sum to a single OR.

artiq/gateware/rtio/phy/edge_counter.py
```diff
@@ -0,0 +1,83 @@
...
+            (count_falling & (~self.input_state & input_state_d)))
+        new_count = Signal.like(current_count)
+        self.comb += new_count.eq(
+            Mux(new_count_raw[counter_width], 2**counter_width - 1,
```
jordens: Is saturation really important here? One could argue that overflow is well defined for a counter in terms of aliasing.
Also the saturation logic as written seems convoluted.

artiq/gateware/eem.py
```diff
@@ -38,20 +38,32 @@ def io(eem, iostandard="LVDS_25"):
             for i in range(8)]
 
     @classmethod
-    def add_std(cls, target, eem, ttl03_cls, ttl47_cls, iostandard="LVDS_25"):
+    def add_std(cls, target, eem, ttl03_cls, ttl47_cls, iostandard="LVDS_25",
+            add_edge_counters=False):
```
jordens: Keeping this in line with the other arguments and the other eems, I'd pass the class to use and not a Boolean. That paves the way for a future counter phy that handles serdes data.

artiq/gateware/rtio/phy/edge_counter.py
sbourdeauducq: There should at least be something to tell you that overflow has occured.

jordens: Ok. Could just use the msb for that (set on overflow, clear on counter clear).

RELEASE_NOTES.rst
klickverbot: Sure – just wasn't sure where to put it. It's in the module documentation for now.

artiq/gateware/rtio/phy/edge_counter.py
klickverbot: I've cleaned up the saturation logic (this code was previously using SERDES inputs with more than one edge per cycle, but I've decided not to spend the time verifying all the edge cases for now).

I don't care too much about exactly overflow is handled, as it should be relatively rare in practice – filling 31 bits at 10 MHz takes more than three minutes. What I do care about is that it doesn't go unnoticed in user code, because it would be a sporadic, hard to debug issue. Hence, I wouldn't want to move away from exceptions on the user side.

We could always expose more details on overflow later, though, and without breaking the API in the process. For example, we could make the gateware use the MSB for overflow as suggested, and either add the wrapped count to the exception thrown or add a flag to the core device driver constructor to switch off the overflow checks.

artiq/gateware/eem.py
klickverbot: Good point (not sure why I didn't implement it this way) – done.

artiq/gateware/rtio/phy/edge_counter.py
klickverbot: Whoops, I had `rio`/`rio_phy` backwards. Fixed.

artiq/gateware/rtio/phy/edge_counter.py
```diff
@@ -0,0 +1,79 @@
...
+    """
+
+    def __init__(self, input_state, counter_width=31):
+        assert counter_width >= 2
+
+        # RTIO interface:
+        #  - output 0: 4 bits, <count_rising><count_falling><send_event><zero_counter>
+        #  - input 0: 32 bits, accumulated edge count
+        self.rtlink = rtlink.Interface(
+            rtlink.OInterface(5, enable_replace=False),
```
jordens: Not 4 bits?

artiq/gateware/rtio/phy/edge_counter.py
```diff
@@ -0,0 +1,79 @@
...
+    which is is also used to request an input event to be emitted with the
+    current counter value.
+
+    :param input_state: The (scalar) input signal to detect edges of. This
+        should already be in the rio_phy clock domain.
+    :param counter_width: The width of the counter register, in bits. Defaults
+        to 31 to match integers being signed in ARTIQ Python.
+    """
+
+    def __init__(self, input_state, counter_width=31):
```
jordens: Not saying that this is wrong, just trying to understand the rationale: Is the reason that you use 31 here just that it makes both interpreting it as a u32 and upcasting to i64 difficult?
Your testbench uses 32 bits.

klickverbot: I'm using 31 bits by default just since using 32 bits is unattractive for user code anyway (have to deal with signedness/extend to i64), and this way we have the extra sign bit to play with in a backwards-compatible fashion in the future (re saturating behaviour vs. wrapping, etc.).

The testbench uses 32 bits and a shorter width just to make sure both compile.

PR #1251, opened 2019-01-20
scheduler: Add shutdown test, minor refactors

klickverbot:
This is just to CI-test a few uncontroversial changes and get them out of the way before a larger imminent PR to extend the scheduler (support for idling experiments that yield to lower-priority runs).

### DIFF
--- a/artiq/test/test_scheduler.py
+++ b/artiq/test/test_scheduler.py
@@ -42,7 +42,7 @@ def _get_expid(name):
 
 def _get_basic_steps(rid, expid, priority=0, flush=False):
     return [
-        {"action": "setitem", "key": rid, "value": 
+        {"action": "setitem", "key": rid, "value":
             {"pipeline": "main", "status": "pending", "priority": priority,
              "expid": expid, "due_date": None, "flush": flush,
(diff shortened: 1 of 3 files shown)


artiq/test/test_scheduler.py
```diff
@@ -185,6 +185,49 @@ def notify(mod):
...
+        scheduler.submit("main", expid_bg, -99, None, False)
+        loop.run_until_complete(background_running.wait())
+
+        scheduler.submit("main", expid, 0, None, False)
+        loop.run_until_complete(empty_ready.wait())
+
+        # At this point, (at least) BackgroundExperiment is still running; make
+        # sure we can stop the scheduler without hanging.
+        loop.run_until_complete(scheduler.stop())
+
```
jordens: At this point you probably want to check that the background job has not completed yet. Otherwise you are relying implicitly on some event loop granularity that I can't derive, right?

klickverbot: The background job will never complete unless termination is requested. `scheduler.stop()` should kill the workers, which is what this is testing.

PR #1262, opened 2019-02-02
Add methods to set experiment pipeline/priority/flush defaults

### LINKED ISSUE #692: Method to set default experiment priority / log level / pipeline
It would be really useful to have a method that can be called from build that sets the default priority for the experiment. It would potentially also be useful to be able to set a default log level and pipeline. This default would be used when the experiment is first opened in a GUI, or used if the user clicks "recompute all arguments".

An example where this is useful is for servo experiments. One often has 3 or so servo experiments that interrupt the running experiment to tweak parameters, with the servos each running at different priorities so they interrupt each other appropriately. At the moment the user needs to set these experiments to have the correct priority in each instance of the GUI they use - this is fragile.

cjbe:
# ARTIQ Pull Request

## Description of Changes
Add methods callable from experiment build() to set pipeline/priority/flush defaults.
Suggestions on a cleaner way of indicating the functionality of the "recompute scheduler options" button without bloating the GUI appreciated.
Closes #692

## Type of Changes

|   | Type |

### All Pull Requests

- [x] Close/update issues.

### Code Changes

### DIFF
--- a/artiq/dashboard/experiments.py
+++ b/artiq/dashboard/experiments.py
@@ -162,14 +162,14 @@ class _ArgumentEditor(QtWidgets.QTreeWidget):
 
     async def _recompute_argument(self, name):
         try:
-            arginfo = await self.manager.compute_arginfo(self.expurl)
+            expdesc = await self.manager.compute_expdesc(self.expurl)
         except:
             logger.error("Could not recompute argument '%s' of '%s'",
                          name, self.expurl, exc_info=True)
             return
         argument = self.manager.get_submission_arguments(self.expurl)[name]
 
-        procdesc = arginfo[name][0]
+        procdesc = expdesc["arginfo"][name][0]
         state = procdesc_to_entry(procdesc).default_state(procdesc)
         argument["desc"] = procdesc
         argument["state"] = state
@@ -272,7 +272,8 @@ class _ExperimentDock(QtWidgets.QMdiSubWindow):
             scheduling["due_date"] = due_date
         datetime_en.stateChanged.connect(update_datetime_en)
 
-        pipeline_name = QtWidgets.QLineEdit()
+        self.pipeline_name = QtWidgets.QLineEdit()
+        pipeline_name = self.pipeline_name
         self.layout.addWidget(QtWidgets.QLabel("Pipeline:"), 1, 2)
         self.layout.addWidget(pipeline_name, 1, 3)
 
@@ -280,9 +281,10 @@ class _ExperimentDock(QtWidgets.QMdiSubWindow):
 
         def update_pipeline_name(text):
             scheduling["pipeline_name"] = text
-        pipeline_name.textEdited.connect(update_pipeline_name)
+        pipeline_name.textChanged.connect(update_pipeline_name)
 
-        priority = QtWidgets.QSpinBox()
+        self.priority = QtWidgets.QSpinBox()
+        priority = self.priority
         priority.setRange(-99, 99)
         self.layout.addWidget(QtWidgets.QLabel("Priority:"), 2, 0)
         self.layout.addWidget(priority, 2, 1)
@@ -293,7 +295,8 @@ class _ExperimentDock(QtWidgets.QMdiSubWindow):
             scheduling["priority"] = value
         priority.valueChanged.connect(update_priority)
 
-        flush = QtWidgets.QCheckBox("Flush")
+        self.flush = QtWidgets.QCheckBox("Flush")
+        flush = self.flush
         flush.setToolTip("Flush the pipeline (of current- and higher-priority "
                          "experiments) before starting the experiment")
         self.layout.addWidget(flush, 2, 2, 1, 2)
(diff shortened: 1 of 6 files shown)


artiq/dashboard/experiments.py
```diff
@@ -341,6 +344,15 @@ def update_repo_rev(text):
             repo_rev.textChanged.connect(update_repo_rev)
             self.repo_rev = repo_rev
 
+        recompute_sched_options = QtWidgets.QToolButton()
```
jordens: I think separating this from the "Recompute Arguments" button is tricky. There may be hidden dependencies in an experiment that make a partial reset of only the scheduling parameters problematic (e.g. code has changed, different devices are used and pipeline name should change. arguments have been removed/moved and limits have changed). Is there a concrete use case for only resetting the scheduling parameters?

cjbe: The use case here is the same as for the individual argument recompute - you have messed up the scheduler parameters testing something (e.g. while tweaking argument values) and now want them to go back to the defaults, but want to keep the arguments.

However the main reason I was keen on a separate scheduler recompute is to keep a logical separation from the argument recompute logic. This separation allows for easy argument plugins (e.g. the [ndscan plugin](http://github.com/OxfordIonTrapGroup/ndscan.git)).

artiq/language/environment.py
```diff
@@ -352,6 +354,32 @@ def setattr_dataset(self, key, default=NoDefault, archive=True):
         dataset and of the attribute are the same."""
         setattr(self, key, self.get_dataset(key, default, archive))
 
+    def set_default_pipeline(self, pipeline):
```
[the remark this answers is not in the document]
cjbe: > And why not allow computation of `due_date`?
I could not think of a good use case for this - if you have one I am happy to add this.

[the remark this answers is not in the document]
cjbe: > What about making this `set_default_scheduling(priority=x, pipeline_name=y, flush=z, due_date=u)`

I don't have a strong preference here. My feeling was in a lot of use-cases one would only want to change one of {priority, pipeline, flush}, hence it would be cleaner to have separate setters.

artiq/dashboard/experiments.py
jordens: But then you'd want individual scheduling parameter recompute buttons (like in the analogy you draw) and the overall recompute would still recompute arguments and scheduling...

The alternative is to make the scheduling defaults an overridable method (`EnvExperiment.get_scheduling_defaults()`). That would clearly separate it, address my concern and allow your use case.

cjbe: @jordens I am not sure I understand correctly:
In your suggestion, the overrideable method would return (say) a dict of default priority/pipeline/flush? Then these default values are used when the dashboard "recompute scheduler defaults" button is clicked?
This does not seem to make the "recompute" behaviour any clearer to me.

Is your objection here linked to the individual scheduler parameter setters? If so, I assume a `set_default_scheduling()` function that sets all the defaults would alleviate this? (The logic being that one setter links to one recompute button). If this is the case, I am happy implementing this.

jordens: Sorry for the late reply. I like the general feature. I'd suggest the following:

* Phrasing: call this "reset" of scheduling settings
* Have the global reset (the recompute all arguments button) perform a reset of all arguments and all scheduling settings.
* API: `EnvExperiment.set_default_scheduling(priority=x, pipeline_name=y, flush=z, due_date=u)` to be in line with the existing "active" API around setting/creating arguments.
* If you need resets for individual items in the scheduling settings, then let's do that in a context menu to not clutter up the always-visible part of the UI.

cjbe: @jordens I agree on all but the second point.

I am hesitant to couple the scheduler options to the "recompute all arguments" button, as this belongs to the ArgumentEditor - keeping this interface clean allows easy custom argument editors (which we are already using for our scan framework [here](https://github.com/dnadlinger/artiq/tree/ndscan)).

Then the options are a "scheduler settings reset" button outside of the argument editor, as in the below, or a context menu - my preference is the former, as this seems to fit well into the current GUI. Do you have strong feelings on this?
image

jordens: I'm ok with keeping it out of the argument editor. But imo that button at that location makes an already cluttered and dense interface worse. It's meaning is unclear and the placement is unintuitive.

cjbe: I do see what you mean. My worry with a context menu is that many people would never notice it existed.

Another option is to shrink the submit button vertically, and add a "reset scheduler settings" button below "terminate instances", but I feel this gives more prominence than is desired to this (probably infrequently used) button. I guess context menu it is...
image

dhslichter: FWIW, I think context menu is the least-bad option here (there aren't any good ones, alas) -- I think this is a rarely used feature, and also that the previous suggested draft shown in https://github.com/m-labs/artiq/pull/1262#discussion_r263166497 is, as @jordens says, too cluttered and un-intuitive as to the function.

cjbe:
@jordens I think these changes cover all your points.

jordens:
Looks good. Please rebase and then this can go in.

PR #1268, opened 2019-02-07
SUServo: remove references to non-existent a0 parameter

hartytp:
I believe this parameter was used in a previous version of the servo, but has since been removed.

### DIFF
--- a/artiq/coredevice/suservo.py
+++ b/artiq/coredevice/suservo.py
@@ -319,7 +319,7 @@ class Channel:
         The recurrence relation is (all data signed and MSB aligned):
 
         .. math::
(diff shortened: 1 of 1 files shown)


jordens:
That's incorrect. Now the equation is wrong. Please revert. If you don't like the notation, then suggest one that's better but still correct.

hartytp:
Okay, sorry, must have misunderstood. Is `a_0` just a fixed value?

jordens:
Yes. It's the normalization factor (or the fixed point scaling).

https://github.com/hartytp/artiq/blob/f3aab2b89156bbc1b12f847093a87a8933293df2/artiq/coredevice/suservo.py#L393-L394

PR #1271, opened 2019-02-08
language: Support appending to datasets

dnadlinger:
This adds `HasEnvironment.append_to_dataset()` for (efficiently) appending to datasets from experiment code.

All the required `sync_struct` machinery is already in place; appending just wasn't exposed previously.

Also adds some unit tests; the dataset API was largely untested before.


dnadlinger:
Assuming this is uncontroversial.

PR #1279, opened 2019-02-16
conda: Bump migen/misoc commits to fix Sayma build

dnadlinger:
Per @vmsch, the Sayma targets apparently don't build with the mi{gen, soc} versions pulled in by the `artiq-dev` conda metapackage. This updates the version references to the latest commits, fixing the build.

### DIFF
--- a/conda/artiq-dev/meta.yaml
+++ b/conda/artiq-dev/meta.yaml
@@ -14,8 +14,8 @@ requirements:
   run:
     - python >=3.5.3,<3.6
     - setuptools 33.1.1
-    - migen 0.8 py35_0+git2d62c0c
-    - misoc 0.12 py35_0+git714ea689
+    - migen 0.8 py35_63+gitafe4405
+    - misoc 0.12 py35_11+git8e033c2c
     - jesd204b 0.10
     - microscope
     - binutils-or1k-linux >=2.27


sbourdeauducq:
FYI, I will probably deprecate conda-dev and all other Conda development packages (rustc, clang, etc.) after the Nix migration is complete. Conda is just too much of an hassle for CI and development whereas Nix works very well.

jordens:
Please only once the migration is really complete and everything is in place. I.e. release conda builds and uploads, ci/hitl testing, ci testing on Windows, documentation, documentation for developing, plan for maintaining precious release builds, etc.

sbourdeauducq:
Upload is not necessary with this script https://github.com/m-labs/artiq/blob/master/conda/install-artiq.py and using it avoids a bunch of problems (due to anaconda.org issues + the conda bugs which are worked around in the script).
I tested the script on two different Linux and two different Windows installations already, and there were no issues.

sbourdeauducq:
@dnadlinger @vmsch Hydra now does continuous builds for Sayma:
https://nixbld.m-labs.hk/build/281
I haven't sorted out yet details like conda packaging or making sure that artiq_flash finds the binaries without passing any special options, but this should ensure that the build (with the new process) doesn't break, and provide a binary cache.

Issue #676, opened 2017-02-27
scalable moninj


sbourdeauducq:
Currently, the dashboard displays every TTL or DDS channel available in the device database, with no way to select or reorder them. This is not nice on a large system (e.g. with DRTIO) that can have hundreds of channels.

I would propose the following:
- merge the TTL and DDS panels.
- the user selects which channels to display (they can add and remove them) from any TTL, DDS, or SAWG device.
- channels can be reordered via drag-and-drop.
- the current selection of channels, and their position, can be saved and restored in "configurations".
- there can be an arbitrary number of configurations, which are created, deleted and named by the user.

See also #603 and #675.

hartytp:
Additionally, while not explicit mentioned in @sbourdeauducq's above post, scalable moninj should also include a rethink of the low-level implementation e.g. to make it play nicer with things like the SU-servo.

sbourdeauducq:
> scalable moninj should also include a rethink of the low-level implementation e.g. to make it play nicer with things like the SU-servo.

That's just feeding the currently monitored address into the PHY so that its internal state can be mapped to block-RAM, right?

jordens:
There are several aspects for scaling up moninj. There is the **data exfiltration** aspect from the RTIO Phys:

There is no intrinsic problem with adding moninj taps to SU-servo.
It's just that doing so for all channels and all profiles is getting
towards monitoring and injecting more than a thousand values.
Monitoring/injecting the current values (current ADC value, active
profile number, rf switch/hold bits, clipping, current ASF, current
FTW) might already be very useful and is much more doable (there are
still some restrictions and interference by e.g. the global servo
enable bit) and would be significantly less complicated or expensive
resource-wise.

In general, if we can come up with a really scalable moninj (scaling
to ~10k values, update based, with the proper state management, and
with proper GUI design etc etc) that would be great for this and also
for other things like SAWG.

Then on the **protocol/networking** side:

There are probably protocols and architectures around that are
well suited and we don't need to invent something. Maybe MQTT, or one of the many other *MQ, or something like what @klickverbot did with the servos in Zurich. Such an approach
would definitely also be nice for Thermostat, Humpback and Booster. It tackles a lot of the issues that crop up with simple SCPI-like interfaces (streaming, broadcasting, multiplexing, notification/callbacks, wire format, relaying, interfaces to databases).

PR #1287, opened 2019-03-10
dashboard: Sort TTL moninj channels by name

dnadlinger:
With growing system complexity, the moninj channel index is no longer a very intuitive ordering for typical end users.

@sbourdeauducq: Okay to merge? I could add a context menu option to switch between orders instead, but from our experience, there really isn't a use case for ordering by channel index.


sbourdeauducq:
Seems OK. But generally, the moninj GUI was designed when systems had a few dozen channels and received no further funding since. https://github.com/m-labs/artiq/issues/676

sbourdeauducq:
Should be done consistently for DDS though. And release notes should be updated.

PR #1299, opened 2019-03-30
coredevice: Fix host-side serialization of (nested) lists

dnadlinger:


artiq/compiler/validators/escape.py
```diff
@@ -301,17 +300,20 @@ def visit_ClassDefT(self, node):
...
 
+        # Adopt the value region for any variables declared on the lhs.
+        for name in target_names:
+            region = self._region_of(name)
+            if isinstance(region, Region) and not region.present():
+                # Find the name's environment to overwrite the region.
+                for env in self.env_stack[::-1]:
+                    if name.id in env:
+                        env[name.id] = value_region
+                        break
```
whitequark: I'm not convinced this is correct. Regions should form a lattice with `Global` on the top and an empty (invalid) region on the bottom, and any path through the lattice should contract the region. Why wouldn't this code let short-lived values escape?

dnadlinger: Agreed – but this only sets the region once. I'm not convinced it is correct either, hence the discussion yesterday, but how (where) else is the region initialised for assignments?

dnadlinger: (Note the `present` check.)

dnadlinger: Let's continue the discussion in #1301.

PR #1300, opened 2019-03-31
compiler: Quote tuples as TTuple values

dnadlinger:
Previously, they would end up being TInstances, rendering them rather useless.

@whitequark: Am I missing something here that would make this undesirable in the general case, or was this just an oversight?

If we were worried about codegen for tuples with a lot of elements, we could emit a reference to a global constant instead.


whitequark:
> Am I missing something here that would make this undesirable in the general case, or was this just an oversight?

Oversight.

PR #1315, opened 2019-05-08
Sync struct exc

hartytp:
# ARTIQ Pull Request

## Description of Changes

Tweak exception handling in `sync_struct`

1. Catch `ConnectionErrors` in `Subscriber._receiver_cr`. Logic for this change is that disconnects from the server are handled by `disconnect_cb` rather than trying to catch exceptions from `_receiver_cr`. Unhanded exceptions in the receiver lead to `Future/Task exception was never retrieved` messages.
2. Flake8 whitespace fixes
3. Catch `ConnectionError` rather than its subclasses. also catch `CancelledError` in a couple of places
4. `except Exception:` rather than bare `except:` to make flake8 happy

### Licensing

### DIFF
--- a/artiq/protocols/sync_struct.py
+++ b/artiq/protocols/sync_struct.py
@@ -154,6 +154,8 @@ class Subscriber:
 
                 for notify_cb in self.notify_cbs:
                     notify_cb(mod)
+        except (ConnectionError, asyncio.CancelledError):
+            pass
         finally:
             if self.disconnect_cb is not None:
                 self.disconnect_cb()
--- a/artiq/protocols/sync_struct.py
+++ b/artiq/protocols/sync_struct.py
@@ -120,7 +120,7 @@ class Subscriber:
             self.writer.write(_protocol_banner)
             self.writer.write((self.notifier_name + "\n").encode())
             self.receive_task = asyncio.ensure_future(self._receive_cr())
-        except Exception:
+        except:
             self.writer.close()
             del self.reader
             del self.writer
--- a/artiq/protocols/sync_struct.py
+++ b/artiq/protocols/sync_struct.py
@@ -33,8 +33,8 @@ class ModAction(Enum):
     required.
 
     The path (member field) the change applies to is given in
-    ``m["path"]`` as a list; elements give successive levels of indexing. (There
-    is no ``path`` on initial initialization.)
+    ``m["path"]`` as a list; elements give successive levels of indexing.
+    (There is no ``path`` on initial initialization.)
 
     Details on the modification are stored in additional data fields specific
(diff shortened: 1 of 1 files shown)


sbourdeauducq:
> `except Exception:` rather than bare `except:` to make flake8 happy

That doesn't just make flake8 happy, that changes the behavior of the programs. Have you studied the impact of that change?

hartytp:
Thanks. I confess I’d forgotten about that difference. I still think we probably don’t want a bare except (should we catch keyboard interrupts here?) but I haven’t thought it through in detail so I’ll revert that change.

Other than that, okay to merge?

artiq/protocols/sync_struct.py
```diff
@@ -154,6 +155,8 @@ def __init__(self, notifier_name, target_builder, notify_cb=None, disconnect_cb=
 
                 for notify_cb in self.notify_cbs:
                     notify_cb(mod)
+        except (ConnectionError, asyncio.CancelledError):
```
sbourdeauducq: ``Subscriber.close`` is already catching ``asyncio.CancelledError``. The logic is that ``wait_for`` will re-raise the exception that terminated the task. Why catch it again here?

artiq/protocols/sync_struct.py
```diff
@@ -311,7 +314,7 @@ def __init__(self, notifiers):
                     await writer.drain()
             finally:
                 self._recipients[notifier_name].remove(queue)
-        except (ConnectionResetError, ConnectionAbortedError, BrokenPipeError, TimeoutError):
+        except (ConnectionError, TimeoutError, asyncio.CancelledError):
```
sbourdeauducq: Like above, ``CancelledError`` is caught in ``asyncio_server.stop``.

sbourdeauducq:
It would be good if this were split into separate PRs (or at least commits):
* flake8
* replace ``ConnectionError`` subclasses with ``ConnectionError``.
* catch ``ConnectionError`` in ``Subscriber._receiver_cr``
* catch ``asyncio.CancelledError`` in the tasks

The first three are not controversial. The last one looks incorrect or I do not understand the logic.

Why is that breaking your code?

artiq/protocols/sync_struct.py
hartytp: Okay, yes, I'll remove the CancelledError. Looking at it again, that was a work-around for a quamash bug that stopped exceptions propagating.

hartytp:
Okay, thanks. I'll drop the last one as that's now fixed in our quamash fork and make a new PR for the others.

hartytp:
Done

PR #1321, opened 2019-05-14
sync_struct: catch errors in notifier cbs

hartytp:
# ARTIQ Pull Request

## Description of Changes

sync_struct: catch exceptions in user callbacks and close the connection to the server. Currently, we only catch `ConnectionErrors` so exceptions in the user callback go un-handled.

### DIFF
--- a/artiq/protocols/sync_struct.py
+++ b/artiq/protocols/sync_struct.py
@@ -15,6 +15,8 @@ import asyncio
 from enum import Enum, unique
 from operator import getitem
 from functools import partial
+import logging
+import traceback
 
 from artiq.monkey_patches import *
 from artiq.protocols import pyon
@@ -22,6 +24,7 @@ from artiq.protocols.asyncio_server import AsyncioServer
 
 
 _protocol_banner = b"ARTIQ sync_struct\n"
+logger = logging.getLogger(__name__)
 
 
 @unique
@@ -153,8 +156,13 @@ class Subscriber:
                 else:
                     process_mod(target, mod)
 
-                for notify_cb in self.notify_cbs:
-                    notify_cb(mod)
+                try:
+                    for notify_cb in self.notify_cbs:
+                        notify_cb(mod)
+                except BaseException:
+                    logger.error("Exception in notifier callback: {}"
+                                 .format(traceback.print_exc()))
+                    break
         except ConnectionError:
             pass
         finally:


artiq/protocols/sync_struct.py
```diff
@@ -154,7 +156,12 @@ def __init__(self, notifier_name, target_builder, notify_cb=None,
                     process_mod(target, mod)
 
                 for notify_cb in self.notify_cbs:
-                    notify_cb(mod)
+                    try:
+                        notify_cb(mod)
+                    except Exception as e:
+                        logger.error("Exception in notifier callback: {}"
+                                     .format(e))
+                        raise ConnectionError
```
sbourdeauducq: raise ConnectionError?

hartytp: Maybe that's a bit of a hack. The intention was to ensure that the Subscriber gets closed when there is an exception in a callback.

Another way of achieving the same is to replace `except ConnectionError` with `Except Exception` in the original except statement.

sbourdeauducq: It's a hack. Please find a clean solution.

hartytp: happy with just expanding the original `except` block to `except Exception`? Or, even `except:` to cover cases like `KeyboardInterrupt` during a notifier cb?

hartytp: On second thoughts, I'd prefer to keep a separate exception block for the callback so that one gets a message in the log when a callback raises an exception, rather than silently dropping the connection.

Refactored to remove the hack.

hartytp:
Force pushed after:
- changing to catching `BaseException`
- using traceback to get a better error message (otherwise coroutine exceptions can be a pain to debug)

artiq/protocols/sync_struct.py
```diff
@@ -153,8 +156,13 @@ def __init__(self, notifier_name, target_builder, notify_cb=None,
...
                     process_mod(target, mod)
 
-                for notify_cb in self.notify_cbs:
-                    notify_cb(mod)
+                try:
+                    for notify_cb in self.notify_cbs:
+                        notify_cb(mod)
+                except BaseException:
+                    logger.error("Exception in notifier callback: {}"
+                                 .format(traceback.print_exc()))
```
sbourdeauducq: ``exc_info=True``?

artiq/protocols/sync_struct.py
```diff
@@ -153,8 +156,13 @@ def __init__(self, notifier_name, target_builder, notify_cb=None,
                 else:
                     process_mod(target, mod)
 
-                for notify_cb in self.notify_cbs:
-                    notify_cb(mod)
+                try:
+                    for notify_cb in self.notify_cbs:
+                        notify_cb(mod)
+                except BaseException:
```
sbourdeauducq: Why ``BaseException``?

sbourdeauducq: IIRC it should be ``Exception`` or just nothing.

sbourdeauducq:
@hartytp Please address comments above + resubmit against https://github.com/m-labs/sipyco

hartytp:
done

Issue #1332, opened 2019-06-20
Hardware CI tests for suservo


dnadlinger:
There should be at least one Urukul/Sampler set up for use with suservo in the CI system. Basic unit tests for the driver should be really useful, as the recently fixed issues show.

(Actually including a dummy feedback loop with a diode detector would be a nice touch, too, if you have the hardware to spare.)

sbourdeauducq:
duplicate of https://github.com/m-labs/artiq/issues/1004

PR #1333, opened 2019-06-20
coredevice/suservo: Fix {get,set}_y_mu() scaling

dnadlinger:
Y0 is 18 bit signed, so the code (correctly) used 2^17 for the scale factor. The maximum representable number is 2^17 - 1, though, so `Channel.set_y(1.0)` would overflow to negative full scale instead.

### DIFF
--- a/artiq/coredevice/suservo.py
+++ b/artiq/coredevice/suservo.py
@@ -450,12 +450,12 @@ class Channel:
         """Get a profile's IIR state (filter output, Y0) in machine units.
 
         The IIR state is also know as the "integrator", or the DDS amplitude
-        scale factor. It is 18 bits wide and unsigned.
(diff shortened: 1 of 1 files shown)


dnadlinger:
(See #1332 – CI tests for this would be really good to have!)

dnadlinger:
@jordens: Ping – I assume this was just an oversight rather than a deliberate decision? (Negative outputs don't seem very useful.)

jordens:
I think this is correct. There are two things:
* Meaningfull full scale is `2**17 - 1`.
* The docs inaccuracies/errors may have been some confusion from when I tried to eek out the (unused) sign bit in the state storage for that to shift one bit less and gain more coefficient dynamic range.

PR #1358, opened 2019-09-05
Correct error in artiq_influxdb_schedule

charlesbaynham:
# ARTIQ Pull Request

## Description of Changes

Fixes error in parameter name for DBWriter.update() call. Also adds logging to stderr for unhandled exceptions in Subscriber._receive_cr(). 

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

### All Pull Requests

### DIFF
--- a/artiq/frontend/artiq_influxdb_schedule.py
+++ b/artiq/frontend/artiq_influxdb_schedule.py
@@ -150,7 +150,7 @@ class Log(dict):
                         "arguments": v["expid"].get("arguments", {}),
                         "flush": v["flush"],
                     },
-                    keys={
+                    tags={
                         "status": "add",
                         "class_name": v["expid"]["class_name"],
                         "file": v["expid"]["file"],
--- a/artiq/frontend/artiq_influxdb_schedule.py
+++ b/artiq/frontend/artiq_influxdb_schedule.py
@@ -194,6 +194,8 @@ class MasterReader(TaskObject):
                     ConnectionRefusedError, ConnectionResetError) as e:
                 logger.warning("Connection to master failed (%s: %s)",
                                e.__class__.__name__, str(e))
+            except Exception as e:
+                logger.exception(e)
             else:
                 logger.warning("Connection to master lost")
             logger.warning("Retrying in %.1f seconds", self.retry)


artiq/protocols/sync_struct.py
```diff
@@ -157,6 +158,8 @@ def __init__(self, notifier_name, target_builder, notify_cb=None,
                     notify_cb(mod)
         except ConnectionError:
             pass
+        except:
+            sys.excepthook(*sys.exc_info())
```
jordens: Could you explain why this is correct?
I'd rather not assume that there is someone reading stderr and try to hit the logging infrastructure instead.

Exceptions in `_receive_cr()` should travel up through `Subscriber.close()` and then end up in the `logger` in `MasterRead._do()`.

artiq/frontend/artiq_influxdb_schedule.py
```diff
@@ -150,7 +150,7 @@ def notify_cb(self, mod):
                         "arguments": v["expid"].get("arguments", {}),
                         "flush": v["flush"],
                     },
-                    keys={
+                    tags={
```
jordens: Oh yes. Refactoring error and lack of testing.

artiq/protocols/sync_struct.py
charlesbaynham: I think the problem actually lies in `MasterReader._do()`. If you look at this loop:

https://github.com/m-labs/artiq/blob/1fb317778a2e12af51f2f83d35f909b52cfb9c68/artiq/frontend/artiq_influxdb_schedule.py#L180-L200

...an error raised in the Subscriber (or in this case by the target) will raise at the `wait_for` on l.190. This isn't handled, so the exception bubbles up to the next try/except, but it isn't handled here either. 

It's therefore raised as a normal exception which is lost because this is happening through `ensure_future`. Worse, the loop stops so the connection is never recovered. 

The correct fix is therefore to catch the exception here and pass it to the `logger` that's available. I'll change this PR accordingly.

jordens: Yes. Good catch!

PR #1359, opened 2019-09-09
tools: Wrap Task _do() calls in a generic exception handler

charlesbaynham:
# ARTIQ Pull Request

## Description of Changes

Wraps the _do() method of a TaskObject, called in TaskObject.start(), in an exception handler to catch and relaunch the method in case of exceptions and log this to the logger. 

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

## Details

The `_do()` method in many concrete implementations of TaskObjects implements error handling, but this is not universal. In e.g. 
https://github.com/m-labs/artiq/blob/98caaebaded20ebc02e9200a9e31ea8f47a8b84a/artiq/frontend/artiq_influxdb.py#L99-L121
...exceptions occuring in the first part will not be caught, causing the Task to end with no error message either on failure or subsequent write attempts. 

This commit wraps the `TaskObject._do()` method in a generic last-line-of-defense task handler, and restarts the method if an unhandled error occurs, logging this to the logger.

### DIFF
--- a/artiq/tools.py
+++ b/artiq/tools.py
@@ -240,7 +240,16 @@ async def asyncio_wait_or_cancel(fs, **kwargs):
 
 class TaskObject:
     def start(self):
-        self.task = asyncio.ensure_future(self._do())
+        async def log_exceptions(awaitable):
+            try:
+                return await awaitable()
+            except asyncio.CancelledError:
+                raise
+            except Exception:
+                logger.error("Unhandled exception in TaskObject task body", exc_info=True)
+                raise
+
+        self.task = asyncio.ensure_future(log_exceptions(self._do))
 
     async def stop(self):
         self.task.cancel()


sbourdeauducq:
Logging an error immediately (instead of having the exception re-raised later by ``asyncio.wait_for(self.task)``) to facilitate debugging those issues is OK.
Restarting the task is not generally the right thing to do and may have unintended side effects.

charlesbaynham:
Done, sorry that took a while

charlesbaynham:
@sbourdeauducq Regarding restarting of failed TaskObjects, I still think that it's a common enough desire that it'd be nice if TaskObject supported it. Here's the code I'm using right now to launch TaskObjects that need to run forever:

```python
def start_and_restart(obj: TaskObject, max_retries=None):
    '''
    Start a TaskObject, and restart it if it fails

    :param obj: TaskObject to be started
    :param max_retries: Maximum number of times to restart the Task
    '''
    async def starter(attempt_no):
        try:
            # Start the Task
            obj.start()
            # Extract the _do() loop task and wait for its completion
            await asyncio.wait([obj.task])
            obj.task.result()
        except asyncio.CancelledError:
            raise
        except Exception:
            # If it completed for any reason other than a cancellation, restart it
            if attempt_no < max_retries:
                logger.exception('Restarting artiq_influx_generic after exception', exc_info=True)
                await starter(attempt_no+1)
            else:
                logger.exception('Max number of retries reached: exiting', exc_info=True)

    asyncio.ensure_future(starter(attempt_no=0))
```

That could be added to the TaskObject if you wanted.

artiq/tools.py
```diff
@@ -240,7 +240,16 @@ def atexit_register_coroutine(coroutine, loop=None):
...
 class TaskObject:
     def start(self):
-        self.task = asyncio.ensure_future(self._do())
+        async def log_exceptions(awaitable):
+            try:
+                return await awaitable()
+            except asyncio.CancelledError:
+                raise
+            except Exception:
+                logger.exception("Unhandled exception in TaskObject task body", exc_info=True)
```
sbourdeauducq: logger.exception?

charlesbaynham: == `logger.error("error message", exc_info=True)`

sbourdeauducq: As I pointed before the rest of ARTIQ uses logger.error, so this should be consistent.

PR #1360, opened 2019-09-09
influx_schedule: log repo_rev along with other info

charlesbaynham:
# Log repo_rev to influxDB

## Type of Changes

|   | Type |

## Summary

Very simple tweak to log `repo_rev` to InfluxDB as a tag, via `artiq_influx_schedule`.

### DIFF
--- a/artiq/frontend/artiq_influxdb_schedule.py
+++ b/artiq/frontend/artiq_influxdb_schedule.py
@@ -148,6 +148,7 @@ class Log(dict):
                         "priority": v["priority"],
                         "due_date": v["due_date"] or -1.,
                         "arguments": v["expid"].get("arguments", {}),
+                        "repo_rev": v["expid"]["repo_rev"],
                         "flush": v["flush"],
                     },
                     tags={


charlesbaynham:
@jordens Really? I'd imagine that's exactly what I'd use it for: "select all data produced by revision X of the code, compare against revision Y". 

For cardinality, yes there are likely to be >100 different `repo_rev`s in the database, but I don't think that's a lot. By comparison, our code will be making a database write every 50ms and could do so for a period of weeks - months. Trying to select over this data for a given revision when `repo_rev` isn't indexed would be pretty slow.

charlesbaynham:
The scheme design guide in the InfluxDB docs says:

> ### Encode meta data in tags
> 
> Tags are indexed and fields are not indexed. This means that queries on tags are more performant than those on fields.
> 
> In general, your queries should guide what gets stored as a tag and what gets stored as a field:
> 
> * Store data in tags if they’re commonly-queried meta data
> * Store data in tags if you plan to use them with GROUP BY()
> * Store data in fields if you plan to use them with an InfluxQL function
> * Store data in fields if you need them to be something other than a string - tag values are always interpreted as strings

I'd say that the `repo_rev` 
 1. Is metadata
 2. May well be grouped by
 3. Probably won't be used in functions
 4. Will always be a string

jordens:
@charlesbaynham you have to look at the entire documentation. Dtoring hashes in tags is explicitly advised against.

> Tags containing highly variable information like UUIDs, hashes, and random strings will lead to a large number of series in the database, known colloquially as high series cardinality. High series cardinality is a primary driver of high memory usage for many database workloads.

But for the experiments that I am working on we are already way above the 100 
revisions you are looking at.

Are you sure your use case of comparing code revisions is that frequent? The way you describe it, it seems you would rather compare and limit by time. It is usually the time that give you the relevant context required for comparison as it points to a dataset epoch. With the same line of argument you would say that filtering by expid-arguments (or even more likely by datasets at that point in time) would be required.

charlesbaynham:
I suppose I'm thinking about it a bit backwards: a more likely process would be "get list of RIDs associated with a particular revision, then retrieve those RIDs from the (indexed) dataset storage". So the query to the scheduler table is actually very lightweight, it's the dataset retrieval by RID that would be slow but that's indexed anyway. 

The [hardware sizing recommendations for InfluxDB](https://docs.influxdata.com/influxdb/v1.7/guides/hardware_sizing/) suggest that 100k ordinality is the limit for "low" performance hardware. With the `repo_rev` included, 
https://github.com/m-labs/artiq/blob/9a697dd3e694914108c9be2c53036bd94d4018a3/artiq/frontend/artiq_influxdb_schedule.py#L153-L159
that'd be of order

* x3 - status
* x100 - class_name
* x1 - file ~= class_name
* x10 - pipeline

= 3000

Adding ~1000 `repo_rev` would bring this to 3m, so I agree, no need.

jordens:
Thanks!

PR #1361, opened 2019-09-11
pc_rpc.Server: Handle generic exceptions without crashing

charlesbaynham:
# ARTIQ Pull Request

## Description of Changes

Add a generic Exception handler to Server to prevent it from crashing unexpectedly during requests. Report these exceptions to the `logger` and also to the requesting Client. 

|   | Type |
| ✓  | :bug: Bug fix  |

### DIFF
--- a/artiq/protocols/asyncio_server.py
+++ b/artiq/protocols/asyncio_server.py
@@ -44,6 +44,8 @@ class AsyncioServer:
 
     def _client_done(self, task):
         self._client_tasks.remove(task)
+        if task.exception() and task.exception() != asyncio.CancelledError:
+            raise task.exception()
 
     def _handle_connection(self, reader, writer):
         task = asyncio.ensure_future(self._handle_connection_cr(reader, writer))
--- a/artiq/protocols/pc_rpc.py
+++ b/artiq/protocols/pc_rpc.py
@@ -150,6 +150,9 @@ class Client:
             if not more:
                 break
             buf += more.decode()
+        if not buf:
+            self.__socket.close()
+            raise ConnectionError('The RPC server closed the connection')
         return pyon.decode(buf)
 
     def __do_action(self, action):


sbourdeauducq:
``_process_action`` is already supposed to catch those exceptions.
Do you have a situation where it fails to?

charlesbaynham:
For the client, this just detects a closed connection and throws an appropriate exception. The client could also attempt to re-establish the connection automatically: that'd be fairly easy to implement if it's desirable.

artiq/protocols/asyncio_server.py
```diff
@@ -44,6 +44,8 @@ def __init__(self):
 
     def _client_done(self, task):
         self._client_tasks.remove(task)
+        if task.exception() and task.exception() != asyncio.CancelledError:
+            raise task.exception()
```
sbourdeauducq: What handles that exception then?
BTW isn't there a asyncio debug mode that already shows those tasks terminating with exceptions? Do we need additional code?

artiq/protocols/pc_rpc.py
```diff
@@ -150,6 +150,9 @@ def __recv(self):
             if not more:
                 break
             buf += more.decode()
+        if not buf:
+            self.__socket.close()
+            raise ConnectionError('The RPC server closed the connection')
```
sbourdeauducq: That should be a separate PR and done consistently (we have the same code pattern e.g. in sync_struct).

artiq/protocols/asyncio_server.py
charlesbaynham: Nothing by default handles this, but the exception deterministically raises here instead of whenever the garbage collector cleans up the task, which varies

sbourdeauducq: Maybe logging.error would be more appropriate then?

sbourdeauducq: Because I don't see any way that this exception could get caught.

charlesbaynham: There's not currently a logger available in `asyncio_server`. I could add it

artiq/protocols/pc_rpc.py
charlesbaynham: Which bit of `sync_struct` do you mean, this part:

https://github.com/m-labs/artiq/blob/e50a6d5aafcc08ecc8ecc30972001cd58c9604a0/artiq/protocols/sync_struct.py#L142-L149

?

sbourdeauducq: Yes.

charlesbaynham: So the `sync_struct` treats Publisher disconnection as a normal event, and just quietly shuts down, calling the `disconnect_cb` callback to let the user do whatever's required to handle it. The same approach for `Client` would make sense, so there's no error logged unless the user adds a callback (preserving current behaviour).

sbourdeauducq: Can you split this into a separate PR? Keep this one about the server.

sbourdeauducq:
@charlesbaynham Please address comments above + resubmit against https://github.com/m-labs/sipyco

charlesbaynham:
> 
> 
> @charlesbaynham Please address comments above + resubmit against https://github.com/m-labs/sipyco

Will do, I plan to get back to this before too long

charlesbaynham:
Closing since this is now addressed in the `sipyco` repo

PR #1366, opened 2019-09-20
environment: move prepare() call's logic to HasEnvironment

charlesbaynham:
## Description of Changes

EnvExperiment prepare() method calls the prepare() method of all children recursively instead of just the first layer. 

## Type of Changes

|   | Type |

### Code Changes

In the `prepare()` method of EnvExperiment, loop through the children of all children recursively calling `prepare()` on each if present.

### DIFF
--- a/artiq/language/environment.py
+++ b/artiq/language/environment.py
@@ -261,6 +261,32 @@ class HasEnvironment:
                             "be called from build()")
         return self.__argument_mgr.get(key, processor, group, tooltip)
 
+    def call_children_method(self, method, *args, **kwargs):
+        '''
+        Calls the named method for each child, if it exists for that child.
+
+        :param method: Name of the method to call
+        :type method: str
+        :param args: Tuple of positional arguments to pass to all children
+        :param kwargs: Dict of keyword arguments to pass to all children
+        :return: Dict of output from children which had the requested method.
+                 Keys are tuples of ([child class], [child object id]).
+        '''
+        ret = dict()
+        for child in self.children:
+            try:
+                child_method = getattr(child, method)
+            except AttributeError:
+                pass
+            else:
+                out = child_method(*args, **kwargs)
+                child_id = id(child)
+                child_class = child.__class__
+                ret[(child_class, child_id)] = out
+
+        return ret
+
+
     def setattr_argument(self, key, processor=None, group=None, tooltip=None):
         """Sets an argument as attribute. The names of the argument and of the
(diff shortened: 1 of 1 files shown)


dnadlinger:
How can children opt out of this if they already implement a recursive prepare/if recursion doesn't make sense?

sbourdeauducq:
Exactly, and you can do ``super().prepare()`` or ``EnvExperiment.prepare(self)`` explicitly if you want to call ``prepare`` on your child experiments.
This kind of breaking change also needs updating RELEASE_NOTES btw.

charlesbaynham:
> Exactly, and you can do `super().prepare()` or `EnvExperiment.prepare(self)` explicitly if you want to call `prepare` on your child experiments.

Hmm perhaps I'm thinking of this wrong then. I've currently got a library of `HasEnvironment`s which do bits of work, some of which benefit from a `prepare` stage. I don't want them showing up in the experiment list so they can't be `EnvExperiment`s, but they still get prepared when they're children of the root `EnvExperiment`. However, `HasEnvironment` has no `prepare()` stage, so I'd have to manually `prepare` each library's children instead of calling a `super()`

Is there currently a way of having libraries of code which don't show up in the experiment list, but still get a `prepare()` stage? Or could the current "prepare all children" logic be moved into a class `PrepareChildren`, which `EnvEnvironment` then inherits from? That'd preserve current behaviour but allow for classes which inherit from both `HasEnvironment` and `PrepareChildren` but don't show up in the experiment list.

charlesbaynham:
> 
> 
> How can children opt out of this if they already implement a recursive prepare/if recursion doesn't make sense?

You're right.

sbourdeauducq:
> Is there currently a way of having libraries of code which don't show up in the experiment list, but still get a `prepare()` stage?

Do they also have run and analyze stages? In this case they should be Experiments, and I suggest adding an "invisible" flag that make them disappear from the list (but has no other effects, e.g. you can still schedule them using the ``artiq_client`` command-line tool).

If not, what about adding a method ``HasEnvironment.call_children_method`` that calls the corresponding method on all the children? (since the notion of "children" is introduced by ``HasEnvironment``, it makes sense that it provides some methods to interact with them)

Then ``EnvExperiment.prepare`` would simply be ``self.call_children_method("prepare")`` and you could also call use it easily in your drivers.

charlesbaynham:
Good idea, I'll write it up

PR #1368, opened 2019-10-04
Fix ad9910 ram mode asf scale error

tballance:
RAM mode amplitude to ASF conversion should be `<< 18` rather than `<< 16` when in `RAM_DEST_ASF` mode.

page 33
https://www.analog.com/media/en/technical-documentation/data-sheets/AD9910.pdf


jordens:
Right. Thanks.

Issue #741, opened 2017-06-07
doc : coredevice method arguments too complex


jbqubit:
Core drivers reference artic.coredevice.sawg and .spline have unnecessarily complex argument types presented in documentation. For example sawg.config.set_clr().

In email, @jordens says
> That's a feature of the compiler, not of SAWG. Feel free to raise an issue.

and @sbourdeauducq says
>Getting sphinx to use str() instead of repr() on type annotations should fix this problem.
```
>>> str(TInt32)
'numpy.int32'
>>> repr(TInt32)
"artiq.compiler.types.TMono('int', OrderedDict([('width', artiq.compiler.types.TValue(32))]))"
```

sbourdeauducq:
https://github.com/sphinx-doc/sphinx/issues/3857
I have tested that the proposed change in ``format_annotation()`` does result in types being nicely printed.

jordens:
We could also just overload repr.

sbourdeauducq:
ARTIQ does the right thing in returning an ``eval()``-parseable string in ``repr()``.

whitequark:
@jordens It's used when debugging, because the user-readable representation can hide some details that may make inference fail. Of course that is not visible from source code.

dnadlinger:
This issue is back after a refactoring in Sphinx; the repr() calls are now in https://github.com/sphinx-doc/sphinx/blob/7faeb793e2e16cde4e5759443fb7f84efddcd9ea/sphinx/util/inspect.py.

dnadlinger:
(Testing patch.)

sbourdeauducq:
We can also use an actual (not monkey) patch on Hydra, the advantage is it'll likely show clearly when it breaks. But users of unpatched sphinx will have the issue.

sbourdeauducq:
It broke again.

PR #1371, opened 2019-10-11
compiler: Prettify repr for built-in types

dnadlinger:
`__repr__()` shows up in the Sphinx-generated documentation for functions with type annotations. Previously, type signatures would often span several lines, as all types would be written out as `artiq.compiler.types.TMono(%s, %s)`, etc.

This change makes those signatures eminently more readable by rendering the types in the way they are usually used in the source code, that is, `TInt32`, `TFloat`, `TList(TInt64)`, etc.

### DIFF
--- a/artiq/compiler/builtins.py
+++ b/artiq/compiler/builtins.py
@@ -12,6 +12,9 @@ class TNone(types.TMono):
     def __init__(self):
         super().__init__("NoneType")
 
+    def __repr__(self):
+        return "TNone"
+
 class TBool(types.TMono):
     def __init__(self):
         super().__init__("bool")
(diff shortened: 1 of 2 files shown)


sbourdeauducq:
https://github.com/m-labs/artiq/issues/741 ?

dnadlinger:
Huh, thanks for the heads-up. Let me make sure our build environment is on the latest version of Sphinx…

sbourdeauducq:
Sphinx didn't handle this issue properly so there's a monkey-patch in conf.py.

dnadlinger:
Seems like the monkey patch no longer works (and they still haven't fixed the issue), see e.g. https://m-labs.hk/artiq/manual-beta/core_drivers_reference.html#artiq.coredevice.edge_counter.EdgeCounter

dnadlinger:
Reinstated monkey patch in https://github.com/m-labs/artiq/commit/371388ecbe28bec74a2775392e440180b6c5bf39 instead.

PR #1385, opened 2019-11-11
Update for AD9910 ram mode

gthickman:
# ARTIQ Pull Request

## Description of Changes

Function set_asf() in class AD9910 should contain <<2, as per ASF register documentation on p. 58 of https://www.analog.com/media/en/technical-documentation/data-sheets/AD9910.pdf.

Updated docstrings for several functions in class AD9910.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

### DIFF
--- a/artiq/coredevice/ad9910.py
+++ b/artiq/coredevice/ad9910.py
@@ -360,6 +360,9 @@ class AD9910:
             (:const:`RAM_DEST_FTW`, :const:`RAM_DEST_POW`,
             :const:`RAM_DEST_ASF`, :const:`RAM_DEST_POWASF`).
         :param ram_enable: RAM mode enable.
+        :param manual_osk_external: Enable OSK pin control in manual OSK mode.
+        :param osk_enable: Enable OSK mode.
+        :param select_auto_osk: Select manual or automatic OSK mode
         """
         self.write32(_AD9910_REG_CFR1,
                      (ram_enable << 31) |
@@ -630,6 +633,10 @@ class AD9910:
 
     @kernel
     def set_frequency(self, frequency):
+        """Set the value stored to the AD9910's FTW register
+
+        :param frequency: frequency to be stored, in Hz
+        """
         return self.set_ftw(self.frequency_to_ftw(frequency))
 
     @kernel
--- a/artiq/coredevice/ad9910.py
+++ b/artiq/coredevice/ad9910.py
@@ -345,8 +345,7 @@ class AD9910:
     @kernel
     def set_cfr1(self, power_down=0b0000, phase_autoclear=0,
                  drg_load_lrr=0, drg_autoclear=0,
-                 internal_profile=0, ram_destination=0, ram_enable=0,
-                 manual_osk_external=0, osk_enable=0, select_auto_osk=0):
+                 internal_profile=0, ram_destination=0, ram_enable=0):
         """Set CFR1. See the AD9910 datasheet for parameter meanings.
 
         This method does not pulse IO_UPDATE.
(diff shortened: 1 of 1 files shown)


jordens:
Please split your changes.

ASF:
I'd personally prefer having the ASF appear as a 16 bit value with the two LSB ignored (in hardware). But since this went in anyway, let's stick with it.
@dnadlinger why is that scale factor 0x3ffe and not 0x3fff? I don't remember why.
It's 0x3fff here: https://github.com/m-labs/artiq/blob/e6db68e62a091cc4bd8b6d84d5b4609ed2986970/artiq/coredevice/ad9910.py#L451

OSK:
Are you sure you have actually tested this? AFAICT it doesn't work: https://github.com/sinara-hw/Urukul/issues/19

dnadlinger:
@jordens: Not sure – we only tend to use the profile mode here, so I haven't ever looked at those registers in detail.

jordens:
But you worked on the `amplitude_to_asf` `asf_to_amplitude` code.

dnadlinger:
Ah, I just added inverses of what was already there:

```diff
dpn@macbook-pro-107:~/Build/Source/artiq (master $%=)$ git show 7bcdeb825
commit 7bcdeb825be05a34adadbefa2eb7980d0f655335
Author: David Nadlinger <code@klickverbot.at>
Date:   Tue Jan 8 02:16:42 2019 +0000

    ad9910: Add inverse FTW/ASF conversions

diff --git a/artiq/coredevice/ad9910.py b/artiq/coredevice/ad9910.py
index 3e2ac009..e08313de 100644
--- a/artiq/coredevice/ad9910.py
+++ b/artiq/coredevice/ad9910.py
@@ -468,23 +468,37 @@ class AD9910:
         """
         return int32(round(self.ftw_per_hz*frequency))

+    @portable(flags={"fast-math"})
+    def ftw_to_frequency(self, ftw):
+        """Return the frequency corresponding to the given frequency tuning
+        word.
+        """
+        return ftw / self.ftw_per_hz
+
     @portable(flags={"fast-math"})
     def turns_to_pow(self, turns):
         """Return the phase offset word corresponding to the given phase
         in turns."""
         return int32(round(turns*0x10000))

-    @portable(flags={"fast-math"})
-    def amplitude_to_asf(self, amplitude):
-        """Return amplitude scale factor corresponding to given amplitude."""
-        return int32(round(amplitude*0x3ffe))
-
     @portable(flags={"fast-math"})
     def pow_to_turns(self, pow_):
         """Return the phase in turns corresponding to a given phase offset
         word."""
         return pow_/0x10000

+    @portable(flags={"fast-math"})
+    def amplitude_to_asf(self, amplitude):
+        """Return amplitude scale factor corresponding to given fractional
+        amplitude."""
+        return int32(round(amplitude*0x3ffe))
+
+    @portable(flags={"fast-math"})
+    def asf_to_amplitude(self, asf):
+        """Return amplitude as a fraction of full scale corresponding to given
+        amplitude scale factor."""
+        return asf / float(0x3ffe)
+
     @kernel
     def set_frequency(self, frequency):
         return self.set_ftw(self.frequency_to_ftw(frequency))
```

jordens:
Well yes. I wanted to know whether you gave any thought to the factor. Because I don't remember what my rationale was for using that value in the first place.

gthickman:
@jordens: The OSK fix does work. The OSK pin is ignored as long as manual_osk_external==0 (AD9910 datasheet p. 54). I've been able to use RAM mode and the ASF register in this way with an Urukul v1.2 board.

gthickman:
Note: changes to set_cfr1() have been taken out and moved into a separate PR.

PR #1406, opened 2019-12-17
artiq_flash: Add --no-variant flag to just use given path verbatim

dnadlinger:
@sbourdeauducq: We have been using a local patch like this for a long, long time, as our Buildbot setup just archives gateware builds into a uniform directory structure by (Buildbot) build target rather than bothering with variants – okay to merge?

### DIFF
--- a/artiq/frontend/artiq_flash.py
+++ b/artiq/frontend/artiq_flash.py
@@ -71,6 +71,8 @@ Prerequisites:
     parser.add_argument("-d", "--dir", help="look for board binaries in this directory")
     parser.add_argument("--srcbuild", help="board binaries directory is laid out as a source build tree",
                         default=False, action="store_true")
+    parser.add_argument("--no-variant", help="do not use the variant machinery - just look in the given directory",
+                        default=False, action="store_true")
     parser.add_argument("--force-rtm", help="force RTM actions on boards/variants that normally do not have a RTM",
                         default=False, action="store_true")
     parser.add_argument("action", metavar="ACTION", nargs="*",
@@ -346,6 +348,9 @@ def main():
         action in args.action
         for action in ["gateware", "rtm_gateware", "bootloader", "firmware", "load", "rtm_load"])
     variant = args.variant
+    if variant is None and args.no_variant:
+        # Just use default actions in the below.
+        variant = ""
     if needs_artifacts and variant is None:
         variants = []
         if args.srcbuild:
@@ -370,7 +375,9 @@ def main():
             raise ValueError("more than one variant found for selected board, specify -V. "
                 "Found variants: {}".format(" ".join(sorted(variants))))
     if needs_artifacts:
-        if args.srcbuild:
+        if args.no_variant:
+            variant_dir = ""
+        elif args.srcbuild:
             variant_dir = variant
         else:
             variant_dir = args.target + "-" + variant


artiq/frontend/artiq_flash.py
```diff
@@ -346,6 +348,9 @@ def main():
         action in args.action
         for action in ["gateware", "rtm_gateware", "bootloader", "firmware", "load", "rtm_load"])
     variant = args.variant
+    if variant is None and args.no_variant:
+        # Just use default actions in the below.
+        variant = ""
     if needs_artifacts and variant is None:
```
sbourdeauducq: This (variant listing) still gets executed when ``--no-variant`` is given. Is that OK?

sbourdeauducq:
The ``artiq_flash`` variant/directory handling is already quite hairy as it is. Is it really the best solution to add more complexity and corner cases to it? Why is your build system not using variants?

artiq/frontend/artiq_flash.py
dnadlinger: Is it? Variant should no longer be `None` then.

Give me ~1h to sort out the SiPyCo deployment so I can test this in situ…

dnadlinger:
You'd need to ask @cjbe about the details, but I suspect he would say that the answer to 

> The artiq_flash variant/directory handling is already quite hairy as it is. Is it really the best solution to add more complexity and corner cases to it? 

was precisely just not to use it.

Also, from a practical perspective during development, not being forced into the variant directory structure is really handy. For example, while iterating on firmware, I tend to just `python -m artiq.gateware.targets … --no-compile-gateware` and `artiq_flash … -d <build_dir> --no-variant firmware start`, without having to copy files around to make artiq_flash happy.

From a design point of view, handling all the OpenOCD setup/target-specific memory ranges/… seems like neatly independent piece of functionality; the variant directory lookup stuff being a pretty orthogonal concern.

sbourdeauducq:
> For example, while iterating on firmware, I tend to just `python -m artiq.gateware.targets … --no-compile-gateware` and `artiq_flash … -d <build_dir> --no-variant firmware start`, without having to copy files around to make artiq_flash happy.

This is what ``--srcbuild`` is for.

artiq/frontend/artiq_flash.py
dnadlinger: Works as intended; the variant listing is not executed.

dnadlinger:
> This is what --srcbuild is for.

Well, `--no-variant` works just the same, except that it's more flexible.

Just for context, we have a central archive of several hundred builds at this point, with some twenty targets each.

This is integrated into our tooling, such that with a simple local configuration mapping logical names to USB hostnames and Kasli serial numbers, we can do things like `./flash.py -n 1552` to flash all DRTIO nodes to a certain build number, or e.g. `./flash -d trap_satellite -g bcd86` to flash just one node to a specific Git hash.

All of this works beautifully and makes it easy even for barely-programmers to iterate on gateware configs or firmware changes. We just need to be able to use the target-related functionality of `artiq_flash` without it also enforcing unrelated notions of what the directory layout is supposed to be. (The latter was only added fairly recently in ARTIQ 4/ 61c64a76bea0aece490daa89f43b93f07e820c9a.)

sbourdeauducq:
> Well, `--no-variant` works just the same, except that it's more flexible.

Would it support flashing Sayma (both AMC and RTM at the same time) from a source build?

dnadlinger:
> Would it support flashing Sayma (both AMC and RTM at the same time) from a source build?

I don't particularly care about the merits of `--srcbuild`; I just want to be able to use the pre-ARTIQ-4 behaviour of artiq_flash.

sbourdeauducq:
> I don't particularly care about the merits of `--srcbuild`; I just want to be able to use the pre-ARTIQ-4 behaviour of artiq_flash.

I was asking because, short of a nice complete refactoring, I am trying to reduce the complexity and amount of corner cases. Removing ``--srcbuild`` and replacing it with ``--no-variant`` would have helped achieve that goal - but we need to flash Sayma.

sbourdeauducq:
``artiq_flash`` will be simplified with a reworking of the bitstream build service.

PR #1409, opened 2019-12-18
compiler: Allow `None` in type hints

dnadlinger:
Similar to how Python itself interprets `None` as `type(None)`,
make it translate to `TNone` in ARTIQ compiler type hints.

We have somewhat unintentionally been running with this change
locally for a long time, and (certainly without me realising) ended up
with `None` being used all over our code base – it seems like this
syntax comes naturally to users without much thought.

(Passes artiq-fast build tests, this time.)

### DIFF
--- a/artiq/compiler/embedding.py
+++ b/artiq/compiler/embedding.py
@@ -950,6 +950,9 @@ class Stitcher:
         return function_node
 
     def _extract_annot(self, function, annot, kind, call_loc, fn_kind):
+        if annot is None:
+            annot = builtins.TNone()
+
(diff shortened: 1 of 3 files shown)


sbourdeauducq:
Instead of deleting the test, could you use something other than ``None`` in it?

dnadlinger:
Done. (Took me a while to realise that I need to add `outputcheck` as a top-level package to Nix to get the executable, rather than as a Python withPackages package.)

sbourdeauducq:
Thanks

PR #1415, opened 2019-12-29
compiler: Improve type inference performance by identity checks/early returns

dnadlinger:
These two commits improve the total compile time (artiq_run invocation to start of kernel upload) in one typical experiment by ~17%.

---

@whitequark: I'm putting this up as a PR just in case you have any ideas to reduce the number of identity `unify()` calls altogether. From a quick look, there doesn't seem to be a pathological cause for those, though.

### DIFF
--- a/artiq/compiler/types.py
+++ b/artiq/compiler/types.py
@@ -81,6 +81,8 @@ class TVar(Type):
             return root
 
     def unify(self, other):
+        if other is self:
+            return
         other = other.find()
 
         if self.parent is self:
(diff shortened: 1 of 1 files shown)


dnadlinger:
@sbourdeauducq: Probably not major enough to mention in the release notes?

sbourdeauducq:
Thanks. Have you run the full test suite?

dnadlinger:
Full as in `artiq-fast`, yes.

PR #1432, opened 2020-03-02
ttl_serdes: detect edges on short pulses

jordens:
Edges on pulses shorter than the RTIO period were missed because the
reference sample and the last sample of the serdes word are the same.

This change enables detection of edges on pulses as short as the
serdes UI (and shorter as long as the pulse still hits a serdes sample
aperture).

In any RTIO period, only the leading event corresponding to the first
edge with slope according to sensitivity is registerd. If the channel is
sensitive to both rising and falling edges and if the pulse is contained
within an RTIO period, or if it is sensitive only to one edge slope and
there are multiple pulses in an RTIO period, only the leading event is
seen. Thus this possibility of lost events is still there. Only the
conditions under which loss occurs are reduced.

In testing with the kasli-ptb6 variant, this also improves resource
usage (a couple hundred LUT) and timing (0.1 ns WNS).

Still needs to see some more hardware testing.

### DIFF
--- a/artiq/gateware/rtio/phy/ttl_serdes_generic.py
+++ b/artiq/gateware/rtio/phy/ttl_serdes_generic.py
@@ -7,7 +7,7 @@ from artiq.gateware.rtio import rtlink
 def _mk_edges(w, direction):
     l = [(1 << i) - 1 for i in range(w)]
     if direction == "rising":
-        l = [2**w - 1 ^ x for x in l]
+        l = [((1 << w) - 1) ^ x for x in l]
     elif direction == "falling":
         pass
     else:
@@ -69,8 +69,8 @@ class InOut(Module):
 
         # Output enable, for interfacing to external buffers.
         self.oe = Signal()
-        # LSB of the input state (for edge detection; arbitrary choice, support for
-        # short pulses will need a more involved solution).
+        # input state exposed for edge_counter: latest serdes sample
+        # support for short pulses will need a more involved solution
         self.input_state = Signal()
 
         # # #
@@ -112,15 +112,16 @@ class InOut(Module):
         i_d = Signal()
         self.sync.rio_phy += [
             i_d.eq(i),
-            self.rtlink.i.stb.eq(
-                sample |
-                (sensitivity[0] & ( i & ~i_d)) |
-                (sensitivity[1] & (~i &  i_d))
-            ),
             self.rtlink.i.data.eq(i),
         ]
 
         pe = PriorityEncoder(serdes_width)
         self.submodules += pe
-        self.comb += pe.i.eq(serdes.i ^ Replicate(i_d, serdes_width))
-        self.sync.rio_phy += self.rtlink.i.fine_ts.eq(pe.o)
+        self.comb += pe.i.eq(
+            (serdes.i ^ Cat(i_d, serdes.i)) & (
+                (serdes.i & Replicate(sensitivity[0], serdes_width)) |
+                (~serdes.i & Replicate(sensitivity[1], serdes_width))))
+        self.sync.rio_phy += [
+            self.rtlink.i.fine_ts.eq(pe.o),
+            self.rtlink.i.stb.eq(sample | ~pe.n),
+        ]
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -8,6 +8,10 @@ ARTIQ-6
 
 Highlights:
 
+* Performance improvements:
+  - #1432: SERDES TTL inputs can now detect edges on pulses that are shorter
+    than the RTIO period
+
 Breaking changes:
 
 


artiq/gateware/rtio/phy/ttl_serdes_generic.py
```diff
@@ -7,7 +7,7 @@
 def _mk_edges(w, direction):
     l = [(1 << i) - 1 for i in range(w)]
     if direction == "rising":
-        l = [2**w - 1 ^ x for x in l]
+        l = [((1 << w) - 1) ^ x for x in l]
```
sbourdeauducq: This doesn't change anything, right? It should be a separate commit (if it needs to be committed at all).

jordens: It makes it consistent with the expression two lines above and is a benign clean-up.
Yes. You can ride the style and procedure rulebook on this. But will that improve the code or will it do the opposite?
Does this change bother you in any way? Is it wrong or detrimental?
I don't understand why you are so fussy about style and procedure here.

sbourdeauducq: > It makes it consistent with the expression two lines above

Ok then commit it. But I still think it should be separate and not tied with the functional change that detects short pulses.

sbourdeauducq:
Probably should be mentioned in release notes.

jordens:
Yes. Mentioning doesn't hurt. Even though I think  it doesn't break any previous contract, just improves on behavior.

sbourdeauducq:
I think we should expand the release notes a bit, i.e. mention what is new and not just breaking changes.

jordens:
IMO we should make the release notes generally more reader-friendly. Also reference the issues that were fixed (where they are mentioned) and the features that are merged. That all becomes easier if we agree to do PRs for as much as possible. But since the autobuild and test doesn't integrate into github anymore, we lack helpful and automatic CI information which makes development unwelcoming.

sbourdeauducq:
> That all becomes easier if we agree to do PRs for as much as possible.

I don't think so. Creating a PR is harder than writing a few lines into the release notes.

> But since the autobuild and test doesn't integrate into github anymore, we lack helpful and automatic CI information which makes development unwelcoming.

Are you seriously suggesting that not having a few ticks/crosses next to commits is a serious obstacle to development? The new build system addresses issues more important than that. Besides, you are welcome to write some code that adds this feature to it.

jordens:
> Creating a PR is harder than writing a few lines into the release notes.

I disagree. It's just a few seconds using the tools available.
But that's not what I meant. I'd like to be able to retrace the changes, the issues, and see the discussion. Not just be pointed at some commit hash having to piece together the story myself.

> Are you seriously suggesting that not having a few ticks/crosses next to commits is a serious obstacle to development?

It's foolish to belittle that development style. It's extremely efficient and the best thing: people understand it and use it.

> The new build system addresses issues more important than that.

I did not mean to insult your current favorite package manager. I'm just pointing out a regression.

> Besides, you are welcome to write some code that adds this feature to it.

Given the choice between learning nix, developing the integration, and maintaining it or living with the regression, I'll choose the latter.

jordens:
All I wanted to say is that I would really enjoy being able to develop on ARTIQ with the convenience that makes rust or other projects so developer-friendly.

sbourdeauducq:
> Given the choice between learning nix, developing the integration, and maintaining it or living with the regression, I'll choose the latter.

There's another option: we do this together. You send me some script that takes as command line arguments the commit ID and the type of mark to add, reads the credentials from some file in /etc/nixos/secret, and connects to GitHub - and I'll add it to Hydra.
The latter is not that complicated and maintaining infrastructure that we control (unlike infrastructure controlled by Microsoft or Anaconda) isn't really much of a hassle.

If you want builds of PRs that seems possible too but I'll have to look into details. Note that building PRs on the previous system was extremely problematic:
* it caused frequent corruption of the conda cache, breaking other builds (the Nix store, albeit more complex, isn't brimming with bugs and has fewer nasty problems than the conda cache) - and in extreme cases even producing incorrect results (i.e. the reported commit ID did not correspond to the source that was built/tested).
* builds were often delayed as builds could not be parallellized and most intermediate results not cached + the machine was slow. Hydra supports parallel (and even distributed) builds and caches intermediate results quite efficiently.
* there wasn't much security and it would have been straightforward for someone to submit and build a PR that contained malicious code and compromised our server. Hydra has *some* sandboxing, though that would have to be carefully reviewed if we allow untrusted code to be built.

artiq/gateware/rtio/phy/ttl_serdes_generic.py
```diff
@@ -112,15 +112,16 @@ def __init__(self, serdes):
...
-            ),
             self.rtlink.i.data.eq(i),
         ]
 
         pe = PriorityEncoder(serdes_width)
         self.submodules += pe
-        self.comb += pe.i.eq(serdes.i ^ Replicate(i_d, serdes_width))
-        self.sync.rio_phy += self.rtlink.i.fine_ts.eq(pe.o)
+        self.comb += pe.i.eq(
+            (serdes.i ^ Cat(i_ref, serdes.i)) & (
```
dnadlinger, line 121: `i_ref`?

jordens: i_d. Yes.

PR #1445, opened 2020-05-01
ad53xx improvements

### LINKED ISSUE #1341: Voltage to MU isn't recognised as an attribute of zotino
`voltage_to_mu()`  isn't recognised as an attribute of the zotino. See my code below for outputting a ramp wave on the Zotino.

```

class ZotinoOutputs(EnvExperiment):
    def build(self): 

        self.setattr_device("core")
        self.setattr_device("zotino0")

    @kernel
    def set_zotino_voltages(self, zotino, voltages):
        self.core.break_realtime()                          
        zotino.init()                                       
        delay(200*us)                                       

        while(1):                                           
            for voltage in voltages:                        
                    
                zotino.write_dac_mu(0, voltage)             
                zotino.load()                              
                delay(20*us)                               
                
    def test_zotinos(self):
        n_steps = 100                                      
        voltages = [0.1*i for i in range(n_steps)]
        mu_voltages = [0]*n_steps
        
        for i in range(n_steps):
            mu_voltages[i] = self.zotino0.voltage_to_mu(voltage, self.offset_dacs,self.vref)
        
        self.set_zotino_voltages(self.zotino0, mu_voltages)    

    def run(self):
        self.core.reset()
        self.test_zotinos()

```

I expected this to output a ramp wave on the zotino, going from 0 to 10V in 0.1V steps. When I manually convert the voltages to machine units, it runs as expected. 
When I try to run this it gives the following error.
AttributeError: 'Zotino' object has no attribute 'voltage_to_mu' 

Having looked in the source code, I don't think the `voltage_to_mu()` is imported into the Zotino source file from the AD9910 file. Could this be why it isn't working?

I'm running ARTIQ v5.6852.c2b4f0cf.beta on Windows 10

### LINKED ISSUE #1443: Documentation: Zotino overflows inside +-10 V range
# Bug Report

## One-Line Summary

Writing voltages close to (but smaller than) +10 V to Zotino leads to overflow in the driver. 

## Issue Details
[Zotino](https://github.com/sinara-hw/Zotino/wiki) is designed to provide +-10 V output range. It seems reasonable to expect voltages in the range [-10, +10] to be supported in the driver. Minimally, voltages in the range (-10, +10) should work. However, values approaching +10 V result in incorrect voltages.

In the current voltage_to_mu implementation voltages from +9.999847412109375 V lead to overflow. I expect this to result in -10 V being applied to the output. (I currently don't have physical hardware access to verify the output voltage.)

https://github.com/m-labs/artiq/blob/85e92ae28cd2de8d68ba15990a33aea1a62a516f/artiq/coredevice/ad53xx.py#L83-L97

### Steps to Reproduce
To reproduce on Zotino, write a voltage close to 10 V to a channel.

```python
self.zotino.write_dac(10 - 1e-5)
```

Or working from `voltage_to_mu()`
```python
offset_dacs=0x2000
vref=5.
voltage=10-1e-5
hex(int(round(0x10000*(voltage/(4.*vref)) + offset_dacs*0x4)))  # gives 0x10000 > 0xffff
```

### Expected Behavior

All voltages between +-10 V should be valid. 
```python
int(0xffff*(voltage/(4.*vref)) + offset_dacs*0x4)
```
seems like the correct conversion. This works for all voltages in the range [-10, +10].

### LINKED ISSUE #1444: Zotino voltage_to_mu overflows silently
# Bug Report

## One-Line Summary
Zotino driver overflows silently.

## Issue Details
The Zotino driver overflows silently. This can lead to vastly incorrect voltages being applied without warning to the user. In combination with https://github.com/m-labs/artiq/issues/1443, this causes silent, incorrect behaviour inside the +-10 V Zotino range.

### Steps to Reproduce
In an experiment, write a voltage close to +10 V to Zotino.
```python
self.zotino.write_dac(10 - 1e-5)
```
This results in incorrect voltages and no warnings/errors

### Expected Behavior

A `ValueError` should be raied on over/underflow.

pathfinder49:
# ARTIQ Pull Request

## Description of Changes
*  Input validation has been added to voltage_to_mu(). (closes #1444)
*  Documentation improvements
    * State valid voltage range (closes #1443)
    * Conversion is only valid for 16-bit members of the family
*  Make `voltage_to_mu()` availible as a `statimethod`. The exising function is maintained for backwads compatibility, but is also exposed as a `staticmethod` of `AD53xx`. (closes #1341)

## Type of Changes

|   | Type |
|  | :bug: Bug fix  |
|   | :hammer: Refactoring  |
| ✓  | :scroll: Docs |

### All Pull Requests

- [x] Update [RELEASE_NOTES.md](../RELEASE_NOTES.md) for new Zotino method

### Code Changes

- [x] Test your changes
    *  Called voltage_to_mu via zotino and imports (both in and outside kernel)
    *  Verified that exception is rasied on over/underflow
- [x] Check, test, and update the [unittests in /artiq/test/](../artiq/test/). 
 There seem to be no Zotino tests.

### Documentation Changes

### Licensing

### DIFF
--- a/artiq/coredevice/ad53xx.py
+++ b/artiq/coredevice/ad53xx.py
@@ -97,11 +97,10 @@ def voltage_to_mu(voltage, offset_dacs=0x2000, vref=5.):
       (default: 0x2000)
     :param vref: DAC reference voltage (default: 5.)
     """
-    volt_lims = [-16 * vref * offset_dacs / (1 << 16),
-                 4 * vref * ((1 << 16) - 1 - 4 * offset_dacs) / (1 << 16)]
-    if voltage < volt_lims[0] or voltage > volt_lims[1]:
+    code = int(round((1 << 16) * (voltage / (4. * vref)) + offset_dacs * 0x4))
+    if code < 0x0 or code > 0xffff:
         raise ValueError("Invalid DAC voltage!")
-    return int(round((1 << 16) * (voltage / (4. * vref)) + offset_dacs * 0x4))
+    return code
 
 
 class _DummyTTL:
--- a/artiq/coredevice/ad53xx.py
+++ b/artiq/coredevice/ad53xx.py
@@ -89,12 +89,18 @@ def voltage_to_mu(voltage, offset_dacs=0x2000, vref=5.):
     voltage when the DAC register is set to mid-scale.
     An offset of V can be used to trim out a DAC offset error of -V.
 
+    Valid voltages are: [-2*vref, + 2*vref - 1 LSB] + voltage offset
+
     :param voltage: Voltage
     :param offset_dacs: Register value for the two offset DACs
       (default: 0x2000)
     :param vref: DAC reference voltage (default: 5.)
     """
-    return int(round(0x10000*(voltage/(4.*vref)) + offset_dacs*0x4))
+    volt_lims = [-16 * vref * offset_dacs / (1 << 16),
+                 4 * vref * ((1 << 16) - 1 - 4 * offset_dacs) / (1 << 16)]
+    if voltage < volt_lims[0] or voltage > volt_lims[1]:
+        raise ValueError("Invalid DAC voltage!")
+    return int(round((1 << 16) * (voltage / (4. * vref)) + offset_dacs * 0x4))
 
 
 class _DummyTTL:
--- a/artiq/coredevice/ad53xx.py
+++ b/artiq/coredevice/ad53xx.py
@@ -157,9 +157,6 @@ class AD53xx:
         self.offset_dacs = offset_dacs
         self.core = dmgr.get(core)
 
-    # make driver accessible (see #1341)
-    voltage_to_mu = staticmethod(voltage_to_mu)
-
     @kernel
     def init(self, blind=False):
         """Configures the SPI bus, drives LDAC and CLR high, programmes
@@ -375,3 +372,14 @@ class AD53xx:
         self.core.break_realtime()
         self.write_offset_mu(channel, 0x8000-offset_err)
         self.write_gain_mu(channel, 0xffff-gain_err)
+
+    @portable
+    def voltage_to_mu(self, voltage):
+        """Returns the 16-bit DAC register value required to produce a given
+        output voltage, assuming offset and gain errors have been trimmed out.
+
+        Valid voltages are: [-2*vref, + 2*vref - 1 LSB] + voltage offset
+
+        :param voltage: Voltage
+        """
+        return voltage_to_mu(voltage, self.offset_dacs, self.vref)
--- a/artiq/coredevice/ad53xx.py
+++ b/artiq/coredevice/ad53xx.py
@@ -80,6 +80,7 @@ def ad53xx_cmd_read_ch(channel, op):
     return AD53XX_CMD_SPECIAL | AD53XX_SPECIAL_READ | (op + (channel << 7))
 
 
+# maintain function definition for backward compatibility
 @portable
 def voltage_to_mu(voltage, offset_dacs=0x2000, vref=5.):
     """Returns the 16-bit DAC register value required to produce a given output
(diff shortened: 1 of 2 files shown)


pathfinder49:
> What's the overhead of the added check and the computation of the bounds? Could you please check that the impact is negligible (just get the rtio counter before and after converting a value gotten from an rpc (so that this isn't optimized away))

I used this code
```python
@kernel
def krun(self):
    number = self.dummy()
    t1 = self.core.get_rtio_counter_mu()
    self.tmp = self.trap_dac.voltage_to_mu(number)
    t2 = self.core.get_rtio_counter_mu()
    print(t2-t1)
```
Without the test I get ~150_000 mu. The current float test gives around ~205_000 mu. Replacing this with an overflow test on the code, gives ~156_000 mu.

> Does the method work in kernels?

It does.

jordens:
The test should be something like this (untested) so that both input and output variables are guaranteed to not be optimized away

```python
def rpc(self, x: TInt32) -> TFloat32:
    return 2.345

@kernel
def run(self):
    dt = 0
    for _ in range(1 << 10):
        i = self.rpc(0)
        dt -= self.core.get_rtio_counter_mu()
        o = self.zotino.voltage_to_mu(i)
        dt += self.core.get_rtio_counter_mu()
        self.rpc(o)
    print(dt >> 10)
```

The method on the object should also supply the correct `offset_dac` values (it has them).
I do think that checking after conversion is much better and faster. Inputs that would be seen as false valid are rather gigantic, right?

pathfinder49:
> The test should be something like this (untested) so that both input and output variables are guaranteed to not be optimized away

I imagined writing the conversion result to an attribute would achieve the same end as passing to an RPC. The attribute is passed to the master after kernel completion.

With the switch to the dac_code check, I beleive this PR to be complete.

jordens:
I think you are right about that.

artiq/coredevice/ad53xx.py
```diff
@@ -151,6 +157,9 @@ def __init__(self, dmgr, spi_device, ldac_device=None, clr_device=None,
         self.offset_dacs = offset_dacs
         self.core = dmgr.get(core)
 
+    # make driver accessible (see #1341)
+    voltage_to_mu = staticmethod(voltage_to_mu)
```
jordens, line 161: `AD53xx.voltage_to_mu` should use `AD53xx.offset_dacs`

pathfinder49: Done. I can rebase, if you like.

artiq/coredevice/ad53xx.py
```diff
@@ -366,3 +372,14 @@ def calibrate(self, channel, vzs, vfs):
...
         self.write_gain_mu(channel, 0xffff-gain_err)
+
+    @portable
+    def voltage_to_mu(self, voltage):
+        """Returns the 16-bit DAC register value required to produce a given
+        output voltage, assuming offset and gain errors have been trimmed out.
+
+        Valid voltages are: [-2*vref, + 2*vref - 1 LSB] + voltage offset
+
+        :param voltage: Voltage
```
jordens, line 383: There is a return value you need to document as well. And the useful information for the input parameter is the unit: "SI Volts".

artiq/coredevice/ad53xx.py
```diff
@@ -108,8 +114,8 @@ def off(self):
 
 
 class AD53xx:
-    """Analog devices AD53[67][0123] family of multi-channel Digital to Analog
-    Converters.
+    """Analog devices AD53[67][0123] family of 16-bit multi-channel Digital to
```
jordens, line 117: There are several 14 bit parts in that family which are also supported by this driver. The words are still 16 bit aligned and word-sized but they are not 16-bit resolution. This comment refers to the resolution.

pathfinder49: I noticed this. However the 14-bit devices are not supported as voltage_to_mu is only valid for the 16-bit resolution family members.

for 16-bit members:
image
for 14-bit members:
image

pathfinder49: I now see, the code is correted in the SPI transfer.
image

jordens:
Thanks

PR #1464, opened 2020-06-15
master: Always write results to HDF5 once run stage is reached

dnadlinger:
Previously, a significant risk of losing experimental results would
be associated with long-running experiments, as any stray exceptions
while run()ing the experiment – for instance, due to infrequent
network glitches or hardware reliability issue – would cause no
HDF5 file to be written. This was especially troublesome as long
experiments would suffer from a higher probability of unanticipated
failures, while at the same time being more costly to re-take in
terms of wall-clock time.

Unanticipated uncaught exceptions like that were enough of an issue
that several Oxford codebases had come up with their own half-baked
mitigation strategies, from swallowing all exceptions in run() by
convention, to always broadcasting all results to uniquely named
datasets such that the partial results could be recovered and written
to HDF5 by a second, manually run cleanup experiment.

This commit addresses the problem at its source, changing the worker
behaviour such that an HDF5 file is always written as soon as run()
starts.

---

If anyone relies on terminating experiments before data is taken in a way
where this would cause too big of a mess, we could also only write files
once a saved dataset has been set, or a certain amount of time has
elapsed. This seemed like the cleaner solution, though, and it is still
possible to kill experiments while they are pending/prepared without
creating unnecessary files.

A possible extension would be to write the exception message/traceback
to HDF5 as well.

---

Not thoroughly tested yet; up for comments.

### DIFF
--- a/artiq/master/scheduler.py
+++ b/artiq/master/scheduler.py
@@ -115,7 +115,6 @@ class Run:
     run = _mk_worker_method("run")
     resume = _mk_worker_method("resume")
     analyze = _mk_worker_method("analyze")
-    write_results = _mk_worker_method("write_results")
 
 
 class RunPool:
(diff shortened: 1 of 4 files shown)


artiq/master/scheduler.py
```diff
@@ -312,11 +311,6 @@ async def _do(self):
                 logger.error("got worker exception in analyze stage of RID %d."
                              " Results will still be saved.", run.rid)
```
sbourdeauducq, line 312: This "Results will still be saved" message should be removed, or there should be a similar message if the ``run`` stage terminates with an exception.

PR #1465, opened 2020-06-16
SUServo: support arbitrary number of Urukuls and add coherent phase mode with per-channel control.

### LINKED ISSUE #1339: suservo: Support more than two Urukuls per Sampler
Currently, the SUServo gateware/driver assume that there are two Urukuls per servo instance, and one Sampler. This isn't unreasonable, as it corresponds to 8 ADC channels and 8 DDSes.

However, in typical experiments, many laser beamlines will end up having more than one wavelength/frequency component, and hence on average more than one DDS channel per ADC channel.

We already support choosing the source ADC channel for each profile, so support for an unequal number of ADC and DDS channels should not be too hard to add.

---

@jordens: Any comments regarding this? I had a quick look at splitting `w.channel` into `w.adc_channel` and `w.dds_channel`, and it doesn't seem too bad, but I'm not sure whether I've noticed all the intricacies of the pipelined design.

pmldrmota:
- Refactoring of the pipelined IIR processing core, accepting any number of ADC input channels and DDS output channels with minimal ressource usage. (Previously restricted to a power of 2.) Resolve #1339.
 - Using half-duplex SPI mode for readback from Urukul CPLDs in NU-mode (DIP switches configured for SUServo operation). To use this feature, the CPLDs require the corresponding update.
 - Extension of the per-channel control interface by a flag , serving to enable coherent DDS phase mode. This works by tracking the DDS-internal phase accumulator as well as the servo runtime to calculate the POW accordingly.
 - Reverse-compatible extension of the channel RTIO addresses for setting the reference time of a channel for coherent phase mode in real time (intended for use in DMA sequences).
 - To ensure coherent phase updates, the io_update line of an Urukul is converted into a SERDES output. The I/O update alignment delays can be configured by writing to servo memory. The optimal alignment delay can be found using the existing method, relying on half-duplex CPLD readback capability.
 - Further, to ensure coherent phase updates, the sync_in line of all Urukuls is connected to a  phy. The optimal sync settings can be found using the existing method, relying on half-duplex CPLD readback capability.
 - Restructuring of the servo RTIO addresses to make space for the I/O update alignment delay memory. This is done efficiently and does not increase the address width.

### DIFF
--- a/artiq/coredevice/ad9910.py
+++ b/artiq/coredevice/ad9910.py
@@ -242,6 +242,10 @@ class AD9910:
 
         :param addr: Register address
         """
+        return self.read32_impl(addr)
+
+    @kernel
+    def read32_impl(self, addr):
         self.bus.set_config_mu(urukul.SPI_CONFIG, 8,
                                urukul.SPIT_DDS_WR, self.chip_select)
         self.bus.write((addr | 0x80) << 24)
@@ -344,7 +348,7 @@ class AD9910:
 
     @kernel
     def set_cfr1(self, power_down=0b0000, phase_autoclear=0,
-                 drg_load_lrr=0, drg_autoclear=0,
+                 drg_load_lrr=0, drg_autoclear=0, phase_clear=0,
                  internal_profile=0, ram_destination=0, ram_enable=0,
                  manual_osk_external=0, osk_enable=0, select_auto_osk=0):
         """Set CFR1. See the AD9910 datasheet for parameter meanings.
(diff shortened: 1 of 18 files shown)


jordens:
From a quick glance this still needs quite some work before it can be reviewed. Please follow the checklist that you deleted from the PR template.
Please split it into meaningful commits. And it would be much easier to review if you also split the PR into parts that can be handled one after the other. I know all this is hard since your code has diverged significantly but it needs to be done. It will significantly accelerate the process.

Issue #1463, opened 2020-06-12
Compilation failure for AD9912 .set introduced


charlesbaynham:
# Bug Report

## One-Line Summary

After commit 2538840756ee76ee49a72dc714361f25be2170ff, running `self.urukul0_ch0.set(100e6)` results in 

```
<artiq>\coredevice\ad9912.py:152:25-152:28: error: the type of this expression cannot be fully inferred
        self.bus.write((pow << 16) | (int32(ftw >> 32) & 0xffff))
                        ^^^
<artiq>\coredevice\ad9912.py:152:25-152:28: note: the expression has type numpy.int?
        self.bus.write((pow << 16) | (int32(ftw >> 32) & 0xffff))
```

### Steps to Reproduce

1. Write the following to a file `test_DDS.py`:
```
from artiq.experiment import EnvExperiment, kernel


class TestDDS(EnvExperiment):

    def build(self, **kwargs):
        self.setattr_device('core')
        self.setattr_device('urukul0_ch0')

    @kernel
    def run(self):
        self.core.reset()
        self.urukul0_ch0.set(100e6)
```

2. Execute `artiq_run test_dds.py` in a directory with your `device_db` for a system with an AD9912 Urukul. 
3. `git bisect` the artiq code to find that before commit 2538840756ee76ee49a72dc714361f25be2170ff this executes as expected, but from this commit onwards you get the error shown above. 

### Your System (omit irrelevant parts)

* Operating System: Windows 10
* ARTIQ version: 2538840756ee76ee49a72dc714361f25be2170ff onwards
* Version of the gateware and runtime loaded in the core device: `5.0.dev+567.g99e490f9;npl2`
* If using Conda, output of `conda list`: [env.txt](https://github.com/m-labs/artiq/files/4771405/env.txt)
* Hardware involved: Kasli rack with AD9912 Urukul

pathfinder49:
Ack, the breaking change should be the `& <mask>` here: https://github.com/m-labs/artiq/blob/1a17d0c869f31163d42b79f9f206c2f5cbcadddd/artiq/coredevice/ad9912.py#L177

According to the [docs](https://m-labs.hk/artiq/manual/compiler.html) this should be inferred as TInt32. I expect wraping the expression in another `int32()` to fix this.

charlesbaynham:
Hmm changing that to 
```
@portable(flags={"fast-math"})
    def turns_to_pow(self, phase):
[...]

PR #1466, opened 2020-06-16
ad9912: Fix typing problem on ad9912

charlesbaynham:
FTW and phase word were ambiguously typed, resulting in failure to compile for Experiments using the AD9912. 

See discussion in #1463

### DIFF
--- a/artiq/coredevice/ad9912.py
+++ b/artiq/coredevice/ad9912.py
@@ -1,5 +1,6 @@
 from numpy import int32, int64
 
+from artiq.language.types import TInt32, TInt64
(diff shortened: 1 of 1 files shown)


jordens:
Thanks

pathfinder49:
@charlesbaynham Could you please confirm: does the ad9912 give the correct output on the hardware? I'm just digging into a related compiler bug that results in incorrect type casting. As far as I can tell the output corruption should still be happening with the type annotation.

charlesbaynham:
@pathfinder49 Yes, both frequency and attenuation seem to behave as expected as viewed on a scope. I haven't tested phase

Issue #1433, opened 2020-03-09
zombie runs


jordens:
# Bug Report

## One-Line Summary

Sometimes a run remains stuck in the "pending" stage and becomes invisible to other runs and is un-deletable.

### Steps to Reproduce

No idea how to reproduce reliably.

In opticlock we have a couple tweak-up scripts hooked up to the F-keys. These scripts only exercise software functions (via controllers) and don't touch the coredevice. They also run in their own pipeline. (There are other runs running in other pipelines, among them runs that do interact with the coredevice). But activity or state changes in other pipelines are not needed.

Depending on conditions they might be submitted at reasonably high rates (~Hz) and will back up somewhat (maybe ~10 runs deep). But they should always be executed in sequence and removed when done.

### Expected Behavior

The runs execute according to the pipeline rules and be removed then.

### Actual (undesired) Behavior

* Sometimes runs get stuck in the pending stage and never run and can't be deleted.

```
+-------+----------+----------------+------+----------+----------+----------------------+-------------+
|  RID  | Pipeline |     Status     | Prio | Due date | Revision |         File         |  Class name |
+-------+----------+----------------+------+----------+----------+----------------------+-------------+
| 88501 |  motor   |    pending     |  0   |          |   N/A    |   tweak_motors.py    |  Tweak1Down |
| 94501 |  motor   |    pending     |  0   |          |   N/A    |   tweak_motors.py    |   Tweak0Up  |
+-------+----------+----------------+------+----------+----------+----------------------+-------------+
```

* Graceful or force deletion have no effect.
* There are no corresponding artiq worker processes (anymore).
* The run did not fail but also didn't leave a h5 file around. It looks like it never got out of pending.
* Nothing in the logs (at INFO level).
* Interestingly, other runs (same pipeline, same priority) have no problem bypassing and skipping the zombies. This should not be the case AFAIK.
* There does not seem to be any effect (side or not) of the zombies.
* This is rare. I've seen it maybe four times over a year and 50k rids including the two runs above.
* I have no idea what to make of the exact 6000 rid difference between these two occurrences. Would be a weird coincidence even given confirmation bias (any multiple of a reasonable power of ten or two would be surprising).
* An `artiq_master` restart obviously clears them.

* I haven't looked into debugging the running artiq_master process and I don't know whether it's easy to do (with `pdb`) and what to look for.

### Your System (omit irrelevant parts)

* Operating System: Linux, Debian buster
[...]

PR #1470, opened 2020-06-19
master/scheduler: Fix priority/due date precedence order when waiting to prepare

dnadlinger:
@pmldrmota: This should do it.

### DIFF
--- a/artiq/master/scheduler.py
+++ b/artiq/master/scheduler.py
@@ -87,17 +87,12 @@ class Run:
             self._notifier[self.rid]["status"] = self._status.name
         self._state_changed.notify()
 
-    # The run with the largest priority_key is to be scheduled first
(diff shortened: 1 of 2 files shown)


sbourdeauducq:
Should this be mentioned in the release notes and/or applied to the release-5 branch?

dnadlinger:
I was torn, but left it off because it took several years for a situation to appear where this happens in practice, so it neither seemed to be a highlight or a breaking change. Happy to add a one-liner if you want to go for more complete release-notes.

And yes, this is strictly a bug fix and can go onto release-5.

jordens:
Is it possible that this fixes #1433?

dnadlinger:
This should only apply for due-date runs, and even those would have been prepared correctly once the pool status changed (e.g. when submitting a deletion).

PR #1479, opened 2020-06-30
runtime: Stop kernel CPU before restarting comms CPU on panic

dnadlinger:
Before, the system would enter a boot loop when a panic occurred
while the kernel CPU was active (and panic_reset == 1), as
kernel::start() for the startup kernel would panic.

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -21,6 +21,8 @@ Highlights:
      be brought up Ctrl-P (Ctrl+Return to immediately submit the selected entry
      with the default arguments).
 * Experiment results are now always saved to HDF5, even if run() fails.
+* Core device: ``panic_reset 1`` now correctly resets the kernel CPU as well if
+  communication CPU panic occurs.
 
 Breaking changes:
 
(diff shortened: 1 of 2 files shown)


dnadlinger:
@sbourdeauducq: This is not yet tested, as I still have no clue where the original panic came from, but this should be an obvious fix to avoid the boot loop mentioned in the commit message. Or is the comms CPU reset supposed to also reset the kernel CPU?

dnadlinger:
@sbourdeauducq: This is not yet tested, as I still have no clue where the original panic came from, but this should be an obvious fix to avoid the boot loop mentioned in the commit message. Or is the comms CPU reset supposed to also reset the kernel CPU?

sbourdeauducq:
>  Or is the comms CPU reset supposed to also reset the kernel CPU?

There are some loose ends like that with the soft reboot. A better solution would be to use the ICAP to reload the gateware completely.

sbourdeauducq:
This is not as hard as it sounds as we can simply send a "reboot from flash" command to the FPGA. I have done that on Spartan-6 before and there were no particular difficulties.

dnadlinger:
Agreed, that would be nice, but I'll leave it on the collective todo list for now, as I have zero experience getting that to work.

PR #1480, opened 2020-07-03
I2C: Duplicate TCA9548 control byte

DonaldKellett:
Duplicate control byte in I<sup>2</sup>C, in accordance with "7. Parameter Measurement Information" in the [TCA9548 datasheet](https://www.ti.com/lit/ds/symlink/tca9548a.pdf?ts=1593747422231&ref_url=https%253A%252F%252Fwww.google.com%252F)

### DIFF
--- a/artiq/firmware/libboard_misoc/i2c.rs
+++ b/artiq/firmware/libboard_misoc/i2c.rs
@@ -183,8 +183,11 @@ mod imp {
         if !write(busno, address << 1)? {
             return Err("PCA9548 failed to ack write address")
         }
-        if !write(busno, channels)? {
-            return Err("PCA9548 failed to ack control word")
+        // Duplicate control byte: one for SCL, one for SDA
+        for _ in 0..2 {
+            if !write(busno, channels)? {
+                return Err("PCA9548 failed to ack control word")
+            }
         }
         stop(busno)?;
         Ok(())


jordens:
What's the thinking here? Does it fix something?
It directly contradicts the vast majority of the datasheet and the behavior we have implemented and observed in the last years. The register is one byte and there is no mention of a "P-port".
They probably just copied the definition of i2c from some other device that had a 16 bit p port.

sbourdeauducq:
From https://www.ti.com/lit/ds/symlink/tca9548a.pdf :
image

sbourdeauducq:
Trying to figure out the Si5324 I2C issue on Kasli 2.0...

jordens:
Have a look at the other pages of the datasheet. Especially page 17, figure 10, and the text of the datasheet stating explicitly how to do a write. As I said, you are probably falling for a copy paste error.

marmeladapk:
From the TCA documentation:

> There is no limit to the number of bytes sent, but the last byte sent is what is in the register.

> If multiple bytes are received by the TCA9548A, it saves the last byte received.

So this changes nothing in operation of the mux. @jordens is right. 

Si5324 issue is due to low level voltage not being low enough.

sbourdeauducq:
OK, reverted.

PR #1484, opened 2020-07-14
firmware/i2c: rewrite I2C implementation

jordens:
* Never drive SDL or SDA high. They are specified to be open
  collector/drain and pulled up by resistive pullups. Driving
  high fails miserably in a multi-master topology (e.g. with
  a USB I2C interface). It would only ever be implemented to
  speed up the bus actively but that's tricky and completely
  unnecessary here.
* Make the handover states between the I2C protocol phases (start, stop,
  restart, write, read) well defined. Add comments stressing those
  pre/postconditions.
* Add checks for SDA arbitration failures and stuck SCL.
* Remove wrong, misleading or redundant comments.

# ARTIQ Pull Request

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |
|    | :sparkles: New feature |
|   | :scroll: Docs |

### All Pull Requests

- [ ] Close/update issues.

### Code Changes

### Documentation Changes

### Git Logistics

  ```
  
  ```

### Licensing

### DIFF
--- a/artiq/firmware/libboard_misoc/i2c.rs
+++ b/artiq/firmware/libboard_misoc/i2c.rs
@@ -14,6 +14,12 @@ mod imp {
         }
     }
 
+    fn scl_i(busno: u8) -> bool {
+        unsafe {
+            csr::i2c::in_read() & scl_bit(busno) != 0
+        }
+    }
+
     fn sda_oe(busno: u8, oe: bool) {
         unsafe {
             let reg = csr::i2c::oe_read();
@@ -49,13 +55,10 @@ mod imp {
     pub fn init() -> Result<(), &'static str> {
         for busno in 0..csr::CONFIG_I2C_BUS_COUNT {
             let busno = busno as u8;
-            // Set SCL as output, and high level
-            scl_o(busno, true);
-            scl_oe(busno, true);
-            // Prepare a zero level on SDA so that sda_oe pulls it down
-            sda_o(busno, false);
-            // Release SDA
+            scl_oe(busno, false);
             sda_oe(busno, false);
+            scl_o(busno, false);
+            sda_o(busno, false);
 
             // Check the I2C bus is ready
             half_period();
@@ -63,9 +66,9 @@ mod imp {
             if !sda_i(busno) {
                 // Try toggling SCL a few times
                 for _bit in 0..8 {
-                    scl_o(busno, false);
+                    scl_oe(busno, true);
                     half_period();
-                    scl_o(busno, true);
+                    scl_oe(busno, false);
                     half_period();
                 }
             }
@@ -73,6 +76,10 @@ mod imp {
             if !sda_i(busno) {
                 return Err("SDA is stuck low and doesn't get unstuck");
             }
+            if !scl_i(busno) {
+                return Err("SCL is stuck low and doesn't get unstuck");
+            }
+            // postcondition: SCL and SDA high
         }
         Ok(())
     }
@@ -81,11 +88,17 @@ mod imp {
         if busno as u32 >= csr::CONFIG_I2C_BUS_COUNT {
             return Err(INVALID_BUS)
         }
-        // Set SCL high then SDA low
-        scl_o(busno, true);
-        half_period();
+        // precondition: SCL and SDA high
+        if !scl_i(busno) {
+            return Err("SCL is stuck low and doesn't get unstuck");
+        }
+        if !sda_i(busno) {
+            return Err("SDA arbitration lost");
+        }
         sda_oe(busno, true);
         half_period();
+        scl_oe(busno, true);
+        // postcondition: SCL and SDA low
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


sbourdeauducq:
>     * Never drive SDL or SDA high.

I believe only SCL was driven high (because this was never meant to be used with multiple masters) - correct?

sbourdeauducq:
With this new code, what happens if the FPGA talks on the I2C bus at the same time as another master?

jordens:
Neither SCL nor SDA should ever be driven high on a normal i2c bus. Single master or not. It just becomes worse with multiple masters.
The same thing will happen as before. One of them will loose arbitration and fail it's transaction. But with the new code at least there is no doubly driven line anymore. And additionally the new code will recognize a couple possible cases of arbitration failure.

sbourdeauducq:
> Neither SCL nor SDA should ever be driven high on a normal i2c bus. Single master or not. It just becomes worse with multiple masters.

What is the issue with driving SCL high in a single-master system?

> One of them will loose arbitration and fail it's transaction.

And it seems the possible failure modes include silent corruption of a transaction's data... Multi-master I2C doesn't sound like such a great idea.

sbourdeauducq:
https://www2.renesas.cn/cn/zh/doc/products/mpumcu/apn/002/rej06b0634_h8sap.pdf has some details about multi-mastering that seems somewhat reliable, though it sounds complicated and I remain a bit skeptical of multi-master I2C. Do we trust other masters to implement something like that correctly?

jordens:
* It's the responsibility of the original implementation/implementer to argue its correctness. By demanding proof how the clearly faulty old implementation is problematic you reverse that responsibility.
* The old implementation violates the I2C protocol.
* For your amusement and doing the work that the reviewer or original implementer should have been done long ago: It already blows up if any slave decides to do clock stretching. The new implementation doesn't support clock stretching either but it prevents doubly driven signals and potential damage.
* I don't understand how your complaints about multi-master affect the merits or validity of this PR. The old implementation prevented multi-master from ever working as demonstrated. The new one enables it for the dominant use cases.

I am in the uncomfortable position of having to defend a supposedly advantageous, correct, and compliant PR because of what appears to be you wanting to save face. This has a chilling effect.
Since there are no arguments against the PR, I'll either close it or merge it.

sbourdeauducq:
The current implementation is intended for simple single-master situations without clock stretching. It is a simple design that doesn't pretend to support any advanced I2C features. It is something we added because Daniel wanted some basic expanders supported on the KC705, and it was later reused for other simple devices like the Si5324 and EEPROMs - none of these are mastering the bus or stretching the clock. For these purposes, as far as I can tell, it is adequate and not "fundamentally broken" or "faulty".
I'm trying to understand what you want to achieve here. If it is simply a step towards supporting all the bells and whistles of I2C (full multi-master and clock stretching support will take a lot more work and testing of third-party devices and software, as we see), and you think that supporting these are a good idea, then just say that.

marmeladapk:
@sbourdeauducq if my understanding of this code is correct then this PR will make Artiq support more edge cases with I2C. It will save us (either someone from your team or from ours) time in the future when we encounter some chip with more fancy I2C implementation.

jordens:
Driving SCL high is fundamentally broken and faulty. It isn't and wasn't adequate by any definition. I don't understand why you claim that. Everybody makes mistakes. You won't loose credibility by admitting that the old code was bad.

This PR fixes a serious bug on all our multi-master implementations. That's all Kaslis, Sayma, Metlino. You don't need to wait for a clock stretching chip or a corruption-proof multi-master protocol to see that.

sbourdeauducq:
> Driving SCL high is fundamentally broken and faulty. It isn't and wasn't adequate by any definition. I don't understand why you claim that.

Simply because, for unidirectional signals (again: this code was intended for simple situations), push-pull is the default choice and more common than open drain and pull-up. Call it a mistake if you wish...

marmeladapk:
@sbourdeauducq Driving SCL high also broke FTDI I2C on Kasli v2.0 when FPGA was programmed.

hartytp:
:(

sbourdeauducq:
Okay, but I don't know why you want those complicated and fragile I2C multi-mastering schemes. For accessing I2C on a EEM crate, we could have done it through the FPGA like we currently do it for the SPI flash. And the MMC on µTCA does not really need to access the same I2C bus as the FPGA (or even better: there should be no MMC at all).

marmeladapk:
@sbourdeauducq Previous code didn't allow to use I2C over USB when ARTIQ was flashed. I consider this as a bug, since everything else on the board suggests that it should be possible.

sbourdeauducq:
With the current code it will only work "most of the time". If by chance you access the I2C bus at the same time as ARTIQ, transfers will get corrupted.

The ARTIQ I2C implementation was never meant to support I2C multi-mastering, and I do not see why it should, because I2C multi-mastering:
* is a lot more complicated to implement correctly than a "I2C proxy bitstream" single-master scheme, similar to the one we use for SPI flash, which would give a similar level of access to the I2C buses of a crate from USB.
* is inherently more fragile: it only takes one non-compliant master for the whole bus to break. And I2C multi-mastering is not a trivial protocol as I mentioned above, so we should expect problems.
* does not provide any clear advantages (is there a case for e.g. flashing I2C EEPROMs without interrupting the operation of ARTIQ?)

PR #1509, opened 2020-08-25
metlino: support EEPROM & EUI48 reading

HarryMakes:
# ARTIQ Pull Request

## Description of Changes

This adds firmware code for Metlino to read the MAC address from the EEPROM's EUI-48 address.

## Details

It is worth noting that AT24MAC402, which is the EEPROM connected to Port 5 of Metlino's I2C switch IC43, has a specific sequence of I2C transaction sequence, which consists of a specific slave address byte. Please refer to Fig. 12-5 of the [datasheet](http://ww1.microchip.com/downloads/en/devicedoc/Atmel-8807-SEEPROM-AT24MAC402-602-Datasheet.pdf) for details. Also note that the EUI-48 address on this EEPROM is specified as: FC-C2-3D-xx-xx-xx.

A minor edit on the firmware code for both Kasli and Metlino has been applied: if I2C transactions on the EEPROM fails, instead of using a default MAC address, the firmware will panic.

### DIFF
--- a/artiq/firmware/libboard_misoc/i2c_eeprom.rs
+++ b/artiq/firmware/libboard_misoc/i2c_eeprom.rs
@@ -1,6 +1,7 @@
 use i2c;
 
-/// [Hardware manual](http://ww1.microchip.com/downloads/en/DeviceDoc/24AA02E48-24AA025E48-24AA02E64-24AA025E64-Data-Sheet-20002124H.pdf)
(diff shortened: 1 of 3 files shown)


sbourdeauducq:
As per earlier Mattermost discussion with @marmeladapk this will have to go through the MMC. That bus is continuously used to monitor temperatures, and there is currently no support for I2C multi-master anywhere AFAICT. Good to know the EEPROM is working, though.

PR #1513, opened 2020-08-31
metlino_sayma_ttl: Example of Metlino-master, Sayma-satellite system

HarryMakes:
# ARTIQ Pull Request

## Description of Changes

This adds Sayma's gateware support as a satellite controlling 2 DIO cards via an FMC-VHDCI-LVDS adaptor. This is demonstrated with an example containing a Metlino master and Sayma satellite.

## Details

The modification of Sayma RTL is to import the existing RTL library for the FMC-VHDCI-LVDS. A minor error in the library has been fixed.

The additional ARTIQ example was created based on `artiq_sinara_tester`. However, it is worth noting that certain functions originally from Sinara tester that does not run in the kernel need to be run in the kernel, as per testing. This might be due to how the list of LED and TTL devices gets initialised in my script, which might have differences from Sinara tester which I have overlooked. Nevertheless, this example suffice the requirement of demonstrating the TTL outputs in a master-satellite topology between Metlino and Sayma.

### DIFF
--- a/artiq/gateware/fmcdio_vhdci_eem.py
+++ b/artiq/gateware/fmcdio_vhdci_eem.py
@@ -19,8 +19,9 @@ def _get_connectors():
         for j, pair in enumerate(eem_fmc_connections[i]):
             for pn in "n", "p":
                 cc = "cc_" if j == 0 else ""
+                lpc_cc = "CC_" if eem_fmc_connections[i][j] in (0, 1, 17, 18) else ""
                 connections["d{}_{}{}".format(j, cc, pn)] = \
-                    "LPC:LA{:02d}_{}{}".format(pair, cc.upper(), pn.upper())
+                    "LPC:LA{:02d}_{}{}".format(pair, lpc_cc, pn.upper())
         connectors.append(("eem{}".format(i), connections))
     return connectors
 
--- a/artiq/gateware/targets/sayma_amc.py
+++ b/artiq/gateware/targets/sayma_amc.py
@@ -12,8 +12,10 @@ from misoc.interconnect.csr import *
 from misoc.targets.sayma_amc import *
 
 from artiq.gateware.amp import AMPSoC
+from artiq.gateware import eem
 from artiq.gateware import rtio
 from artiq.gateware import jesd204_tools
+from artiq.gateware import fmcdio_vhdci_eem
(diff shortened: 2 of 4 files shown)


artiq/gateware/fmcdio_vhdci_eem.py
```diff
@@ -19,8 +19,9 @@ def _get_connectors():
         for j, pair in enumerate(eem_fmc_connections[i]):
             for pn in "n", "p":
                 cc = "cc_" if j == 0 else ""
```
sbourdeauducq, line 21: This variable is no longer used, remove

artiq/gateware/targets/sayma_amc.py
```diff
@@ -284,7 +286,7 @@ def __init__(self, platform, sys_crg, jesd_crg, dac):
 
 class Satellite(SatelliteBase):
     """
-    DRTIO satellite with local DAC/SAWG channels.
+    DRTIO satellite with local DAC/SAWG channels, as well as FMC channels.
```
sbourdeauducq, line 289: TTL channels via FMC and VHDCI carrier

artiq/gateware/targets/sayma_amc.py
```diff
@@ -307,7 +309,7 @@ def __init__(self, jdcg_type, **kwargs):
         self.csr_devices.append("slave_fpga_cfg")
         self.config["SLAVE_FPGA_GATEWARE"] = 0x200000
 
-        rtio_channels = []
+        self.rtio_channels = rtio_channels = []
```
sbourdeauducq, line 312: Where is this used?

artiq/gateware/fmcdio_vhdci_eem.py
HarryMakes: It is used for constructing the `eem?` connectors that the `add_std()` functions for each class in `artiq.gateware.eem` can use when requesting.

artiq/gateware/targets/sayma_amc.py
HarryMakes: Used by `add_std()` functions in `artiq.gateware.eem`

artiq/gateware/fmcdio_vhdci_eem.py
sbourdeauducq: OK sorry I missed it

PR #1514, opened 2020-08-31
build_soc: override identifier_str only for gateware

astro:
Actually, we want to override the identifier_str just for the gateware but not the for the firmware.

This fixes the `software ident` output. Should we change the arg into `gateware_identifier_str` in all callers for more clarity?


sbourdeauducq:
> Should we change the arg into `gateware_identifier_str` in all callers for more clarity?

Yes, please send another PR

PR #1541, opened 2020-11-04
compiler: Raise exception on failed assert()s rather than panic

### LINKED ISSUE #1539: artiq.test.coredevice.test_embedding.ArrayQuotingTest fails on Zynq
```
ERROR: test_quoting (artiq.test.coredevice.test_embedding.ArrayQuotingTest)
  File "/nix/store/dqfdq864hrn1v8xs3glcrgd7zqcim70h-python3.8-artiq-6.7477.d672d2fc.beta/lib/python3.8/site-packages/artiq/test/coredevice/test_embedding.py", line 462, in test_quoting
    self.create(_ArrayQuoting).run()
  File "/nix/store/dqfdq864hrn1v8xs3glcrgd7zqcim70h-python3.8-artiq-6.7477.d672d2fc.beta/lib/python3.8/site-packages/artiq/language/core.py", line 54, in run_on_core
    return getattr(self, arg).run(run_on_core, ((self,) + k_args), k_kwargs)
  File "/nix/store/dqfdq864hrn1v8xs3glcrgd7zqcim70h-python3.8-artiq-6.7477.d672d2fc.beta/lib/python3.8/site-packages/artiq/coredevice/core.py", line 135, in run
    self.comm.load(kernel_library)
  File "/nix/store/dqfdq864hrn1v8xs3glcrgd7zqcim70h-python3.8-artiq-6.7477.d672d2fc.beta/lib/python3.8/site-packages/artiq/coredevice/comm_kernel.py", line 363, in load
    raise LoadError(self._read_string())
artiq.coredevice.comm_kernel.LoadError: core1 failed to process data

```

Device log:
```
[  3093.860055s] ERROR(runtime::kernel::core1): failed to load shared library: symbol lookup error: abort
```

dnadlinger:
This allows assert() to be used on Zynq, where abort() is not
currently implemented for kernels. Furthermore, this is arguably
the more natural implementation of assertions on all kernel targets
(i.e. where embedding into host Python is used), as this way, the
exception information actually makes it to the user.

Since this does not implement printing of the subexpressions, I
left the old print+abort implementation as default for the time
being.

GitHub: Fixes #1539.

---

Needs testing still. (Even if Zynq was the specific motivation here, I've wanted something like this for a while, as assert()s are phenomenally undebuggable with core log forwarding as unreliable as it currently is.)

### DIFF
--- a/artiq/compiler/targets.py
+++ b/artiq/compiler/targets.py
@@ -80,6 +80,9 @@ class Target:
         determined from data_layout due to JIT.
     :var now_pinning: (boolean)
         Whether the target implements the now-pinning RTIO optimization.
+    :var raise_assertion_errors: (bool)
+        Whether to raise an AssertionError on failed assertions or abort/panic
+        instead.
     """
     triple = "unknown"
     data_layout = ""
@@ -87,6 +90,7 @@ class Target:
     print_function = "printf"
     little_endian = False
     now_pinning = True
+    raise_assertion_errors = False
 
     tool_ld = "ld.lld"
     tool_strip = "llvm-strip"
@@ -276,6 +280,10 @@ class CortexA9Target(Target):
     print_function = "core_log"
     little_endian = True
     now_pinning = False
+
+    # Support for marshalling kernel CPU panics as RunAborted errors is not
+    # implemented in the ARTIQ Zynq runtime.
+    raise_assertion_errors = True
 
     tool_ld = "armv7-unknown-linux-gnueabihf-ld"
     tool_strip = "armv7-unknown-linux-gnueabihf-strip"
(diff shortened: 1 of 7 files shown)


artiq/compiler/targets.py
```diff
@@ -277,6 +281,10 @@ class CortexA9Target(Target):
     little_endian = True
     now_pinning = False
 
+    # Support for mashalling kernel CPU panics as RunAborted errors is not
```
sbourdeauducq, line 284: marshalling

sbourdeauducq:
Would there be an issue with raising AssertionError in all cases and simplifying this code? Is the printing of the expressions really that useful?

Issue #1345, opened 2019-07-30
hdf5 results: use attributes for small datasets


jordens:
# ARTIQ Feature Request

## Problem this request addresses

A HDF5 dataset uses at least a few kB of disk space while most ARTIQ datasets in practice are smaller than the minimum HDF5 dataset overhead and many (e.g. scalars) are even smaller than their key. This is an inefficient use of disk space.

## Describe the solution you'd like

* Investigate `H5D_COMPACT` layout.
* Investigate storing ARTIQ datasets as HDF5 attributes.
* Investigate limitations w.r.t. maximum object header size (64kB) and "dense attributes" as (implicit?) extension.
* Investigate when and how an ARTIQ dataset should be an HDF5 dataset and when an attribute. Merging of the key spaces.

## Additional context

* #1272 for reorganizing hdf5 dataset groups

airwoodix:
Could we also add compression to the discussion? As was pointed out on #m-labs, there's support for compression filters for hdf5 datasets. I did [some tests](https://github.com/airwoodix/hdf5-compression-tests) to get an impression on readily available and working filters, as well as crude performance.
It seems realistic to me to enable Zstandard compression by default on all array datasets (e.g. images), assuming that smaller data is stored as attributes. But this requires further testing, especially on Windows. More / better testing very welcome.
At least exposing h5py `create_dataset` options would be great.

Issue #1272, opened 2019-02-10
hdf5 results: merge "archive" and "datasets" groups


jordens:
# ARTIQ Feature Request

## Problem this request addresses

Currently in the HDF5 results, there are get() datasets in the "archive" group and set() datasets in the "datasets" group. This is a bit confusing and doesn't seem necessary since we assume/warn/enforce that get and set datsets share a namespace and an entry is never both. 

## Describe the solution you'd like

Merge the groups. The browser already builds a merged namespace of the two groups.

## Additional context

Maybe #1171

airwoodix:
I also find this very confusing.
@hartytp @jordens  Are there code bases that heavily rely on the `datasets`/`archive` split? If so, what's the use case?
Merging the two sets would also nicely simplify the code in `master.worker_db`.

hartytp:
@airwoodix @dpn is a better person to ask than me (I typically only interact with HDF5 results via tooling he wrote so am not really conscious of it's internal structure).

jordens:
I think it's only `browser` and maybe the experiment load-from-hdf that would look at it. As I said. I don't think it's necessary. It's a remnant from before #1171

Issue #505, opened 2016-07-11
Save to hdf5 before experiment is complete


dleibrandt:
As far as I can tell, the hdf5 files are not written until the end of an experiment.  Is this correct?

If not, how do I save to hdf5 before the experiment is complete?

If so, I see this as a significant problem.  We don't want to lose all the data if our computer crashes in the middle of a run.  Also, we frequently run analysis scripts in the middle of experiment runs, which might be matlab or something else that doesn't play nice with artiq.

jordens:
Correct.

You are free to open and write files to do checkpointing whenever you want.
The code that does the hdf5 file writing is here: https://github.com/m-labs/artiq/blob/master/artiq/master/worker_impl.py#L230
You should be able to open that same file and write to it before `write-results` happens. IIRC the scheduler device knows about `rid`.
Be aware that, if you want to read from another process with the hdf5 file being written to, single-writer-multiple-reader requires a bit of special care in h5py.
If you need automatic checkpointing in the background we'd need a full specification.

jordens:
http://docs.h5py.org/en/latest/swmr.html

sbourdeauducq:
> If you need automatic checkpointing in the background we'd need a full specification.

May not be fully in the background. An API such as ``Experiment.write_results()`` that is called automatically after ``Experiment.analyze()`` and may be called explicitly at any time by the user may be a good choice.

jordens:
Yes. Or open the results file early (in `prepare()`) and expose the handle so that partial or intermediate things can be written without having to push all datasets.

dleibrandt:
I think ideally I would like:
1. All datasets are automatically written to the hdf5 near the beginning of the experiment (maybe at the end of prepare or the start of run?)
2. Periodically, datasets that have been modified can be re-written via some user called function `Experiment.update_hdf5()` (ideally only the parts of the datasets that have actually been modified would be written for speed reasons)
3. `Experiment.update_hdf5()` would be called automatically at the end of analyze()

If number 2 is too hard to implement, it might be OK if the user has the option to either write all the datasets or specify which datasets to write.

dnadlinger:
#1464 should improve the situation considerably by always saving to HDF5 once the run stage is reached, i.e. even if it is finished with an exception. If the user crashes/deadlocks the worker process before the exception handler runs, data can of course still be lost, so further checkpointing might still be a good idea.

PR #1544, opened 2020-11-16
datasets: support compression in HDF5 archives

airwoodix:
# ARTIQ Pull Request

## Description of Changes

This patch adds opt-in compression for datasets in HDF5 archives.

The `allow_compression` flag to `HasExperiment.set_dataset` decides whether compression would be acceptable for the dataset. If this is the case and the dataset type is a numpy array and it's size is larger than 300 (*), the data is passed through HDF5's `gzip` filter when stored/retrieved.

**This patch breaks the format of the dataset persistence file** such that persistent datasets can keep track of their compression hint.

The choice of compression algorithm/level is motivated by:

- `zlib` is part of the HDF5 distribution, so available everywhere where HDF5 is available
- `compression_level=9`: if the user wants to compress, then size matters

Testing status:

- (updated) unit tests in `test_datasets.py` pass
- full test suite passes (with fix in `test_scheduler:test_pause`, see 14b925e)
- tested with simple experiments on an `artiq_master`. Confirmed compression ratio on resulting HDF5 file (ca. 10 on 1024x1024 integral random data).
- tested validity of the resulting HDF5 file with `h5ls`/`h5dump` and [ViTables](https://vitables.org/)
- dashboard launches with no errors, loads persistent datasets, shows the compression hint, dataset editing works

(*) threshold taken from [flammkuchen](https://github.com/portugueslab/flammkuchen/blob/master/flammkuchen/hdf5io.py#L149-L152)

### Related Issue

#1345 

## Type of Changes

|   | Type |

### All Pull Requests

### Code Changes

### Documentation Changes

### Git Logistics

  ```
  
  ```

### Licensing


sbourdeauducq:
This needs documentation and a release notes update (especially since the format changed).

jordens:
Nice work. This is really useful for the neutrals and camera images.
I have the feeling that it would be better to explicitly have the user do `compress=True` instead of "allowing" and then doing it based on somewhat arbitrary size rules. Pretty sure, the size rule is inaccurate for non-f32/f64/i32 data types and depends a lot on the actual data (which only the user knows about). Doing it explicitly also mirrors and defers the semantics of `group.create_dataset()`. In #1345 you did express desire to expose `create_dataset()` more directly.

airwoodix:
Fair point. I was probably trying to put too many safeguards.
Then one could also expose an nice interface to `compression` and `compression_opts` similarly to [what PyTables does](https://www.pytables.org/usersguide/libref/helper_classes.html#tables.Filters).
Exposing the compression level is probably useful. I find exposing another compressor than gzip dangerous because it adds dependencies on the machines doing the analysis (and the one running the master). (Optionally) pulling PyTables might not be that bad though.
Overall, I like the simplicty of `compress=True` and no-worries/no-options for the end-user but this might get annoying for some.
What's your opinion on this?

jordens:
If they want to use a different compression they should be allowed to do so.
Maybe `group.create_dataset(k, data=v["value"], **dataset_options)` and allow the user to do `set(k, v, dataset_options=dict(compress=True, ...))`?
But we are now fighting the design limitations of the dataset database...

Alternatively, as you indicated, expose the hdf5/group handle to the user early in the experiment. That would enable #505 and give them full transparent access to all the hdf5 functionality (instead of having to wrap it through the dataset api).

airwoodix:
> But we are now fighting the design limitations of the dataset database...

Yes. Already in the fact that one cannot easily set (smart) defaults for the dataset database (wrapping the set_dataset call surely works but is not very elegant).

Sounds like a lot of room for discussion. I opened a new issue for that #1545. I guess I'll leave this PR on hold until #1545 converges.

airwoodix:
Considering that #1545 would create too many breaking changes, I opted for forwarding extra keyword arguments to ``set_dataset`` to the HDF5 backend. This covers all compression / filtering use cases but none of the extras listed in #1545. The dashboard shows which filters are applied. I updated the tests.

We can debate whether `set_dataset` should glob and forward keyword arguments or take a dictionary. Since I anticipate compression to dominate the use cases, I find the syntax ``set_dataset(name, value, compression="gzip")`` nicer than ``set_dataset(name, value, hdf5_options=dict(compression="gzip"))``. ARTIQ specific keywords should anyway be listed explicitly. I guess that name collisions are unlikely to happen.

airwoodix:
@jordens would you support merging this version (I'd rebase beforehand), exposing the h5py API through the the dataset API?
I know it doesn't help with #505 but am not convinced by exposing the ``h5py.File`` object directly, because it would be a major rework of the master ``worker_impl``. Or do you have an easier solution? #1545 (or similar) is way too invasive I guess.

artiq/dashboard/datasets.py
```diff
@@ -83,16 +83,29 @@ def get_edit_widget_value(self):
 
 
 class Model(DictSyncTreeSepModel):
-    def __init__(self,  init):
-        DictSyncTreeSepModel.__init__(self, ".",
-                                      ["Dataset", "Persistent", "Value"],
-                                      init)
+    def __init__(self, init):
+        DictSyncTreeSepModel.__init__(
+            self, ".", ["Dataset", "Persistent", "HDF5 filters", "Value"], init
```
jordens, line 88: I would opt for leaving out `HDF5 filters` here. That table is already very dense and hard to read because the `Persistent` column is a visual barrier that makes it hard to associate key and the corresponding value. Now with a potentially wide filter column in between this becomes even trickier.

jordens (changes requested):
Other than the UI change, I think this provides a good benefit.

One thing that would make life a lot easier is an automatic and silent conversion of the `dataset_db.pyon` format. I'm fine if this is only forward conversion and not backward.

airwoodix:
> One thing that would make life a lot easier is an automatic and silent conversion of the ``dataset_db.pyon`` format.

I don't think that I get your point. Do you mean enabling, e.g., compression in ``dataset_db.pyon``? What would be the use case?

jordens:
> I don't think that I get your point. Do you mean enabling, e.g., compression in dataset_db.pyon? What would be the use case?

I meant an automatic conversion of the `device_db.pyon` before this PR to the new one. But now that I look at it, the format doesn't seem to change. So scratch my original request.
But that again makes me wonder: should the `hdf5_options` be persisted together with persistent datasets?

airwoodix:
Thanks for the review.
I cleaned everything up, squashed and rebased on current ``master``in  d8b1e59.
Also added unit tests for the internal dataset representation (``test_dataset_db.py``). This is particularly relevant because applets depend on that.

> But that again makes me wonder: should the hdf5_options be persisted together with persistent datasets?

They are. Datasets are now represented by mappings ``name: {"value": foo, "persist": bool, "hdf5_options": dict}`` and saved as such (including ``persist``) in ``dataset_db.pyon``.

> I meant an automatic conversion of the device_db.pyon before this PR to the new one.

As a frontend tool? or just a snippet in the changelog/manual? It's essentially a three-liner (sketchy, not tested):

```python
from sipyco import pyon
from artiq.master.databases import make_dataset

datasets = pyon.load_file("dataset_db.pyon")
new = {name: make_dataset(value=value, persist=True) for name, value in datasets.items()}
pyon.store_file("dataset_db.pyon", new)
```

A frontend tool could offer in-place/backup options etc, but is that really necessary?

jordens:
> A frontend tool could offer in-place/backup options etc, but is that really necessary?

I'm looking at the case where you run an experiment and want to upgrade ARTIQ. Now, the user has to manually find, copy-paste, and run that five-liner. A transparent (i.e. one-time without user interaction built into artiq_master) conversion of `dataset_db.pyon` to the new format would seem a decent improvement of user experience when upgrading. Since `dataset_db.pyon` is already "owned" and under control of the master, we can change it.

jordens:
But maybe there is also no reason to change the format. The `persist` flag is obviously pointless since the presence of a key means exactly that. The HDF flags would only be relevant for presisted datasets that are retrieved by `get()` in an experiment and that way stored into the results. Is this common or could we live with HDF flags being forgotten for persisted datasets?

airwoodix:
> Is this common or could we live with HDF flags being forgotten for persisted datasets?

Wouldn't this imply that we'd get different behavior between saving the dataset in the "creating" experiment's HDF archive and saving from the "reusing" experiment? (beyond the former going to ``/datasets`` and the latter to ``/archive``, which is also confusing (#1272))

jordens:
I agree it's not optimal. The other potential way would be to just add a metadata storage `dataset_meta_db.pyon` where you can persist HDF5 options. That would get around having to change the format and it would be fully backwards and forward compatible.

airwoodix:
The problem goes beyond the persistence format. Upgrading will also break all custom applets since they call ``data[dataset_name][1]`` in ``data_changed()``. In that sense, isn't it reasonable to have an extra manual step upon upgrade? (it would then only be annoying for users running on _beta_)

It could also be an option to revert back to a tuple-based representation, probably as a ``namedtuple`` to keep the expressiveness of the ``dict``-based approach. But this makes discriminating between the old and new formats difficult in ``DatasetDB.__init__``:

```python
file_data = {
    "a": (1, 2),
    "b": ([1, 2, 3], {"compression": "gzip"}),
}

Dataset = namedtuple("Dataset", "persist,value,hdf5_options", defaults=[{}])
datasets = {k: Dataset(True, *v)  # can't decide if one needs to unpack v or not
            for k, v in file_data.items()}
```
As a fix, one could add support for namedtuples in ``sipyco.pyon``.

An extra file surely works, but that's not very scalable. Why should we serialize an internal data structure into two (or more) files? This sounds brittle and doesn't solve the issue with the applets (or other consumers of the internal dataset representation).

jordens:
It will be annoying for those upgrading from ARTIQ-6 to ARTIQ-7, which should ultimately be everyone.
But your arguments make sense. Format looks good.

One thing we can do towards future-proofing this a bit is to not have `**hdf5_option` collect all `kwargs` (in `set_dataset()` and `DatasetDB.set()`) and instead pass them explicitly as `hdf5_options=dict(compression="gzip")`. That would allow distinction of metadata namespaces.

airwoodix:
Yes, that makes a lot of sense. Changed in 8bedf27.

jordens (approved):
LGTM. Thanks.

sbourdeauducq:
@airwoodix There are bugs in the Windows tests (of course). I've fixed some but I cannot figure out this one, so I disabled the corresponding tests (a518963a).
```
ERROR: test_update_hdf5_options (artiq.test.test_dataset_db.TestDatasetDB)
  File "C:\Users\wfvm\Anaconda3\envs\artiq-env\lib\site-packages\artiq\test\test_dataset_db.py", line 31, in setUp
    self.save_ddb_to_disk()
  File "C:\Users\wfvm\Anaconda3\envs\artiq-env\lib\site-packages\artiq\test\test_dataset_db.py", line 37, in save_ddb_to_disk
    self.ddb.save()
  File "C:\Users\wfvm\Anaconda3\envs\artiq-env\lib\site-packages\artiq\master\databases.py", line 73, in save
    pyon.store_file(self.persist_file, data)
  File "C:\Users\wfvm\Anaconda3\envs\artiq-env\lib\site-packages\sipyco\pyon.py", line 222, in store_file
    os.replace(tmpname, filename)
PermissionError: [WinError 5] Access is denied: 'C:\\Users\\wfvm\\AppData\\Local\\Temp\\tmpxr_aj1re' -> 'C:\\Users\\wfvm\\AppData\\Local\\Temp\\tmpozw_bid7'

```

Are you able to fix that?

sbourdeauducq:
Well the Windows insanity continues, I'll just remove test_dataset_db completely.
```
ERROR: test_update_hdf5_options (artiq.test.test_dataset_db.TestDatasetDB)
  File "C:\Users\wfvm\Anaconda3\envs\artiq-env\lib\site-packages\artiq\test\test_dataset_db.py", line 35, in tearDown
    os.unlink(self.persist_file.name)
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\wfvm\\AppData\\Local\\Temp\\tmph0imosnd'

```

airwoodix:
Thanks @sbourdeauducq for the merge and the emergency fix on windows.

> Are you able to fix that?

I'd need to find a windows VM. Will see what can be done.

PR #1550, opened 2020-11-26
doc: fix missing instructions for bypassing Si5324 on Kasli

HarryMakes:


doc/manual/installing.rst
```diff
@@ -287,9 +287,9 @@ For DRTIO systems, the startup kernel should wait until the desired destinations
 
 If you are using DRTIO and the default routing table (for a star topology) is not suitable to your needs, prepare and load a different routing table. See :ref:`Using DRTIO <using-drtio>`.
 
-* Select the RTIO clock source (KC705 only)
+* Select the RTIO clock source (KC705 and Kasli)
 
-The KC705 may use either an external clock signal or its internal clock. The clock is selected at power-up. Use one of these commands: ::
+The KC705 may use either an external clock signal or its internal clock. The clock is selected at power-up. For Kasli, if an external clock is present as the input, setting the RTIO clock source to "external" would bypass the Si5324 synthesiser. To select the source, use one of these commands: ::
```
sbourdeauducq, line 292: "if an external clock is present as the input"? It's bypassing it in all cases regardless of what signal is present at the input.

HarryMakes: I will remove that. In that case, is the following sentence in bold from [doc/manual/core_device.rst](https://github.com/m-labs/artiq/blob/master/doc/manual/core_device.rst#clocking) also redundant?

> On Kasli, when set to e, the rtio_clock setting overrides the built-in (and variant-dependent) Si5324 synthesizer configuration and disables the Si5324. **The user must apply a clock at the RTIO frequency on the Kasli front panel SMA.**

sbourdeauducq: Technically it's redundant, but it clarifies.

sbourdeauducq: You can add that an input clock *must* be present when the external clock is selected.

HarryMakes: Fixed with e29148fa6024be1095c3684e9a32e025ed35de6d.

Issue #1543, opened 2020-11-13
Experiment package imports with git integration


ljstephenson:
# Question

Making the experiment repository a basic python package and trying to import from other files within it doesn't work in conjunction with git integration; I'm wondering what the canonical way of doing this is.

## Description

To elucidate, consider the following example experiment repository:

```
experiments/
    foo/
        bar.py
    baz/
        qux.py
    __init__.py
```

`qux.py`:

```python
from artiq.experiment import *

class Qux(HasEnvironment):
    def build(self):
        pass

    def frobnicate(self):
        print("hello from Qux!")
```

`bar.py`:

```python
from artiq.experiment import *
from experiments.baz.qux import Qux

class Bar(EnvExperiment):
    def build(self):
         self.qux = Qux(self)

    def run(self):
         self.qux.frobnicate()
```

(The reason for having such a structure is that most experiments are composed of repetitive blocks of standard operations, for example state preparation, and these should obviously be reused throughout the more complex composite experiments.)

If this is just on the filesystem, then running `artiq_master -r experiments` works perfectly, `Bar` shows up in the dashboard explorer, and running it prints the expected message.

However, if I follow the git integration steps (initialize a bare repository in another directory, say `experiments.git`, initialise `experiments` as a repository, configure `experiments.git` as a remote and push to it), `artiq_master -g -r experiments.git` fails to load the experiment: the import fails because the module `experiments` does not exist (relative imports also fail). I understand that this is expected due to the nature of python -- **what I don't understand is how one is supposed to make use of the git integration, while also having a sensible codebase that doesn't have unnecessarily duplicated code.**

The obvious workaround is to maintain a library of the common functionality that is installed as a standard python package in the conda/nix environment; but then the "building block" type experiments are separate from the "composite" experiments, and composite experiments can easily be building blocks for yet more complicated experiments, so in fact everything may as well reside in this separate package (as we did in Oxford for the HOA2 experiment). At this point the git integration is useless, because all the functionality is no longer in the experiment repository itself, and so its git hash is irrelevant. One would have to look at the run time of the experiment, cross reference with commit times in the external package, and hope that it was in a clean unmodified state when the experiment was run - the exact problem that the git integration seeks to solve.

1. I'm curious as to other people's experiences: do you use git integration? How are your experiments structured?
1. Would it be difficult to load the repository as a python package? (I'll confess to not having dived into the source yet to investigate)
1. **EDIT:** Have I missed something obvious?

[...]

PR #1571, opened 2020-12-21
Expose `wd` via scheduler

b-bondurant:
# ARTIQ Pull Request

## Description of Changes
Add `get_wd(rid)` method to scheduler and scheduler virtual device. For the use case I have in mind, this would retrieve the repository path when using a filesystem backend, but it could also be useful for retrieving the path of the tmpdir when using the git backend.

Hardly a "feature," but not technically a bugfix either.

Signed-off-by: Brad Bondurant <brad.bondurant@duke.edu>

## Type of Changes

|   | Type |

### All Pull Requests

This does add to the scheduler API, but I wasn't sure if it was significant enough to warrant updating the release notes.

### Code Changes

Current unit tests pass, no new tests added. Manual testing done with filesystem backend, returns path to repository as expected.

### Git Logistics

  ```
  
  ```

### Licensing

### DIFF
--- a/artiq/master/scheduler.py
+++ b/artiq/master/scheduler.py
@@ -485,3 +485,11 @@ class Scheduler:
                     return False
                 return r.priority_key() > run.priority_key()
         raise KeyError("RID not found")
+
+    def get_wd(self, rid):
+        """Returns the ``wd`` attribute of the run with the specified RID."""
+        for pipeline in self._pipelines.values():
+            if rid in pipeline.pool.runs:
+                run = pipeline.pool.runs[rid]
+                return run.wd
+        raise KeyError("RID not found")
--- a/artiq/frontend/artiq_master.py
+++ b/artiq/frontend/artiq_master.py
@@ -123,6 +123,7 @@ def main():
         "scheduler_request_termination": scheduler.request_termination,
         "scheduler_get_status": scheduler.get_status,
         "scheduler_check_pause": scheduler.check_pause,
+        "scheduler_get_wd": scheduler.get_wd,
         "ccb_issue": ccb_issue,
     })
     experiment_db.scan_repository_async()
--- a/artiq/master/worker_impl.py
+++ b/artiq/master/worker_impl.py
@@ -122,6 +122,13 @@ class Scheduler:
         make_parent_action("scheduler_request_termination"))
     get_status = staticmethod(make_parent_action("scheduler_get_status"))
 
+    _get_wd = staticmethod(make_parent_action("scheduler_get_wd"))
+    @host_only
+    def get_wd(self, rid=None):
+        if rid is None:
+            rid = self.rid
+        return self._get_wd(rid)
+
 
 class CCB:
     issue = staticmethod(make_parent_action("ccb_issue"))


artiq/master/scheduler.py
```diff
@@ -485,3 +485,11 @@ def check_pause(self, rid):
                     return False
                 return r.priority_key() > run.priority_key()
         raise KeyError("RID not found")
+
+    def get_wd(self, rid):
+        """Returns the ``wd`` attribute of the run with the specified RID."""
```
airwoodix, line 490: ``wd`` should be spelled out if this is part of the user-facing API. Attributes of `artiq.master.scheduler.Run` are not documented.

artiq/master/scheduler.py
```diff
@@ -485,3 +485,11 @@ def check_pause(self, rid):
...
+        for pipeline in self._pipelines.values():
+            if rid in pipeline.pool.runs:
+                run = pipeline.pool.runs[rid]
+                return run.wd
+        raise KeyError("RID not found")
```
airwoodix, line 495: Matter of taste: why not?
```python
try:
    return pipeline.pool.runs[rid].wd
except KeyError:
    raise KeyError("RID not found") [from None]
```

airwoodix (commented):
I unfortunately cannot test this right now: what's the difference between `wd` and `dirname(expid["file"])`?
Given #1543, adding the canonical usage to the docs would be a great improvement IMHO.

artiq/master/scheduler.py
b-bondurant: I was going for consistency with `check_pause()`, but yeah there's no reason why not

b-bondurant:
>  what's the difference between wd and dirname(expid["file"])

`expid['file']` points all the way to the file containing the experiment in question, but has no knowledge of which part of that path refers to the repository root.

> Given #1543, adding the canonical usage to the docs would be a great improvement IMHO.

I hadn't seen that issue, but I agree that this might be helpful there, and it's actually pretty similar to the use case I have in mind, although I'm not using the git backend. I'm writing a sort of base class that needs to import other experiments (specified by the user when they implement the class) based on a `file` and `class_name`, and I decided to use `artiq.tools.file_import` and `.get_experiment` - but unfortunately I need absolute paths since the repository root path isn't directly exposed anywhere AFAICT. For the use case in that issue, it seems like they could use the same `artiq.tools` if they knew the directory of the git checkout (which I think is what `wd` would be in that case).

b-bondurant:
An alternate implementation of this functionality would be to simply add a `wd` field to the run's notifier - that might actually be more favorable as it's less intrusive and doesn't technically change the API.

artiq/master/scheduler.py
b-bondurant: Actually, if you put this whole try/except inside the loop then I don't think it would work since this would raise a key error for every pipeline that doesn't contain that RID, even if there's one that does.

sbourdeauducq:
After https://github.com/m-labs/artiq/issues/1543#issuecomment-731927807 we could just do ``os.getcwd()`` in experiments I think, without requiring this relatively complex change?

b-bondurant:
Yes! If the working directory was set to the repository rather than the results directory, I think that would take care of things. Is that a change you plan on introducing into `release-5` as well?

sbourdeauducq:
>  Is that a change you plan on introducing into `release-5` as well?

No, and this PR would not be eligible for release-5 either.

b-bondurant:
> An alternate implementation of this functionality would be to simply add a wd field to the run's notifier - that might actually be more favorable as it's less intrusive and doesn't technically change the API.

Would you consider it for release-5 if I went with this approach instead? Just to make do until we're ready to update to the next release.

artiq/master/scheduler.py
airwoodix: Good point indeed, sorry. Missed one level of indentation.

b-bondurant: Probably a moot point anyway as it looks like this PR will be rejected for release-5, and unnecessary for master/future release-6

sbourdeauducq:
No, we don't introduce new features on ``release-5`` either.
If you're using Nix, you can use ``overrideAttrs`` and ``patches`` to cleanly apply the change (which would be https://github.com/m-labs/artiq/issues/1543#issuecomment-731927807 to end up in ``master``) to your local ARTIQ-5 installation.

b-bondurant:
Understood, thanks!

lriesebos:
@sbourdeauducq I see you have reopened this PR a while ago. Is there a way @b-bondurant can refactor this change to get this functionality into `master` at some point?

sbourdeauducq:
Do you still need it if we do #1543?

b-bondurant:
> Do you still need it if we do #1543?

I think we would, but I'm not familiar enough with Python's import machinery to be sure.

My specific use case is a bit different from those that have been discussed so far in that issue. I'm using `artiq.tools.file_import` for runtime imports in a class I wrote for DAX. It's a bit complicated to explain, so I'll just give a brief summary and then link to the relevant code.

Essentially, this class (`CalibrationJob`) receives the paths (absolute or relative to the repository directory) and class names of two experiments via class variables from a user-implemented subclass, then wraps them into a single "meta" experiment and inserts that experiment into the global namespace of the file containing the `CalibrationJob` subclass so that it may be picked up by ARTIQ as part of the repository.

class def: https://gitlab.com/duke-artiq/dax/-/blob/master/dax/base/scheduler.py#L693
runtime imports: https://gitlab.com/duke-artiq/dax/-/blob/master/dax/base/scheduler.py#L827-830
current hack to resolve repository-relative paths: https://gitlab.com/duke-artiq/dax/-/blob/master/dax/base/scheduler.py#L1068-1077
decorator to create meta experiment at import time and insert into user file: https://gitlab.com/duke-artiq/dax/-/blob/master/dax/base/scheduler.py#L1112

sbourdeauducq:
@b-bondurant Since we didn't reach consensus on an alternative approach, let's merge this. Can you address the review items?

lriesebos:
Maybe it is possible to rename the function to `get_repository()` to leave room for #1747 ?

b-bondurant:
> Maybe it is possible to rename the function to `get_repository()` to leave room for #1747 ?

@sbourdeauducq what do you think about something like that? It also might be clearer to users since (as @airwoodix pointed out), `wd` isn't documented, and AFAIK everywhere else in the docs it's referred to as "repository". Also, given recent PR #1767, it's possible users would expect `get_wd` to point to the subdirectory, while calling it `get_repository` would (IMO) remove any potential confusion.

sbourdeauducq:
What happens with experiments submitted "outside the repository"? This corner case is also what is plaguing the other proposals.

b-bondurant:
It returns `None`, which I've explained in my updated docstring (unpushed since I'm planning to amend and force push to make the switch to `get_repository_dir` if we decide that's what we want to call it)

b-bondurant:
And I actually didn't need to do any special handling for that case since I'm just returning `run.wd`, which gets set to `None` by default if there's no `repo_rev` field in the expid, i.e. for out-of-repo experiments.

b-bondurant:
For git-based repositories (i.e. using `artiq_master -g ...`), it returns the temp dir containing the checkout that the master is using. I think that would be the desired behavior, since getting the path of a bare git repo would be useless, but I'll make sure to note that in the docstring.

PR #1618, opened 2021-02-27
added functions to get freq/phase/amp/att for ad99xx and urukul

### LINKED ISSUE #1616: get functions for frequency, amplitude, phase parameters of AD9910
# ARTIQ Feature Request

There is no an explicit function to get the frequency, amplitude, or phase from a AD9910 device. I understand that the values can be get using `read32` function with the corresponding registry addresses, but I think having those explicit get functions for the AD9910 device helps reducing some of the code duplication that I imagine many labs using urukul have. Please let me know if this is already addressed somewhere or this idea does not make sense.

Add `get_ftw`, `get_asf`, `get_pow` functions in `artiq.coredevice.ad9910.AD9910` class, and add `get_frequency`. `get_amplitude`, and `get_phase` as corresponding counterparts in natural units.

lriesebos:
Signed-off-by: Leon Riesebos <leon.riesebos@duke.edu>

# ARTIQ Pull Request

## Description of Changes

Added `get()` and `get_mu()`, `get_att()`, and `get_att_mu()` functions for the AD9910 and AD9912 coredevice drivers.

### Related Issue

Closes #1616 

## Type of Changes

|   | Type |

### All Pull Requests

- [x] Close/update issues.

### Code Changes

### Documentation Changes

### Git Logistics

  ```
  
  ```

### Licensing

### DIFF
--- a/artiq/coredevice/ad9912.py
+++ b/artiq/coredevice/ad9912.py
@@ -1,6 +1,6 @@
 from numpy import int32, int64
 
-from artiq.language.types import TInt32, TInt64
+from artiq.language.types import TInt32, TInt64, TFloat, TTuple
 from artiq.language.core import kernel, delay, portable
 from artiq.language.units import ms, us, ns
 from artiq.coredevice.ad9912_reg import *
@@ -39,12 +39,12 @@ class AD9912:
             self.sw = dmgr.get(sw_device)
             self.kernel_invariants.add("sw")
         self.pll_n = pll_n
-        sysclk = self.cpld.refclk/[1, 1, 2, 4][self.cpld.clk_div]*pll_n
+        sysclk = self.cpld.refclk / [1, 1, 2, 4][self.cpld.clk_div] * pll_n
         assert sysclk <= 1e9
-        self.ftw_per_hz = 1/sysclk*(int64(1) << 48)
+        self.ftw_per_hz = 1 / sysclk * (int64(1) << 48)
 
     @kernel
-    def write(self, addr, data, length):
+    def write(self, addr: TInt32, data: TInt32, length: TInt32):
         """Variable length write to a register.
         Up to 4 bytes.
 
@@ -55,14 +55,14 @@ class AD9912:
         assert length > 0
         assert length <= 4
         self.bus.set_config_mu(urukul.SPI_CONFIG, 16,
-                urukul.SPIT_DDS_WR, self.chip_select)
+                               urukul.SPIT_DDS_WR, self.chip_select)
         self.bus.write((addr | ((length - 1) << 13)) << 16)
-        self.bus.set_config_mu(urukul.SPI_CONFIG | spi.SPI_END, length*8,
-                urukul.SPIT_DDS_WR, self.chip_select)
-        self.bus.write(data << (32 - length*8))
+        self.bus.set_config_mu(urukul.SPI_CONFIG | spi.SPI_END, length * 8,
+                               urukul.SPIT_DDS_WR, self.chip_select)
+        self.bus.write(data << (32 - length * 8))
 
     @kernel
-    def read(self, addr, length):
+    def read(self, addr: TInt32, length: TInt32) -> TInt32:
         """Variable length read from a register.
         Up to 4 bytes.
 
@@ -73,15 +73,15 @@ class AD9912:
         assert length > 0
         assert length <= 4
         self.bus.set_config_mu(urukul.SPI_CONFIG, 16,
-                urukul.SPIT_DDS_WR, self.chip_select)
+                               urukul.SPIT_DDS_WR, self.chip_select)
         self.bus.write((addr | ((length - 1) << 13) | 0x8000) << 16)
         self.bus.set_config_mu(urukul.SPI_CONFIG | spi.SPI_END
-                | spi.SPI_INPUT, length*8,
-                urukul.SPIT_DDS_RD, self.chip_select)
+                               | spi.SPI_INPUT, length * 8,
+                               urukul.SPIT_DDS_RD, self.chip_select)
         self.bus.write(0)
         data = self.bus.read()
         if length < 4:
-            data &= (1 << (length*8)) - 1
+            data &= (1 << (length * 8)) - 1
         return data
 
     @kernel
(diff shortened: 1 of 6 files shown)


lriesebos:
@fanmingyu212 , maybe you can test these changes?

fanmingyu212 (commented):
See https://gist.github.com/fanmingyu212/8b5de74d45334c0338bc30223b221a26 for the test code. In short, all functions in ad9910.py and urukul.py are tested. ad9912.py are not tested as I don't have the hardware. The gateware version used for testing is 7.7581.e577542f.beta.

`get()`, `get_mu()`, `get_att_mu()`, and `get_att()` in ad9910.py are tested.

`set_att()`, `get_att_mu_()`, and `get_att()` in urukul.py are also tested.

`mu_to_att()` and `att_to_mu()` in urukul.py are confirmed to be inverse functions of each other, but I don't have a great way to explicitly test the result here.

airwoodix:
@lriesebos Thanks a lot for the patch. I think the following points should be addressed:

- type hints: I guess that they're not strictly needed, or are they? I see their relevance, especially for `get()` and think they make a lot of sense in the coredevice area. I'd however add them everywhere or not at all;
- It'd be great to also add counterparts to `set_frequency()` & co for reading the FTW, POW and ASF registers (@0x7-0x9);
- would also be nice to add @fanmingyu212 's tests in `test/coredevice/test_ad9910.py`;
- what's the rationale behind `cpld.get_att_mu_()`? Are there situations where you can't afford the overhead of cycling once through the shift register? The SR management is already confusing for newcomers (see forum), does this help or add more confusion? With this patch `get_att_mu()` returns the hardware value but `get_att()` returns the cached value.

lriesebos:
@fanmingyu212 thanks for testing! I have merged parts of your testing code into the ARTIQ tests in a way that fits the existing format, see eda2a04 . Though I am not able to run these test cases here, so I guess CI will tell us if there are bugs in the test code.

For the other points:
- @airwoodix I think type hints are not required, but they do give the compiler direct instructions on what types to infer. I found it especially useful for functions that return a `TInt64` as it will not be implicitly casted to a `TInt32` when assigned to a variable. For parameters, not typing them might allow the user to put in different types (e.g. a `TInt32` or a `TInt64). I am both fine with typing everything or with minimal typing. @sbourdeauducq , what do you recommend/prefer?
- I added get functions for the io update registers, see fe7bc94 .
- For `cpld.get_att_mu_()`, my idea was to not add any new complexity to the existing confusion around the shift register. New users already have to learn to call `get_att_mu()` once before doing other attenuation operations and that remains the same. Though if others see this differently, I am happy to change it. I see the following options:
   1. Keep the existing structure, users need to call `get_att_mu()` before they can use `get_att_mu_()`
   2. `get_att_mu_()` will always call `get_att_mu()` first, so now users can call one of both to cache the register value
   3. Add a flag `get_att_mu_(cached=True)`, now users can choose between 1 and 2, for example with 1 as default

sbourdeauducq:
>     * For parameters, not typing them might allow the user to put in different types (e.g. a `TInt32` or a `TInt64`). I am both fine with typing everything or with minimal typing. @sbourdeauducq , what do you recommend/prefer?

Put the type annotations on kernels, NAC3 (the new compiler) will require them.

lriesebos:
@airwoodix , I implemented your suggestion, `get_att_mu_()` now reads the shift register before returning the value. I also refactored the three files of interest and added typing for all kernel functions.

@fanmingyu212 could you be so kind to test the ad9910 and urukul driver in its completeness after my refactoring and typing?

artiq/coredevice/ad9912.py
```diff
@@ -123,7 +123,7 @@ def set_att_mu(self, att):
         self.cpld.set_att_mu(self.chip_select - 4, att)
 
     @kernel
-    def set_att(self, att):
+    def set_att(self, att: TInt32):
```
fanmingyu212, line 126: Should the type here be TFloat instead of TInt32?

artiq/coredevice/urukul.py
```diff
@@ -343,13 +365,41 @@ def get_att_mu(self):
...
+        The result is stored and will be used in future calls of
+        :meth:`set_att_mu` and :meth:`set_att`.
+
+        .. seealso:: :meth:`get_att_mu`
+
+        :param channel: Attenuator channel (0-3).
+        :return: 8-bit digital attenuation setting:
+            255 minimum attenuation, 0 maximum attenuation (31.5 dB)
+        """
+        return int32((self.get_att_mu >> (channel * 8)) & 0xff)
```
fanmingyu212, line 386: Should it be `self.get_att_mu()` instead of `self.get_att_mu` here?

fanmingyu212 (changes requested):
@lriesebos I did testing with AD9910 and found 2 possible bugs as commented. After changing those two lines, the [test code](https://gist.github.com/fanmingyu212/8b5de74d45334c0338bc30223b221a26) (only slightly modified in the `test_get_att` function with 1 ms delay between `ad9910.get_att` and `ad9910.get_att_mu` to prevent a RTIO underflow) runs with no problem.

lriesebos:
@fanmingyu212 thanks again for testing!

@sbourdeauducq do we need to do anything else to get this merged?

.gitignore
```diff
@@ -9,6 +9,7 @@ __pycache__/
 *.pcap
 *.prof
 .ipynb_checkpoints
+/.idea/
```
airwoodix, line 12: This should go to your ``.git/info/exclude`` or your global gitignore (``config core.excludesFile``).

artiq/coredevice/urukul.py
```diff
@@ -343,13 +365,41 @@ def get_att_mu(self):
...
                                SPIT_ATT_WR, CS_ATT)
-        delay(10*us)
+        delay(10 * us)
         self.att_reg = self.bus.read()
         self.bus.write(self.att_reg)  # shift in current value again and latch
         return self.att_reg
 
     @kernel
-    def set_sync_div(self, div):
+    def get_att_mu_(self, channel: TInt32) -> TInt32:
```
airwoodix, line 374: This should have a more explicit name. Maybe ``get_channel_att_mu``?

artiq/coredevice/urukul.py
```diff
@@ -343,13 +365,41 @@ def get_att_mu(self):
...
+        .. seealso:: :meth:`get_att_mu`
+
+        :param channel: Attenuator channel (0-3).
+        :return: 8-bit digital attenuation setting:
+            255 minimum attenuation, 0 maximum attenuation (31.5 dB)
+        """
+        return int32((self.get_att_mu() >> (channel * 8)) & 0xff)
+
+    @kernel
+    def get_att(self, channel: TInt32) -> TFloat:
```
airwoodix, line 389: For consistency, would it make sense to make this return all the attenuations in dB and add ``get_channel_att()``? The symmetry between machine and human units wouldn't be perfect (I guess that return the 32 bits of the SR in ``get_att_mu()`` is more appropriate than a tuple) but IMHO better.

airwoodix:
@lriesebos Thanks for the feedback and the changes. The additions are really great! I just think that the attenuator getter interface should be less cryptic. What do you think about changing the function names as suggested?

artiq/coredevice/urukul.py
lriesebos: I renamed the function. `get_att()` can not be implemented without storing extra information because a locally generated list can not be returned from a function (i.e. object lifetime). To implement this, there needs to be a second `att_reg` that stores the float values. That sounds a bit overkill to me and I think we can leave that function out. Do you agree?

airwoodix: Thanks! I agree: I guess that most code will use ``get_channel_att`` anyway.

fanmingyu212:
The latest changes also run with no problem on our setup (ad9910).

sbourdeauducq:
Can you clean up the commit list a bit? For example, make it only two commits, the first one with the style changes in the existing code, and the second one that adds the functions.

lriesebos:
@sbourdeauducq commits are cleaned up as suggested. if there is anything else you would like me to do, let me know.

sbourdeauducq:
Thanks. I assume this is tested.

PR #1628, opened 2021-03-09
fastino: ensure `xxx_to_mu()` methods return int32 on the host

pathfinder49:
# ARTIQ Pull Request
## Description of Changes

Currently running `voltage_to_mu()` or `voltage_group_to_mu()` on the host will
convert all machine unit values to int64. This leads to issues when machine units
are returned from RPCs.

This PR allows `xxx_to_mu()` methods for Fastino to return the required int32 type on the core and host device.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

### All Pull Requests

- [ ] Update [RELEASE_NOTES.rst](../RELEASE_NOTES.rst) Minor bug fix. I'm also not aware of other Fastino users at this time.

### Code Changes

- [x] Test your changes or have someone test them. Tested by setting and verifying fastino voltages with `xxx_to_mu` calculated on the core and host device.

### Licensing

### DIFF
--- a/artiq/coredevice/fastino.py
+++ b/artiq/coredevice/fastino.py
@@ -1,6 +1,7 @@
 """RTIO driver for the Fastino 32channel, 16 bit, 2.5 MS/s per channel,
 streaming DAC.
 """
+from numpy import int32
 
 from artiq.language.core import kernel, portable, delay
 from artiq.coredevice.rtio import (rtio_output, rtio_output_wide,
@@ -112,7 +113,7 @@ class Fastino:
         :param voltage: Voltage in SI Volts.
         :return: DAC data word in machine units, 16 bit integer.
         """
-        data = int(round((0x8000/10.)*voltage)) + 0x8000
+        data = int32(round((0x8000/10.)*voltage)) + int32(0x8000)
         if data < 0 or data > 0xffff:
             raise ValueError("DAC voltage out of bounds")
         return data
@@ -129,7 +130,7 @@ class Fastino:
             v = self.voltage_to_mu(voltage[i])
             if i & 1:
                 v = data[i // 2] | (v << 16)
-            data[i // 2] = v
+            data[i // 2] = int32(v)
 
     @kernel
     def set_dac(self, dac, voltage):


pathfinder49:
@sbourdeauducq Are there any issues/ questions regarding this PR? Otherwise, it may be best to merge before Fastino has wider adoption.

sbourdeauducq:
> Currently running voltage_to_mu() or voltage_group_to_mu() on the host will convert all machine unit values to int64.

Isn't it returning ``int``, not ``int64``?

Isn't there the same situation with Zotino?
https://github.com/m-labs/artiq/blob/master/artiq/coredevice/ad53xx.py#L85

pathfinder49:
> Isn't it returning `int`, not `int64`?

Yes, what I tried to express was that the python int is an int64 (not numpy.int64). This is an issue for Fastino as, unlike Zotino, the `voltage_group_to_mu()` method packs 2 16 bit dac words into a single (64 bit) integer. If thhe machine unit values are calulated on the master 1 of 2 things will happen.
1.  If transfered to the kernel via an attribute, I would expect compilation errors. The 32 significant bits in the 64 bit host integer become 64 bit integers on the core, violating type expectations for `set_group_mu()`
2. If calculated in an RPC there is data corruption. For the code to compile, an RPC returning Fastino machine units would be annotated `-> TList(TInt32)`. This results in the RPC type casting the 64 bit integers to **signed** 32 bit integers. This results in the most significant bit of the the 32 data-bits in the int64 being discarded as it exeeds the range of a **signed** 32 bit integer. This does not result in erros/warings. the only indication this is happening is a message logged at debug level. (I've been meaning to open a seperate issue on the lack of log warning/error)

sbourdeauducq:
>     2\. If calculated in an RPC, there is data corruption. For the code to compile, an RPC returning Fastino machine units would be annotated `-> TList(TInt32)`. This results in the RPC type casting the 64 bit integers to **signed** 32 bit integers. Consequently the most significant bit of the the 32 data-bits in the int64 being discarded, as it exeeds the range of a **signed** 32 bit integer.

Integer overflow is not data corruption, and in this case I believe it would produce the correct result. The MSB would not be discarded, but simply potentially interpreted as sign bit e.g. if you multiply or print the number, but in this case it is equivalent.

sbourdeauducq:
>     1. the 64 bit host-integer

Python has bigints. There is no such thing as a "64-bit host integer" with the ``int`` type.

artiq/coredevice/fastino.py
```diff
@@ -112,7 +113,7 @@ def voltage_to_mu(self, voltage):
         :param voltage: Voltage in SI Volts.
         :return: DAC data word in machine units, 16 bit integer.
         """
-        data = int(round((0x8000/10.)*voltage)) + 0x8000
+        data = int32(round((0x8000/10.)*voltage)) + int32(0x8000)
```
sbourdeauducq, line 116: A single ``int32(... + 0x8000)`` would be clearer.

sbourdeauducq:
>     1. If transfered to the kernel via an attribute, I would expect compilation errors.

Nothing to do with "64-bit host ints" but indeed if the value is >= 0x80000000 the compiler infers ``int64`` instead of ``int32``, as can be demonstrated by this example:

```python


class IntTypes(EnvExperiment):
    def build(self):
        self.setattr_device("core")
        self.test1 = 0x40000000
        self.test2 = 0x80000000

    def print_type(self, x):
        print(type(x))

    @kernel
    def run(self):
        self.print_type(self.test1)
        self.print_type(self.test2)
```

So point (1) is valid.

pathfinder49:
> Integer overflow is not data corruption, and in this case I believe it would produce the correct result.

Unfortunatly the bits get discarded. In the example below `int32_from_rpc` exhibits the behaviour need to not loose data bits. 

```Python3
import numpy as np


class RPCIntCasting(EnvExperiment):
    def build(self):
        self.setattr_device("core")
        self.uint32 = 0x80000000

    def int_from_rpc(self, x) -> TList(TInt32):
        return [int(x)]

    def int32_from_rpc(self, x) -> TList(TInt32):
        return [np.uint32(x).view(np.int32)]

    @kernel
    def run(self):
        print(self.int_from_rpc(self.uint32))
        print(self.int32_from_rpc(self.uint32))
```

The results of the two rpc implementations differ.
```
print:[1515870810]
print:[-2147483648]
```

sbourdeauducq:
> `print:[1515870810]`

This sounds like a bug with the compiler or runtime, and it's not about bits that get discarded.
1515870810 is 0x5a5a5a5a, which does not contain bits from your original number but is part of the header used in the internal core device/host communication protocol.

pathfinder49:
Thanks :) I'm just starting on an issue

artiq/coredevice/fastino.py
pathfinder49: I agree that would be nicer. Unfortunatly, it does not compile on the coredevice.

pathfinder49:
@sbourdeauducq thank you for your feedback and clearing up my superficial understanding of the underlying problem. I believe this workaround still adds value to the Fastino driver. Can this be merged?

artiq/coredevice/fastino.py
jordens: `int32(int(round(..)) + 0x8000)`

artiq/coredevice/fastino.py
```diff
@@ -128,7 +129,7 @@ def voltage_group_to_mu(self, voltage, data):
         for i in range(len(voltage)):
             v = self.voltage_to_mu(voltage[i])
             if i & 1:
-                v = data[i // 2] | (v << 16)
+                v = data[i // 2] | (v << int32(16))
             data[i // 2] = v
```
jordens, line 133:
```suggestion
                v = data[i // 2] | (v << 16)
            data[i // 2] = int32(v)
```

Would make the intention significantly clearer to me.

jordens (changes requested):

artiq/coredevice/fastino.py
pathfinder49: Ack

artiq/coredevice/fastino.py
pathfinder49: Unfortunately, that doesn't compile either.

jordens: Weird.

jordens:
LGTM. Thanks.

PR #1657, opened 2021-04-14
Phaser

### LINKED ISSUE #1630: Phaser NCO does not generate correct frequency
# Bug Report

## One-Line Summary
Numerically Controlled Oscillator (NCO) is not working.

## Issue Details
The NCO upconversion set_nco_frequency() does not change the generated frequency. Instead the frequency remains as it were with set_nco_frequency(0). No error is raised.
### Steps to Reproduce

````
class PhaserHarmonics(EnvExperiment):
     def build(self):
        self.setattr_device("core")
        self.setattr_device("phaser_1")
        self.setattr_device("scheduler")
    @kernel
    def run(self):
        f = self.phaser_1
        self.core.reset()
        self.core.break_realtime()
        f.init()
        self.core.break_realtime()
        ch = f.channel[0]
        ch.set_att(0*dB)
        ch.set_duc_frequency(120*MHz)
        ch.set_nco_frequency(200*MHz)
        ch.set_nco_phase(0.)
        ch.set_duc_phase(0.)
        ch.set_duc_cfg(select=0, clr=0)
        delay(.1*ms)
        ch.oscillator[0].set_frequency(0)
        delay(0.1*ms)
        ch.oscillator[0].set_amplitude_phase(1.)
        delay(0.1*ms)
        f.duc_stb()
        pausecheck = 0
        while True:
            delay(50*ms)

            pausecheck += 1
            if pausecheck % 40 == 0:
                if self.scheduler.check_pause():
                    break
````

### Expected Behavior
tone at 320 MHz

tone at 120 MHz

Using Artiq master & Phaser master gateware / firmware.
ARTIQ version: ARTIQ v6.0.dev+66.ga8c5ad3b

### LINKED ISSUE #1638: Phaser: Rename `ndiv` -> `nint`.
# Bug Report

## One-Line Summary

Phaser-Upconverter LO configuration seems to be missing options.

## Issue Details
I'm attempting to create a `device_db` for a phaser-upconverter. With the current coredevice driver, the registers inside the TRF372017 must be configured manually. The documentations references the TRF372017 data-sheet without further explanation. 
https://github.com/m-labs/artiq/blob/d04bcd8754f11780d02d179285712980701d6518/artiq/coredevice/trf372017.py#L1-L6
To avoid misunderstandings, I'll summarise my understanding of the required configuration options. Currently, all bar one of these options seem to be exposed.

From the data-sheet it seems like the following values must be configured to achieve a desired LO frequency:
* LO_DIV_SEL
* PLL_DIV_SEL
* NINT (**this name does not appear in the driver**, @jordens am I missing something obvious?)
* PRSC_SEL
* RDIV 

For "fractional mode" the following additional options are used
* NFRAC and
image

If an external VCO/LO is provided the following options are needed:
* EN_EXTVCO
* NEG_VCO

### Expected Behavior
All the options above should be exposed and the names should match with the data-sheet. (Naming already matches for the exposed options)

### LINKED ISSUE #1643: Incorrect VCO calibration procedure for Phaser upconverter
# Bug Report

## One-Line Summary

The automatic on-chip VCO selection is performed before the requisite configuration data has been written.

## Issue Details
The trf performs automatic calibration of it VCO when the EN_CAL bit is set. The core device driver currently writes all settings in order of register address order. This results in the EN_CAL bit being set before all registers defining the VCO frequency have been set. For example both EN_FRAC and NFRAC are set after EN_CAL. The data-sheet explicitly states that EN_CAL must be set after the frequencies have been configured.

image
If the register values have not been set the VCO calibration will be performed for the wrong frequency.

Currently configuration of the trf is performed in  `Phaser.init()`, code below. 
https://github.com/m-labs/artiq/blob/d04bcd8754f11780d02d179285712980701d6518/artiq/coredevice/phaser.py#L319-L321
`trf_mmap` is a register-address-ordered list of register-values. These values are written in the list order.
### Expected Behavior

EN_CAL should be written only once the relevant values have been configured. From my current understanding of the data-sheet setting EN_CAL last should be acceptable.

### LINKED ISSUE #1648: Phaser Upconverter: Improve default configuration.
# Bug Report

## One-Line Summary
The Phaser upconverter phase-frequency detector is not configured at the frequency targeted by the hardware design.

## Issue Details
Phaser was designed [targeting a phase-frequency detector clock of 62.5 MHz](https://github.com/sinara-hw/Phaser/issues/49#issuecomment-555163525). This frequency should be kept high to keep phase noise low. However, the current default is a 6 MHz phase-frequency detector clock.

### Expected Behavior

Phaser defaults to a 125 MHz clock from Kasli, instead of an external reference. The phase-frequency detector should default to 62.5 MHz.

### Suggested Fix
Default to `rdiv = 2`.
To keep the calibration clock within it's 1 MHz spec, we will also need `cal_clk_sel = 0b1110 ` (DIV64)
https://github.com/m-labs/artiq/blob/8002fcf8bbeb3425f19d66244130c6f1b1f9b7a3/artiq/coredevice/trf372017.py#L7-L12

### LINKED ISSUE #1650: Phaser:  DAC `sif_sync` not cleared in `init()`
# Bug Report

## One-Line Summary

During `init()` the sif_sync bit of the Phaser DAC is set and not cleared. This blocks further sync until manually cleared.

## Issue Details
The `Phaser.init()` method sets the `sif_sync` bit. This is required to apply the written NCO values. Unless the bit is cleared, any changes to the NCO frequency can not be applied.
image

Currently, this bit is set when applying the DAC memory map (as defined in `artiq.coredevice.dac34h84`). This is not a suitable interface for syncing the NCO from user code.

### Expected Behavior
The bit should be cleared during `Phaser.init()`. Once the bit is cleared, the NCO frequency can be written using the `set_nco_frequency` method. 

A reasonable approach would be to provide a `sync_nco` method that applies written NCO changes. This would also avoid the need for manually accessing the DAC registers to update the NCO frequency. Currently this has to be done as shown below:
```Python
        data = self.phaser.dac_mmap[-9]
        data0 = data & ~int32(0b10)
        data1 = data
        self.phaser.dac_write(data0 >> 16, data0)
        self.phaser.dac_write(data1 >> 16, data1)
```

The bit is not cleared during `Phaser.init()`. This blocks further syncing of NCO frequency changes, until the bit has been manually cleared.

### LINKED ISSUE #1651: Phaser: Oscillator 0 amplitude not cleared after initialization
# Bug Report

## One-Line Summary

`Phaser.init()` sets the `phaser.channel[i].oscillator[0]` to full scale, and does not disable it again.

## Issue Details
During initialisation, phaser performs tests of each channel using the (Kasli) 25 MS/s NCO. For this `phaser.channel[i].oscillator[0]` is written to full scale. Thhi oscillator is not disabled/set to 0 amplitude again during initialisation. At the end of initialisation dac_tx is enabled. As the oscilator has not been disabled, ~~this~~ (Edit: configuring the DUC/Kasli-NCO/DAC-NCO frequency) applies continuous full scale RF the the Phaser outputs. (Confirmed on hardware)
https://github.com/m-labs/artiq/blob/fc59791583905f9e35a4ef0bc997c21ebc649a64/artiq/coredevice/phaser.py#L288-L295

I see two points of concern:
1. The user does not expect phaser to be continuously producing RF after initialisation. 
2. Phaser may plausibly drive loads that are damaged by continuous full power (or inappropriate output frequency). RF output should only be enabled when needed/requested by the user.

pathfinder49:
# ARTIQ Pull Request

## Description of Changes
Phaser quality of life improvements

1. The current Phaser core device driver requires users perform register level manipulations of the DAC to use/change the frequency of the coarse mixer and NCO. This PR adds methods that allow users to reconfigure the NCO and coarse mixer frequency without needing to refer to the DAC-datasheet.
2. Document the need to enable the DAC mixer in order to use the NCO
2. An offset error for the DAC coarse mixer bits is fixed.
3. `Phaser.channel.oscillator[0]` is no-longer special cased after `init()`. One of the main features of Phaser is the ability to pulse shape. In this context, enabling RF output without specifying an amplitude leads to breakdown of abstraction. 
4. The TRF Phaser-Frequency-Detector (PFD) clock now defaults to the hardware design target of 62.5 MHz. This configuration should offer the lowest phase noise and will be a reasonable starting point for users to perform adjustments. Currently the PFD is set to an arbitrarily chosen frequency.
5. Refactor the TRF calibration procedure to follow that described in the data-sheet.
6. Refactor the `NDIV` -> `NINT` to match the documentation naming. 
7. Work around a bug that leads to malformed output when initialising phaser with `mixer_ena=1` & `nco_ena=0` after a power cycle.

### Related Issue
closes #1651
closes #1650
closes #1648
closes #1643
closes #1638
closes #1630

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |
| ✓  | :scroll: Docs |

### All Pull Requests

- [x] Close/update issues.

### Code Changes

- [x] Test your changes or have someone test them. Changes were tested by measuring the Phaser output on a 4 GHz scope.

### Documentation Changes

### Git Logistics

  ```
  
  ```

### Licensing

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -9,8 +9,17 @@ ARTIQ-7
 Highlights:
 * WRPLL
 * ``get()``, ``get_mu()``, ``get_att()``, and ``get_att_mu()`` functions added for AD9910 and AD9912
+* Phaser:
+   - Improved documentation
+   - Expose the DAC coarse mixer and sif_sync
+   - Exposes upconverter calibration and enabling/disabling of upconverter LO & RF outputs.
 
 Breaking changes:
+* Updated Phaser-Upconverter default frequency 2.875 GHz. The new default uses the target PFD
+  frequency of the hardware design.
+* `Phaser.init()` now disables all Kasli-oscillators. This avoids full power RF output being
+  generated for some configurations.
+* Phaser: fixed coarse mixer frequency configuration
 
 
 ARTIQ-6
--- a/artiq/coredevice/dac34h84.py
+++ b/artiq/coredevice/dac34h84.py
@@ -201,7 +201,7 @@ class DAC34H84:
         mmap.append(
             (0x0d << 16) |
             (self.cmix_fs8 << 15) | (self.cmix_fs4 << 14) |
-            (self.cmix_fs2 << 12) | (self.cmix_nfs4 << 11) |
+            (self.cmix_fs2 << 13) | (self.cmix_nfs4 << 12) |
             (self.qmc_gainb << 0))
         mmap.append((0x0e << 16) | (self.qmc_gainc << 0))
         mmap.append(
--- a/artiq/coredevice/dac34h84.py
+++ b/artiq/coredevice/dac34h84.py
@@ -178,7 +178,8 @@ class DAC34H84:
             (self.collisiongone_ena << 12) | (self.sif4_ena << 7) |
             (self.mixer_ena << 6) | (self.mixer_gain << 5) |
             (self.nco_ena << 4) | (self.revbus << 3) | (self.twos << 1))
-        mmap.append((0x03 << 16) | (self.coarse_dac << 12) | (self.sif_txenable << 0))
+        mmap.append((0x03 << 16) | (self.coarse_dac << 12) |
+                    (self.sif_txenable << 0))
         mmap.append(
             (0x07 << 16) |
             (self.mask_alarm_from_zerochk << 15) | (1 << 14) |
--- a/artiq/coredevice/dac34h84.py
+++ b/artiq/coredevice/dac34h84.py
@@ -110,7 +110,7 @@ class DAC34H84:
     syncsel_mixercd = 0b1001  # sif_sync and register write
     syncsel_nco = 0b1000  # sif_sync
     syncsel_fifo_input = 0b10  # external lvds istr
-    sif_sync = 1
+    sif_sync = 0
 
     syncsel_fifoin = 0b0010  # istr
     syncsel_fifoout = 0b0100  # ostr
--- a/artiq/coredevice/phaser.py
+++ b/artiq/coredevice/phaser.py
@@ -323,6 +323,7 @@ class Phaser:
             delay(.2*ms)
             for data in channel.trf_mmap:
                 channel.trf_write(data)
+            channel.cal_trf_vco()
 
             delay(2*ms)  # lock
             if not (self.get_sta() & (PHASER_STA_TRF0_LD << ch)):
@@ -331,6 +332,7 @@ class Phaser:
             if channel.trf_read(0) & 0x1000:
                 raise ValueError("TRF R_SAT_ERR")
             delay(.1*ms)
+            channel.en_trf_out()
 
         # enable dac tx
         self.set_cfg(clk_sel=self.clk_sel)
@@ -689,6 +691,7 @@ class PhaserChannel:
         self.phaser = phaser
         self.index = index
         self.trf_mmap = TRF372017(trf).get_mmap()
+
         self.oscillator = [PhaserOscillator(self, osc) for osc in range(5)]
 
     @kernel
@@ -890,6 +893,32 @@ class PhaserChannel:
         delay((1 + 1)*34*4*ns)
         return self.trf_write(0x00000008 | (cnt_mux_sel << 27),
                               readback=True)
+
+    @kernel
+    def cal_trf_vco(self):
+        """Start calibration of the upconverter (hardware variant) VCO.
+
+        TRF outputs should be disabled during VCO calibration.
+        """
+        self.trf_write(self.trf_mmap[1] | (1 << 31))
+
+    @kernel
+    def en_trf_out(self, rf=1, lo=0):
+        """Enable the rf/lo outputs of the upconverter (hardware variant).
+
+        :param rf: 1 to enable RF output, 0 to disable
+        :param lo: 1 to enable LO output, 0 to disable
+        """
+        data = self.trf_read(0xc)
+        delay(0.1 * ms)
+        # set RF and LO output bits
+        data = data | (1 << 12) | (1 << 13) | (1 << 14)
+        # clear to enable output
+        if rf == 1:
+            data = data ^ (1 << 14)
+        if lo == 1:
+            data = data ^ ((1 << 12) | (1 << 13))
+        self.trf_write(data)
 
 
 class PhaserOscillator:
--- a/artiq/coredevice/phaser.py
+++ b/artiq/coredevice/phaser.py
@@ -1,6 +1,8 @@
+from numpy import int32, int64
+
 from artiq.language.core import kernel, delay_mu, delay
 from artiq.coredevice.rtio import rtio_output, rtio_input_data
-from artiq.language.units import us, ns, ms, MHz, dB
+from artiq.language.units import us, ns, ms, MHz
 from artiq.language.types import TInt32
 from artiq.coredevice.dac34h84 import DAC34H84
 from artiq.coredevice.trf372017 import TRF372017
--- a/artiq/coredevice/phaser.py
+++ b/artiq/coredevice/phaser.py
@@ -889,7 +889,7 @@ class PhaserChannel:
         :param phase: NCO phase in turns
         """
         pow = int32(round(phase*(1 << 16)))
-        self.set_duc_phase_mu(pow)
+        self.set_nco_phase_mu(pow)
 
     @kernel
     def set_att_mu(self, data):
--- a/artiq/coredevice/phaser.py
+++ b/artiq/coredevice/phaser.py
@@ -279,6 +279,16 @@ class Phaser:
             else:
                 raise ValueError("DAC alarm")
 
+        # avoid malformed output for: mixer_ena=1, nco_ena=0 after power up
+        self.dac_write(self.dac_mmap[2] >> 16, self.dac_mmap[2] | (1 << 4))
+        delay(40*us)
+        self.dac_sync()
+        delay(100*us)
+        self.dac_write(self.dac_mmap[2] >> 16, self.dac_mmap[2])
+        delay(40*us)
+        self.dac_sync()
+        delay(100*us)
+
         # power up trfs, release att reset
         self.set_cfg(clk_sel=self.clk_sel, dac_txena=0)
 
--- a/artiq/coredevice/phaser.py
+++ b/artiq/coredevice/phaser.py
@@ -229,6 +229,8 @@ class Phaser:
         for data in self.dac_mmap:
             self.dac_write(data >> 16, data)
             delay(40*us)
+        self.dac_sync()
+        delay(40*us)
 
         # pll_ndivsync_ena disable
         config18 = self.dac_read(0x18)
@@ -557,6 +559,15 @@ class Phaser:
         return self.dac_read(0x06, div=257) >> 8
 
     @kernel
+    def dac_sync(self):
+        """Trigger DAC synchronisation for both output channels.
+        If the DAC-NCO is enabled, this applies NCO frequency changes."""
+        config1f = self.dac_read(0x1f)
+        delay(.1*ms)
+        self.dac_write(0x1f, config1f & ~int32(1 << 1))
+        self.dac_write(0x1f, config1f | (1 << 1))
+
+    @kernel
     def get_dac_alarms(self):
         """Read the DAC alarm flags.
 
@@ -768,6 +779,7 @@ class PhaserChannel:
     @kernel
     def set_nco_frequency_mu(self, ftw):
         """Set the NCO frequency.
+        The frequency is only applied after DAC synchronisation.
 
         :param ftw: NCO frequency tuning word (32 bit)
         """
@@ -777,6 +789,7 @@ class PhaserChannel:
     @kernel
     def set_nco_frequency(self, frequency):
         """Set the NCO frequency in SI units.
+        The frequency is only applied after DAC synchronisation.
 
         :param frequency: NCO frequency in Hz (passband from -400 MHz
             to 400 MHz, wrapping around at +- 500 MHz)
--- a/artiq/coredevice/phaser.py
+++ b/artiq/coredevice/phaser.py
@@ -571,6 +571,27 @@ class Phaser:
         self.dac_write(0x1f, config1f | (1 << 1))
 
     @kernel
+    def set_dac_cmix(self, fs_8_step):
+        """Set the DAC coarse mixer frequency for both channels
+
+        The selected coarse mixer frequency becomes active without explicit
+        synchronisation.
+
+        Use of the coarse mixer requires the DAC mixer to be enabled. The mixer
+        can be configured via the `dac` device_db entries.
+
+        :param fs_8_step: coarse mixer frequency shift in 125 MHz steps. This
+            should be an integer between -3 and 4 (inclusive).
+        """
+        # values recommended in data-sheet
+        #         0       1       2       3       4       -3      -2      -1
+        vals = [0b0000, 0b1000, 0b0100, 0b1100, 0b0010, 0b1010, 0b0001, 0b1110]
+        cmix = vals[fs_8_step%8]
+        config0d = self.dac_read(0x0d)
+        delay(.1*ms)
+        self.dac_write(0x0d, (config0d & ~(0b1111 << 12)) | (cmix << 12))
+
+    @kernel
     def get_dac_alarms(self):
         """Read the DAC alarm flags.
 
@@ -784,6 +805,9 @@ class PhaserChannel:
         """Set the NCO frequency.
         The frequency is only applied after DAC synchronisation.
 
+        Use of the NCO requires the DAC mixer to be enabled. The mixer can be
+        configured via the `dac` device_db entries.
+
         :param ftw: NCO frequency tuning word (32 bit)
         """
         self.phaser.dac_write(0x15 + (self.index << 1), ftw >> 16)
@@ -793,6 +817,9 @@ class PhaserChannel:
     def set_nco_frequency(self, frequency):
         """Set the NCO frequency in SI units.
         The frequency is only applied after DAC synchronisation.
+
+        Use of the NCO requires the DAC mixer to be enabled. The mixer can be
+        configured via the `dac` device_db entries.
 
         :param frequency: NCO frequency in Hz (passband from -400 MHz
             to 400 MHz, wrapping around at +- 500 MHz)
--- a/artiq/coredevice/phaser.py
+++ b/artiq/coredevice/phaser.py
@@ -186,6 +186,9 @@ class Phaser:
         is_baseband = hw_rev & PHASER_HW_REV_VARIANT
 
         gw_rev = self.read8(PHASER_ADDR_GW_REV)
+        if debug:
+            print(gw_rev)
+            self.core.break_realtime()
         delay(.1*ms)  # slack
 
         # allow a few errors during startup and alignment since boot
--- a/artiq/coredevice/phaser.py
+++ b/artiq/coredevice/phaser.py
@@ -86,7 +86,8 @@ class Phaser:
     LVDS bus operating at 1 Gb/s per pin pair and processed in the DAC (Texas
     Instruments DAC34H84). On the DAC 2x interpolation, sinx/x compensation,
     quadrature modulator compensation, fine and coarse mixing as well as group
-    delay capabilities are available.
+    delay capabilities are available. If desired, these features my be
+    configured via the `dac` dictionary.
 
     The latency/group delay from the RTIO events setting
     :class:`PhaserOscillator` or :class:`PhaserChannel` DUC parameters all the
@@ -574,7 +575,18 @@ class Phaser:
     @kernel
     def dac_sync(self):
         """Trigger DAC synchronisation for both output channels.
-        If the DAC-NCO is enabled, this applies NCO frequency changes."""
+
+        The DAC sif_sync is de-asserts, then asserted. The synchronisation is
+        triggered on assertion.
+
+        By default, the fine-mixer (NCO) and QMC are synchronised. This
+        includes applying the latest register settings.
+
+        The synchronisation sources may be configured through the `syncsel_x`
+        fields in the `dac` configuration dictionary (see `__init__()`).
+
+        .. note:: Synchronising the NCO clears the phase-accumulator
+        """
         config1f = self.dac_read(0x1f)
         delay(.1*ms)
         self.dac_write(0x1f, config1f & ~int32(1 << 1))
@@ -584,11 +596,12 @@ class Phaser:
     def set_dac_cmix(self, fs_8_step):
         """Set the DAC coarse mixer frequency for both channels
 
+        Use of the coarse mixer requires the DAC mixer to be enabled. The mixer
+        can be configured via the `dac` configuration dictionary (see
+        `__init__()`).
+
         The selected coarse mixer frequency becomes active without explicit
         synchronisation.
-
-        Use of the coarse mixer requires the DAC mixer to be enabled. The mixer
-        can be configured via the `dac` device_db entries.
 
         :param fs_8_step: coarse mixer frequency shift in 125 MHz steps. This
             should be an integer between -3 and 4 (inclusive).
@@ -813,10 +826,12 @@ class PhaserChannel:
     @kernel
     def set_nco_frequency_mu(self, ftw):
         """Set the NCO frequency.
-        The frequency is only applied after DAC synchronisation.
 
-        Use of the NCO requires the DAC mixer to be enabled. The mixer can be
-        configured via the `dac` device_db entries.
+        This method stages the new NCO frequency, but does not apply it.
+
+        Use of the DAC-NCO requires the DAC mixer and NCO to be enabled. These
+        can be configured via the `dac` configuration dictionary (see
+        `__init__()`).
 
         :param ftw: NCO frequency tuning word (32 bit)
         """
@@ -826,10 +841,12 @@ class PhaserChannel:
     @kernel
     def set_nco_frequency(self, frequency):
         """Set the NCO frequency in SI units.
-        The frequency is only applied after DAC synchronisation.
 
-        Use of the NCO requires the DAC mixer to be enabled. The mixer can be
-        configured via the `dac` device_db entries.
+        This method stages the new NCO frequency, but does not apply it.
+
+        Use of the DAC-NCO requires the DAC mixer and NCO to be enabled. These
+        can be configured via the `dac` configuration dictionary (see
+        `__init__()`).
 
         :param frequency: NCO frequency in Hz (passband from -400 MHz
             to 400 MHz, wrapping around at +- 500 MHz)
@@ -841,6 +858,16 @@ class PhaserChannel:
     def set_nco_phase_mu(self, pow):
         """Set the NCO phase offset.
 
+        By default, the new NCO phase applies on completion of the SPI
+        transfer. This also causes a staged NCO frequency to be applied.
+        Different triggers for applying nco settings may be configured through
+        the `syncsel_mixerxx` fields in the `dac` configuration dictionary (see
+        `__init__()`).
+
+        Use of the DAC-NCO requires the DAC mixer and NCO to be enabled. These
+        can be configured via the `dac` configuration dictionary (see
+        `__init__()`).
+
         :param pow: NCO phase offset word (16 bit)
         """
         self.phaser.dac_write(0x12 + self.index, pow)
@@ -848,6 +875,16 @@ class PhaserChannel:
     @kernel
     def set_nco_phase(self, phase):
         """Set the NCO phase in SI units.
+
+        By default, the new NCO phase applies on completion of the SPI
+        transfer. This also causes a staged NCO frequency to be applied.
+        Different triggers for applying nco settings may be configured through
+        the `syncsel_mixerxx` fields in the `dac` configuration dictionary (see
+        `__init__()`).
+
+        Use of the DAC-NCO requires the DAC mixer and NCO to be enabled. These
+        can be configured via the `dac` configuration dictionary (see
+        `__init__()`).
 
         :param phase: NCO phase in turns
         """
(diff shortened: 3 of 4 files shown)


artiq/coredevice/phaser.py
```diff
@@ -549,6 +568,50 @@ def get_dac_temperature(self) -> TInt32:
...
         return self.dac_read(0x06, div=257) >> 8
 
+    @kernel
+    def dac_sync(self):
+        """Trigger DAC synchronisation for both output channels.
+        If the DAC-NCO is enabled, this applies NCO frequency changes."""
+        config_1f = self.dac_read(0x1f)
+        delay(.1*ms)
+        self.dac_write(0x1f, config_1f | (1 << 1))
+        self.dac_write(0x1f, config_1f & ~int32(1 << 1))
```
jordens, line 578: No need for the int32 here.

artiq/coredevice/phaser.py
```diff
@@ -549,6 +568,50 @@ def get_dac_temperature(self) -> TInt32:
         """
         return self.dac_read(0x06, div=257) >> 8
 
+    @kernel
+    def dac_sync(self):
+        """Trigger DAC synchronisation for both output channels.
+        If the DAC-NCO is enabled, this applies NCO frequency changes."""
```
jordens, line 574: No. It may do all kinds of things depending on the sync enable bits. And if NCO sync from sif_sync is not set, then it precisely doesn't apply the NCO frequency change.

artiq/coredevice/phaser.py
```diff
@@ -1,6 +1,8 @@
+from numpy import int32, int64
```
jordens, line 1: remove int64

RELEASE_NOTES.rst
```diff
@@ -9,8 +9,10 @@ ARTIQ-7
 Highlights:
 * WRPLL
 * ``get()``, ``get_mu()``, ``get_att()``, and ``get_att_mu()`` functions added for AD9910 and AD9912
+* Phaser coredevice driver: refactor, bug fixes and added documentation
```
jordens, line 12: This is not a refactor. The important point that users need to be aware of for their existing code is that the defaults and the behavior changes. Added documentation and vague "bug fixes" doesn't mean anything for existing code.

artiq/coredevice/phaser.py
```diff
@@ -886,6 +958,30 @@ def trf_read(self, addr, cnt_mux_sel=0) -> TInt32:
...
+    @kernel
+    def cal_vco(self):
+        """Start VCO calibration."""
+        self.trf_write(self.trf_mmap[1] | (1 << 31))
+
+    @kernel
+    def en_out(self, rf=1, lo=0):
+        """Enable the rf/lo outputs.
+
+        It is recommended to disable TRF outputs during VCO calibration.
```
jordens, line 970: You only call this when enabling it making that comment confusing.

artiq/coredevice/phaser.py
```diff
@@ -886,6 +958,30 @@ def trf_read(self, addr, cnt_mux_sel=0) -> TInt32:
         return self.trf_write(0x00000008 | (cnt_mux_sel << 27),
                               readback=True)
 
+    @kernel
+    def cal_vco(self):
+        """Start VCO calibration."""
+        self.trf_write(self.trf_mmap[1] | (1 << 31))
```
jordens, line 964: Why do you write from the map here but from the readback elsewhere?

artiq/coredevice/phaser.py
```diff
@@ -886,6 +958,30 @@ def trf_read(self, addr, cnt_mux_sel=0) -> TInt32:
...
+
+    @kernel
+    def en_out(self, rf=1, lo=0):
+        """Enable the rf/lo outputs.
+
+        It is recommended to disable TRF outputs during VCO calibration.
+
+        :param rf: 1 to enable RF output, 0 to disable
+        :param lo: 1 to enable LO output, 0 to disable"""
+        data = self.trf_read(0xc)
```
jordens, line 974: Why not take the value in `trf_mmap` as above?

artiq/coredevice/phaser.py
```diff
@@ -886,6 +958,30 @@ def trf_read(self, addr, cnt_mux_sel=0) -> TInt32:
...
+        delay(0.1 * ms)
+        # outputs disabled
+        data = data | (1 << 12) | (1 << 13) | (1 << 14)
+        # toggle if enabled
```
jordens, line 978: Do you mean "clear to enable"?

artiq/coredevice/trf372017.py
```diff
@@ -4,20 +4,20 @@ class TRF372017:
...
     icp_double = 0
-    cal_clk_sel = 12  # /16, 4b
+    cal_clk_sel = 0b1110  # div64, 4b
 
-    ndiv = 420  # 16b
-    pll_div_sel = 0  # /1, 2b
-    prsc_sel = 1  # 8/9
+    nint = 23  # 16b - lowest value suitable for fractional & integer mode
+    pll_div_sel = 0b01  # div2, 2b
+    prsc_sel = 0  # 4/5
```
jordens, line 16: Why change that?

artiq/coredevice/trf372017.py
```diff
@@ -4,20 +4,20 @@ class TRF372017:
...
     neg_vco = 1
     icp = 0  # 1.94 mA, 5b
     icp_double = 0
-    cal_clk_sel = 12  # /16, 4b
+    cal_clk_sel = 0b1110  # div64, 4b
 
-    ndiv = 420  # 16b
-    pll_div_sel = 0  # /1, 2b
-    prsc_sel = 1  # 8/9
+    nint = 23  # 16b - lowest value suitable for fractional & integer mode
```
jordens, line 14: What is the LO frequency with this? That would be good to document.
Do you approximate the previous default frequency or is there a reason to change that?

artiq/coredevice/trf372017.py
```diff
@@ -59,8 +59,8 @@ class TRF372017:
     ioff = 0x80  # 8b
     qoff = 0x80  # 8b
     vref_sel = 4  # 0.85 V, 3b
-    tx_div_sel = 1  # div2, 2b
-    lo_div_sel = 3  # div8, 2b
+    tx_div_sel = 0  # div1, 2b
```
jordens, line 62: Why change these?

artiq/coredevice/trf372017.py
```diff
@@ -84,6 +84,9 @@ def __init__(self, updates=None):
             setattr(self, key, value)
 
     def get_mmap(self):
+        """Memory map for TRF372017
+
+        VCO calibration, and RF/LO outputs should be initially disabled"""
```
jordens, line 89: That's a "must" now. And it's unclear what the user needs to do to enable them.

artiq/coredevice/phaser.py
```diff
@@ -549,6 +568,50 @@ def get_dac_temperature(self) -> TInt32:
...
+            cmix = 0b0010
+        elif fs_8_step == -1:
+            cmix = 0b1110
+        elif fs_8_step == -2:
+            cmix = 0b0001
+        elif fs_8_step == -3:
+            cmix = 0b1010
+        else:
+            raise ValueError("Invalid coarse mixer freqeuncy")
+        config_0d = self.dac_read(0x0d)
```
jordens, line 611: `config0d` would be consistent.

artiq/coredevice/phaser.py
```diff
@@ -549,6 +568,50 @@ def get_dac_temperature(self) -> TInt32:
...
+        The selected coarse mixer frequency becomes active without explicit
+        synchronisation.
+
+        Use of the coarse mixer requires the DAC mixer to be enabled. The mixer
+        can be configured via the `dac` device_db entries.
+
+        :param fs_8_step: coarse mixer frequency shift in 125 MHz steps. This
+            should be an integer between -3 and 4 (inclusive).
+        """
+        if fs_8_step == 0:
```
jordens, line 593: Can this become a lookup `cmix = [0x000, ...][fs_8_steps]`, or better yet just a computation with one or two bit shuffles?
Exploit the periodicity of the Nyquist zone and twos complement (`-4 === 4` and `-1 === 7` etc).

artiq/coredevice/phaser.py
```diff
@@ -549,6 +568,50 @@ def get_dac_temperature(self) -> TInt32:
...
+
+    @kernel
+    def set_dac_cmix(self, fs_8_step):
+        """Set the DAC coarse mixer frequency for both channels
+
+        The selected coarse mixer frequency becomes active without explicit
+        synchronisation.
+
+        Use of the coarse mixer requires the DAC mixer to be enabled. The mixer
+        can be configured via the `dac` device_db entries.
```
jordens, line 588: Do you mean `... must be enabled via..`? And the fact that those may come from the device_db is a technical detail that is not guaranteed. You can also instantiate Phaser in different ways. I'd remove that in favor of saying "must be enabled through the `dac` configuration dictionary (see init())`.

RELEASE_NOTES.rst
```diff
@@ -9,8 +9,10 @@ ARTIQ-7
...
 
 Breaking changes:
+* Updated Phaser-Upconverter default frequency configuration. The new default uses the design target PFD frequency
```
jordens, line 15: Also breaking changes to sif_sync, oscillators, etc!

artiq/coredevice/dac34h84.py
```diff
@@ -200,7 +201,7 @@ def get_mmap(self):
         mmap.append(
             (0x0d << 16) |
             (self.cmix_fs8 << 15) | (self.cmix_fs4 << 14) |
-            (self.cmix_fs2 << 12) | (self.cmix_nfs4 << 11) |
+            (self.cmix_fs2 << 13) | (self.cmix_nfs4 << 12) |
```
jordens, line 204: Confirmed.

artiq/coredevice/phaser.py
```diff
@@ -549,6 +568,50 @@ def get_dac_temperature(self) -> TInt32:
         """
         return self.dac_read(0x06, div=257) >> 8
 
+    @kernel
+    def dac_sync(self):
+        """Trigger DAC synchronisation for both output channels.
+        If the DAC-NCO is enabled, this applies NCO frequency changes."""
+        config_1f = self.dac_read(0x1f)
+        delay(.1*ms)
+        self.dac_write(0x1f, config_1f | (1 << 1))
```
jordens, line 577: As I explained, it's better to first clear and then set, and leave it set. Assuming the rising edge is what matters, this makes it idempotent. The way you do it will fail if someone sets it in the device_db.

artiq/coredevice/phaser.py
```diff
@@ -549,6 +568,50 @@ def get_dac_temperature(self) -> TInt32:
         """
         return self.dac_read(0x06, div=257) >> 8
 
+    @kernel
+    def dac_sync(self):
+        """Trigger DAC synchronisation for both output channels.
+        If the DAC-NCO is enabled, this applies NCO frequency changes."""
+        config_1f = self.dac_read(0x1f)
```
jordens, line 575:
```suggestion
        config1f = self.dac_read(0x1f)
```
would be consistent.

artiq/coredevice/phaser.py
```diff
@@ -760,6 +824,10 @@ def set_duc_phase(self, phase):
     @kernel
     def set_nco_frequency_mu(self, ftw):
         """Set the NCO frequency.
+        The frequency is only applied after DAC synchronisation.
+
+        Use of the NCO requires the DAC mixer to be enabled. The mixer can be
```
jordens, line 829: And presumably also the NCO needs to be enabled? Copy paste error?

artiq/coredevice/phaser.py
```diff
@@ -886,6 +958,30 @@ def trf_read(self, addr, cnt_mux_sel=0) -> TInt32:
...
+        """Enable the rf/lo outputs.
+
+        It is recommended to disable TRF outputs during VCO calibration.
+
+        :param rf: 1 to enable RF output, 0 to disable
+        :param lo: 1 to enable LO output, 0 to disable"""
+        data = self.trf_read(0xc)
+        delay(0.1 * ms)
+        # outputs disabled
+        data = data | (1 << 12) | (1 << 13) | (1 << 14)
```
jordens, line 977: You want to say which flags these are.

artiq/coredevice/trf372017.py
```diff
@@ -4,20 +4,20 @@ class TRF372017:
...
-    pll_div_sel = 0  # /1, 2b
-    prsc_sel = 1  # 8/9
+    nint = 23  # 16b - lowest value suitable for fractional & integer mode
+    pll_div_sel = 0b01  # div2, 2b
+    prsc_sel = 0  # 4/5
     vco_sel = 2  # 2b
     vcosel_mode = 0
     cal_acc = 0b00  # 2b
-    en_cal = 1
+    en_cal = 0
```
jordens, line 20: You need to document that the user must not change that.

artiq/coredevice/dac34h84.py
```diff
@@ -110,7 +110,7 @@ class DAC34H84:
     syncsel_mixercd = 0b1001  # sif_sync and register write
     syncsel_nco = 0b1000  # sif_sync
     syncsel_fifo_input = 0b10  # external lvds istr
-    sif_sync = 1
+    sif_sync = 0
```
jordens, line 113: Document that the user must not change that.

artiq/coredevice/phaser.py
```diff
@@ -549,6 +568,50 @@ def get_dac_temperature(self) -> TInt32:
         """
         return self.dac_read(0x06, div=257) >> 8
 
+    @kernel
+    def dac_sync(self):
+        """Trigger DAC synchronisation for both output channels.
```
jordens, line 573: Explain that this only synchronizes the DAC-internal functionality where `sif_sync` is enabled enabled.

jordens (changes requested):
Forgot to say this: Thanks for the PR. This is a valuable step into the right direction!

> The current Phaser core device driver requires users perform register level manipulations of the DAC to use/change the frequency of the coarse mixer and NCO. This PR adds methods that allow users to reconfigure the NCO and coarse mixer frequency without needing to refer to the DAC-datasheet.

Unfortunately I suspect your PR will lead to regressions. It will break use cases where the synchronization doesn't happen via `sif_sync` but instead via other synchronization means (like toggling a different sync source flag).

Also, since you now special-case `sif_sync` in the initialization, you need to document that the user must not use it in the settings dict anymore (or for that matter can't use it at all, since you now rely on it being in a certain value or else `dac_sync` breaks). You may actually want to remove it there.
This is the tricky thing with abstracting over an API: it's easy to break things. If you want to do it, you need to be more careful.

> Document the need to enable the DAC mixer in order to use the NCO

Why is that bit of information needed here but not any other from the datasheet?
Unfortunately this PR isn't very consistent. Or is the intent not to add/update documentation at all?

> An offset error for the DAC coarse mixer bits is fixed.

Ack. Good! You could land that right away.

> Phaser.channel.oscillator[0] is no-longer special cased after init(). One of the main features of Phaser is the ability to pulse shape. In this context, enabling RF output without specifying an amplitude leads to breakdown of abstraction.

Ok. But if you rationale is removing footguns and special cases, you forget about the phase offset. If you want to be consistent, you should also clear phase offset for all oscillators after the test.

> The TRF Phaser-Frequency-Detector (PFD) clock now defaults to the hardware design target of 62.5 MHz. This configuration should offer the lowest phase noise and will be a reasonable starting point for users to perform adjustments. Currently the PFD is set to an arbitrarily chosen frequency.

But why change the default frequency values?

> Refactor the TRF calibration procedure to follow that described in the data-sheet.

You need to document the effect on the usability and constraints of the configuration flags in the dictionary.

> Refactor the NDIV -> NINT to match the documentation naming.

That's not a refactor, but a breaking rename.

> Work around a bug that leads to malformed output when initialising phaser with mixer_ena=1 & nco_ena=0 after a power cycle.

Could you describe that in more detail? I've seen unsynchronized mixing but I had the impression that synchronization always fixed it.

artiq/coredevice/trf372017.py
pathfinder49: The old value is no-longer valid with the new PFD clock

jordens: Ack.

artiq/coredevice/trf372017.py
pathfinder49: I'm not aware of a requirement to do so. It is against the manufacturer recommendation and may lead to trouble with the calibration in marginal cases. The PR adds methods in the phaser driver to enable these RF/LO outputs and VCO calibration. It seems reasonable for users to look through the phaser documentation/source to find these. Most users will also not need to touch this. The VCO is calibrated and the TRF RF output is enabled in `init()`.

artiq/coredevice/trf372017.py
pathfinder49: The VCO frequency already needed to be changed to accommodate the new PFD default. As these outputs are already changing, it seemed sensible to consolidate their values. Defaulting to `div1` seems sensible, I'm open to suggestions for other values.

artiq/coredevice/trf372017.py
pathfinder49: See below

artiq/coredevice/trf372017.py
jordens: If the user enables calibration initially, the calibration that you implement will break, no?

Still, if a user sees "RF output should initially be disabled" they ask themselves what they need to do to (a) heed that advice and (b) enable RF out.

artiq/coredevice/trf372017.py
pathfinder49: This does not approximate the previous frequency. With these setting the default VCO frequency is 2.875 GHz. This was selected as it is the lowest VCO frequency allowing both fractional and integer mode. These values are likely to be reconfigured when used in an experiment. To me, the most sensible default is the one requiring the least modifications in test-setups.

artiq/coredevice/phaser.py
pathfinder49: Yes, though these amount to the same thing here.

artiq/coredevice/phaser.py
pathfinder49: This register contains the `en_frac` bit. Some users may need to reconfigure the VCO frequency in their experiments This may include switching between integer and fractional mode. It seems imprudent to assume this bit is not changed by any user.

artiq/coredevice/phaser.py
pathfinder49: Fair, `en_trf_out` would be better

artiq/coredevice/phaser.py
pathfinder49: I agree this should be readback. I encountered a bug I haven't got to the bottom of when using read-back for this method. I can investigate further if you believe this to be a major issue, it does prevent users from changing many VCO frequency settings

artiq/coredevice/phaser.py
pathfinder49: sure

artiq/coredevice/phaser.py
pathfinder49: Absolutely, The main reason for adding this was the confusion in #1630. Happy to also reference the need to enable the NCO.

artiq/coredevice/phaser.py
pathfinder49: Ack, this could be done more neatly. It's a backwards twos complement.

artiq/coredevice/phaser.py
pathfinder49: > "must be enabled through the `dac` configuration dictionary (see init())`.

Ack.

The the mixer can also be enabled by manually using `dac_write`. The device_db comment is intended to point unfamiliar users towards an appropriate place to configure this.

artiq/coredevice/phaser.py
pathfinder49: sure

artiq/coredevice/phaser.py
jordens: Maybe there is something wrong with readback. But I don't know whether that's due to the chip or not. Unclear whether a register can actually be fully read back. https://github.com/quartiq/phaser/issues/17

artiq/coredevice/phaser.py
pathfinder49: > No. It may do all kinds of things depending on the sync enable bits.

Absolutely. By default, updates to the QMC and NCO frequency are triggered by sif_sync. Of these only the NCO frequency is directly exposed to the user. It seems fair for those using `dac_write` or, changing non-frequency related fields in the `dac` configuration dictionary, should be familiar with the DAC data-sheet.

artiq/coredevice/phaser.py
pathfinder49: See below. It is probably worth adding another paragraph here.

artiq/coredevice/phaser.py
pathfinder49: It is used here https://github.com/pathfinder49/artiq/blob/e74f1d20f6af03ddd3dab706948c4218d7e183f4/artiq/coredevice/phaser.py#L360
TBH, I'm not sure how the current code compiles given its not imported.

RELEASE_NOTES.rst
pathfinder49: Ack, breaking changes are:
* New LO frequency
* New Kasli oscillator state after `init`

RELEASE_NOTES.rst
pathfinder49: Sure

artiq/coredevice/phaser.py
jordens: Yes. Some globals are available in ARTIQ-Python beyond the standard Python globals.

artiq/coredevice/phaser.py
jordens: Ok. The we should make that a rule to use current state as much as possible.

pathfinder49:
> Could you describe that in more detail? I've seen unsynchronized mixing but I had the impression that synchronization always fixed it.

Sure, I can open an issue. Where would be an appropriate place to discuss this?

artiq/coredevice/trf372017.py
jordens: Ah. But the divider was 2. So the LO was at ~1.4 GHz (https://github.com/quartiq/phaser/issues/3).
Then those values are fine with me. Can you add a line saying what frequency the LO and VCO are at with this?

artiq/coredevice/trf372017.py
jordens: Sounds good.

artiq/coredevice/phaser.py
jordens: Keep them like this then.

artiq/coredevice/phaser.py
jordens: Then we should rather do that (tell people that the initial settings dictionaries are accessible through the device db) summarily in the class or __init__ docstring and not in every method.

RELEASE_NOTES.rst
jordens: And the abstraction over and ownership of `sif_sync`.

artiq/coredevice/phaser.py
jordens: That may be tricky because of the temporal granularity of the commands in the control data stream.
But again to clarify: is it true that the rising edge of `sif_sync` does the synchronization (i.e. not level sensitive and not falling egde)? Then I would strongly prefer this method to do "clear-set" and not "set-clear" for robustness reasons.

artiq/coredevice/dac34h84.py
jordens: Are you sure? `dac_sync` does set, then clear. If `sif_sync` is set earlier during the mmap writing, then this is set-set-clear and the only rising edge is the first one during mmap-writing. (under the assumptions of https://github.com/m-labs/artiq/pull/1657#discussion_r614159355).

jordens:
> Sure, I can open an issue. Where would be an appropriate place to discuss this?

Sounds like a coredevice driver issue (you are also addressing it here). -> ARTIQ.

artiq/coredevice/trf372017.py
pathfinder49: Sure

artiq/coredevice/dac34h84.py
pathfinder49: Either way, there would be one rising edge. All relevant configuration data is written before sif_sync when writing the dac mmap, either timing is acceptable.

artiq/coredevice/trf372017.py
pathfinder49: > If the user enables calibration initially, the calibration that you implement will break, no?

It is unspecified behaviour


> "RF output should initially be disabled" they ask themselves what they need to do to (a) heed that advice and (b) enable RF out.

Ack, I'll add comments to the bit entries instead.

artiq/coredevice/dac34h84.py
jordens: But then your entire line of argument about `sif_sync=1` by default being wrong and it not being toggled in init breaks down.
Then let's keep it at 1 by default, remove the `dac_sync()` in init, and make it clear-then-set.
Less code is typically better. And this would also remove the implicit `sif_sync=0` predicate for `dac_sync()`.

pathfinder49: Ok, I'll update it to clear then set.

jordens:
Ok. Ping me when you want another review here.

artiq/coredevice/phaser.py
pathfinder49: My initial impression was, that there was a compiler issue. Printing mmap and readback values gave the same result. However, the hardware responded differently when using one or the other (no printing)

pathfinder49:
I've cleaned most of this up and could push a draft for review. There are some issues that I haven't had the time to track down. Namely, the readback issue for the TRF VCO calibration and NCO phase updates not updating the double-buffered NCO settings. (Beyond a copy paste error in the current set_nco_phase implementation).

pathfinder49:
> 
> 
> I've cleaned most of this up and could push a draft for review. There are some issues that I haven't had the time to track down. Namely, the readback issue for the TRF VCO calibration and NCO phase updates not updating the double-buffered NCO settings. (Beyond a copy paste error in the current set_nco_phase implementation).

Both of these issues aren't regressions from from the current code. It's unclear to me whether they should block this PR.

jordens:
They shouldn't.

One note: We should generally reduce comment and documentation duplication and only document one of the `_mu()`/-non`_mu()` methods and refer to the documentation of the other.

jordens:
@pathfinder49 ready for review?

jordens (approved):

pathfinder49:
If those issues aren't blocking this is readty for review

PR #1658, opened 2021-04-15
sinara_tester: Skip tests for SUServos

HarryMakes:
# ARTIQ Pull Request

## Description of Changes

This skips tests for all Urukul CPLDs and AD9910s belonging to any SUServos as defined in the given `device_db.py`.

Note that this is currently applied to **ARTIQ-6**, and I have only tested it on ARTIQ-6 platforms.

### Related Issue

Currently, the tester scans and tests all Urukul CPLDs and AD9910s as if the boards are standalone, but the calls to their class method are not fully compatible with Urukul in SUServo modes. For example, the SUServo CPLDs and AD9910s need to be init'ed "blindly" due to the lack of readback, and the AD9910s' `chip_select` must be 3 for multi-channel control via `CFG.MASK_NU` instead of 4-7 for single-channel control. Plus, the SUServo Urukuls do not use a dedicated EEM LVDS pair as for IO_UPDATE, so the tester should not attempt to find a `io_update_device` TTLOut channel for these CPLDs.

If the user intends to test SUServo functionalities on the boards, they should try out the [Kasli-SUServo example](https://github.com/m-labs/artiq/tree/master/artiq/examples/kasli_suservo).

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |
|   | :sparkles: New feature |
|   | :hammer: Refactoring  |
|   | :scroll: Docs |

### DIFF
--- a/artiq/frontend/artiq_sinara_tester.py
+++ b/artiq/frontend/artiq_sinara_tester.py
@@ -56,6 +56,8 @@ class SinaraTester(EnvExperiment):
         self.mirny_cplds = dict()
         self.mirnies = dict()
 
+        suservo_arg_devices = []    # for collecting CPLDs & AD9910s used for SUServos
+
         ddb = self.get_device_db()
         for name, desc in ddb.items():
             if isinstance(desc, dict) and desc["type"] == "local":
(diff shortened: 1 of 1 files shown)


sbourdeauducq:
What about adding a test for SU-Servo (based on the example that you linked) to artiq_sinara_tester instead?

PR #1676, opened 2021-05-17
jsonschema: only allow enable_sata_drtio=true for Kasli if v1.0/1.1

HarryMakes:
# ARTIQ Pull Request

## Description of Changes

When validating with jsonschema, if `target` is `kasli`, `enable_sata_drtio` must be `false` (which is the default value) unless `hw_rev` is v1.0 or v1.1.

This respects the FPGA bank I/O definitions on [Migen](https://github.com/m-labs/migen/blob/master/migen/build/platforms/sinara/kasli.py). For example, v2.0 in fact does not have a SATA connector.

Prior to this PR, if `enable_sata_drtio` is `true` for Kasli v2.0 targets, simply a `Resource not found` error is raised from Migen when the SoC builder is called from `gateware.targets.kasli_generic`. An error raised from JSON validation, instead, might be more helpful than the error raised by Migen.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |
|   | :sparkles: New feature |
|   | :hammer: Refactoring  |
|   | :scroll: Docs |

### DIFF
--- a/artiq/coredevice/coredevice_generic.schema.json
+++ b/artiq/coredevice/coredevice_generic.schema.json
@@ -71,6 +71,26 @@
             }
         }
     },
+    "if": {
+        "properties": {
+            "target": { "const": "kasli" },
+            "hw_rev": {
+                "not": {
+                    "oneOf": [
+                        { "const": "v1.0" },
+                        { "const": "v1.1" }
+                    ]
+                }
+            }
+        }
+    },
+    "then": {
+        "properties": {
+            "enable_sata_drtio": {
+                "const": false
+            }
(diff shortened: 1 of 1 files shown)


HarryMakes:
In hindsight, perhaps adding this check on the `oneOf` check below would fit the structure more properly. We (or I) should do that if there are more constraints to add in the future. 

https://github.com/m-labs/artiq/blob/fc59791583905f9e35a4ef0bc997c21ebc649a64/artiq/coredevice/coredevice_generic.schema.json#L78-L89

PR #1696, opened 2021-06-09
enables custom length input FIFOs with DIOs

### LINKED ISSUE #1693: larger RTIO input buffer?
# Question

Would it be possible to make kasli be able to buffer more input timestamps?

## Description

When using functions such as `gate_rising_mu` and `timestamp_mu` to collect and readout timestamps, only a small number of RTIO inputs (64?) can be buffered by kasli before it raises a `RTIOOverflow` exception. Would it be possible to increase the number of timestamps it can buffer to a significantly larger number such as 1024?

fanmingyu212:
# ARTIQ Pull Request

## Description of Changes

Enables the user to set a custom length input FIFO on DIO channels.

Added a new `custom_ififo_lengths` key in `dio` objects in the "`__variant__.json`" description file to set custom input FIFO lengths for some channels. The value is a JSON key value object of `{"channel_index": ififo_length}`. See an example below

```json
{
    "type": "dio",
    "ports": [1],
    "bank_direction_low": "input",
    "bank_direction_high": "output",
    "edge_counter": true,
    "custom_ififo_lengths": {"0": 1024}
},
```

I tested the above code on a Kasli and it increases the input FIFO length of the first channel (index 0) of the DIO to 1024. I don't think this change would work on KC705 or ZC706 boards though.

### Related Issue

Closes #1693 

## Type of Changes

|   | Type |

### All Pull Requests

- [x] Close/update issues.

### Code Changes

### Git Logistics

  ```
  
  ```

### Licensing

### DIFF
--- a/artiq/coredevice/coredevice_generic.schema.json
+++ b/artiq/coredevice/coredevice_generic.schema.json
@@ -168,6 +168,10 @@
                         "bank_direction_high": {
                             "type": "string",
                             "enum": ["input", "output"]
+                        },
+                        "custom_ififo_lengths": {
+                            "type": "object",
+                            "default": {}
                         }
                     },
                     "required": ["ports", "bank_direction_low", "bank_direction_high"]
--- a/artiq/gateware/eem.py
+++ b/artiq/gateware/eem.py
@@ -43,7 +43,16 @@ class DIO(_EEM):
 
     @classmethod
     def add_std(cls, target, eem, ttl03_cls, ttl47_cls, iostandard=default_iostandard,
-            edge_counter_cls=None):
+            edge_counter_cls=None, custom_ififo_lengths={}):
+
+        def append_to_rtio_channels(phy, ch_index):
+            ch_index_str = str(ch_index)
+            if ch_index_str in custom_ififo_lengths:
+                ififo_depth = custom_ififo_lengths[ch_index_str]
+                target.rtio_channels.append(rtio.Channel.from_phy(phy, ififo_depth=ififo_depth))
+            else:
+                target.rtio_channels.append(rtio.Channel.from_phy(phy))
+
         cls.add_extension(target, eem, iostandard=iostandard)
 
         phys = []
@@ -53,13 +62,14 @@ class DIO(_EEM):
             phy = ttl03_cls(pads.p, pads.n, dci=dci)
             phys.append(phy)
             target.submodules += phy
-            target.rtio_channels.append(rtio.Channel.from_phy(phy))
+            append_to_rtio_channels(phy, i)
         for i in range(4):
-            pads = target.platform.request("dio{}".format(eem), 4+i)
+            ch_index = 4 + i
+            pads = target.platform.request("dio{}".format(eem), ch_index)
             phy = ttl47_cls(pads.p, pads.n, dci=dci)
             phys.append(phy)
             target.submodules += phy
-            target.rtio_channels.append(rtio.Channel.from_phy(phy))
+            append_to_rtio_channels(phy, ch_index)
 
         if edge_counter_cls is not None:
             for phy in phys:
--- a/artiq/gateware/eem_7series.py
+++ b/artiq/gateware/eem_7series.py
@@ -16,7 +16,8 @@ def peripheral_dio(module, peripheral, **kwargs):
     eem.DIO.add_std(module, peripheral["ports"][0],
         ttl_classes[peripheral["bank_direction_low"]],
         ttl_classes[peripheral["bank_direction_high"]],
-        edge_counter_cls=edge_counter_cls, **kwargs)
+        edge_counter_cls=edge_counter_cls,
+        custom_ififo_lengths=peripheral["custom_ififo_lengths"], **kwargs)
 
 
 def peripheral_urukul(module, peripheral, **kwargs):


artiq/gateware/eem.py
```diff
@@ -53,13 +54,21 @@ def add_std(cls, target, eem, ttl03_cls, ttl47_cls, iostandard=default_iostandar
...
+                target.rtio_channels.append(rtio.Channel.from_phy(phy, ififo_depth=ififo_depth))
+            else:
+                target.rtio_channels.append(rtio.Channel.from_phy(phy))
         for i in range(4):
             pads = target.platform.request("dio{}".format(eem), 4+i)
             phy = ttl47_cls(pads.p, pads.n, dci=dci)
             phys.append(phy)
             target.submodules += phy
-            target.rtio_channels.append(rtio.Channel.from_phy(phy))
+            if 4+i in custom_ififo_chs:
```
sbourdeauducq, line 67: store 4+i somewhere

artiq/coredevice/coredevice_generic.schema.json
```diff
@@ -168,6 +168,10 @@
                         "bank_direction_high": {
                             "type": "string",
                             "enum": ["input", "output"]
+                        },
+                        "custom_ififo_lengths": {
+                            "type": "array",
```
sbourdeauducq, line 173: Why not a dictionary?

fanmingyu212: Is there a dictionary object in json?

sbourdeauducq: Sure

fanmingyu212: If I use the `{"0": 1024}` syntax, I think the channel number has to be a string (while it should be an int). If you think that is better, I can make that change.

sbourdeauducq: That's standard enough that ``json.dumps`` silently converts the keys to strings.

artiq/gateware/eem.py
```diff
@@ -53,13 +53,22 @@ def add_std(cls, target, eem, ttl03_cls, ttl47_cls, iostandard=default_iostandar
             phy = ttl03_cls(pads.p, pads.n, dci=dci)
             phys.append(phy)
             target.submodules += phy
-            target.rtio_channels.append(rtio.Channel.from_phy(phy))
+            if i in custom_ififo_lengths:
```
sbourdeauducq, line 56: Did you test this?

artiq/gateware/eem.py
```diff
@@ -53,13 +53,22 @@ def add_std(cls, target, eem, ttl03_cls, ttl47_cls, iostandard=default_iostandar
...
+                target.rtio_channels.append(rtio.Channel.from_phy(phy))
         for i in range(4):
-            pads = target.platform.request("dio{}".format(eem), 4+i)
+            ch_index = 4 + i
+            pads = target.platform.request("dio{}".format(eem), ch_index)
             phy = ttl47_cls(pads.p, pads.n, dci=dci)
             phys.append(phy)
             target.submodules += phy
-            target.rtio_channels.append(rtio.Channel.from_phy(phy))
+            if ch_index in custom_ififo_lengths:
```
sbourdeauducq, line 67: Factor this (l67-l71).

artiq/gateware/eem.py
fanmingyu212: Testing now and I see why it would not work.

fanmingyu212:
@sbourdeauducq now the code is tested and should work.

What is the best place to document changes like this? Maybe adding a section in the manual for details about developing would be good, as said in https://github.com/m-labs/artiq/issues/1693#issuecomment-856315228? Documentation could be a separate PR. I don't feel confident enough to make changes such as adding a section to the manual.

dhslichter:
@fanmingyu212 what is the rationale for doing this as a modification of `add_std()` rather than making a new `add_custom()` or similar?  The idea of `add_std()` is to be a "default" method for adding an EEM.  If you want to do something different, it makes more sense to me to make a second method which enable more fine-grained control, and leave `add_std()` alone.  This has the additional benefit that it doesn't mess with anyone's existing code that uses `add_std()` (for example introducing weird corner cases).

fanmingyu212:
@dhslichter it sounds like a good idea to make a new `add_custom` function. The new function can be called in `eem_7series.py` so this change would not impact other gateware targets. I will implement that.

sbourdeauducq:
That sounds messy with code duplication and unmaintainable. Your modification to add_std was backward compatible.

fanmingyu212:
If we want to add a new `add_custom` function, we might want to let the `add_std` function call `add_custom` with some of the parameters set to default. This does not seem like the design philosophy of the `eem.py` module, as in all classes in `eem.py` the `add_std` function is the only function that defines the hardware. Thus it could make sense to change `add_std` directly.

I was concerned about breaking code in other gateware targets other than `eem_7series` if `add_std` is changed, but @sbourdeauducq pointed out that the function should be backward compatible.

dhslichter:
I'm fine with this, the `append_to_rtio_channels()` method basically does what I would want an `add_custom()` method to do and this plays more nicely with the json methods for building Kasli bitstreams.  I agree that it's better to avoid code duplication.

han94ros:
In which file do I set this line
""custom_ififo_lengths": {"0": 1024}"

fanmingyu212:
The code in this PR was developed for larger input buffers to timestamp many PMT counts. We found out in usual conditions (<100 kCt/s, Kasli 2.0), the default buffer will not overflow if we continuously read out timestamps in the buffer into an array in the kernel. Therefore, I will close this PR for now.

PR #1704, opened 2021-06-25
compiler: allows string annotation

pca006132:
# ARTIQ Pull Request

## Description of Changes

According to PEP484, type hint can be a string literal for forward references. With PEP563, type hint would be preserved in annotations in string form.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

### All Pull Requests

### Code Changes

Test cases:
```python

class ArtiqTest(EnvExperiment):

    def build(self):
        self.setattr_device('core')

    @kernel
    def run(self):
        self.foo(1)

    @kernel
    def foo(self, v: 'TInt32'):
        self.core.reset()
        print('foo=', v)
```

The original version would fail with error message
```
test.py:14:1: error: type annotation for argument 'v', ''TInt32'', is not an ARTIQ type
def foo(self, v: 'TInt32'):
^                          
test.py:14:1: note: in kernel function here
def foo(self, v: 'TInt32'):
^                          
test.py:11:9-11:17: note: while inferring a type for an attribute 'foo' of a host object
        self.foo(1)
        ^^^^^^^^
```
while the fixed version would pass successfully run the code.

### Git Logistics

  ```
  
  ```

### Licensing

### DIFF
--- a/artiq/compiler/embedding.py
+++ b/artiq/compiler/embedding.py
@@ -949,6 +949,27 @@ class Stitcher:
         if annot is None:
             annot = builtins.TNone()
 
+        if isinstance(function, SpecializedFunction):
+            host_function = function.host_function
+        else:
+            host_function = function
+
+        embedded_function = host_function.artiq_embedded.function
+
+        if isinstance(embedded_function, str):
+            embedded_function = host_function
+
+        if isinstance(annot, str):
+            if annot not in embedded_function.__globals__:
+                diag = diagnostic.Diagnostic(
+                        "error",
+                        "type annotation for {kind}, {annot}, is not defined",
+                        {"kind": kind, "annot": repr(annot)},
+                        self._function_loc(function),
+                        notes=self._call_site_note(call_loc, fn_kind))
+                self.engine.process(diag)
+            annot = embedded_function.__globals__[annot]
+
         if not isinstance(annot, types.Type):
             diag = diagnostic.Diagnostic("error",
                 "type annotation for {kind}, '{annot}', is not an ARTIQ type",


dnadlinger:
(Incidentally, this is the issue I encountered when trying 3.9 a while back.)

lriesebos:
@sbourdeauducq do you think this could be backported to ARTIQ 6?

sbourdeauducq:
Why not use 7 if you need this?
You can also try ``from __future__ import annotations`` which might solve your initial problem (I haven't tried if the legacy compiler accepts it).

lriesebos:
For our running systems, we stick with the latest release version of ARTIQ, currently 6.

The issue fixed by this and the other PR is actually caused by using `from __future__ import annotations`. See https://git.m-labs.hk/M-Labs/nac3-spec/issues/19 for the original issue. I guess it could be considered a bugfix.

If you do not think this should be backported, then I will solve it by removing the future annotations import in my file and use string type annotations where postponed evaluation of annotations is needed. The ARTIQ types obviously do not need postponed evaluation.

sbourdeauducq:
FYI even Python 3.11 does not implement the new behavior:
```
Python 3.11.0a4 (main, Jan 13 2022, 19:39:06) [GCC 10.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> class X:
...   x: int
... 
>>> X.__annotations__
{'x': <class 'int'>}
>>> def foo(x: int):
...   pass
... 
>>> foo.__annotations__
{'x': <class 'int'>}
```

NAC3 does depend on use of ``from __future__ import annotations`` for postponed evaluation on the other hand; string annotations are not going to be supported.

Have you tested the release-6 compiler with those changes and are you able to tell that this would not cause regressions?

PR #1716, opened 2021-07-08
GUI: create new datasets

### LINKED ISSUE #1707: Dashboard: create dataset
# ARTIQ Feature Request

Currently it's easy to edit / delete a dataset from the dashboard but not to create a new one. Is there a reason for this, or just an omission?

Option to create a new dataset from the dashboard.

SidaChen1999:
# ARTIQ Pull Request

## Description of Changes
Adding new datasets function in artiq_dashboard. The users can type in the name and value of the datasets in corresponding boxes. The data type will be shown next to the value box in real-time. It will show "N/A" if the input cannot be recognized and if the user still hits enter it will not proceed but show a log error. The decoder is `sipyco::pyon`
Warn!: string should be input like `"something"` with a double-quote.

image

### Related Issue

Closes #1707 

## Type of Changes

|   | Type |

### All Pull Requests

- [x] Close/update issues.

### Code Changes

### Git Logistics

  ```
  
  ```

### Licensing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,7 +16,7 @@ __pycache__/
 /.coverage
 /artiq/test/lit/*/Output/
 /artiq/binaries
-/artiq/firmware/target/
+/artiq/firmware/**/target/
 /misoc_*/
 /artiq_*/
 
--- a/.gitignore
+++ b/.gitignore
@@ -16,7 +16,7 @@ __pycache__/
 /.coverage
 /artiq/test/lit/*/Output/
 /artiq/binaries
-/artiq/firmware/**/target/
+/artiq/firmware/target/
 /misoc_*/
 /artiq_*/
 
--- a/artiq/dashboard/datasets.py
+++ b/artiq/dashboard/datasets.py
@@ -134,14 +134,15 @@ class Creator(QtWidgets.QDialog):
     def dtype(self):
         txt = self.value_widget.text()
         try:
-            self.data_type.setText(type(pyon.decode(txt)).__name__)
-            self.value_widget.setStyleSheet("")
-            self.ok.setEnabled(True)
+            pyon.decode(txt)
         except:
             pixmap = self.style().standardPixmap(
                 QtWidgets.QStyle.SP_MessageBoxWarning)
             self.data_type.setPixmap(pixmap)
             self.ok.setEnabled(False)
+        else:
+            self.data_type.setText(type(pyon.decode(txt)).__name__)
+            self.ok.setEnabled(True)
 
 
 class Model(DictSyncTreeSepModel):
--- a/artiq/dashboard/datasets.py
+++ b/artiq/dashboard/datasets.py
@@ -135,13 +135,13 @@ class Creator(QtWidgets.QDialog):
         txt = self.value_widget.text()
         try:
             self.data_type.setText(type(pyon.decode(txt)).__name__)
-        except:
-            self.value_widget.setStyleSheet("background-color: rgb(200, 0, 0)")
-            self.data_type.setText('N/A')
-            self.ok.setEnabled(False)
-        else:
             self.value_widget.setStyleSheet("")
             self.ok.setEnabled(True)
+        except:
+            pixmap = self.style().standardPixmap(
+                QtWidgets.QStyle.SP_MessageBoxWarning)
+            self.data_type.setPixmap(pixmap)
+            self.ok.setEnabled(False)
 
 
 class Model(DictSyncTreeSepModel):
--- a/artiq/dashboard/datasets.py
+++ b/artiq/dashboard/datasets.py
@@ -5,7 +5,7 @@ import numpy as np
 from PyQt5 import QtCore, QtWidgets
 from sipyco import pyon
 
-from artiq.tools import short_format
+from artiq.tools import short_format, exc_to_warning
 from artiq.gui.tools import LayoutWidget, QRecursiveFilterProxyModel
 from artiq.gui.models import DictSyncTreeSepModel
 from artiq.gui.scientific_spinbox import ScientificSpinBox
@@ -82,6 +82,7 @@ class StringEditor(Editor):
     def get_edit_widget_value(self):
         return self.edit_widget.text()
 
+
 class Creator(QtWidgets.QDialog):
     def __init__(self, parent, dataset_ctl):
         QtWidgets.QDialog.__init__(self, parent=parent)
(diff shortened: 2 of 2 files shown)


dnadlinger:
The "real-time" data type display is a neat touch; I like it.

hartytp:
❤️

artiq/dashboard/datasets.py
```diff
@@ -81,6 +82,63 @@ def get_edit_widget(self, initial_value):
...
+        self.dataset_ctl = dataset_ctl
+
+        self.setWindowTitle("Create dataset")
+        grid = QtWidgets.QGridLayout()
+        self.setLayout(grid)
+
+        grid.addWidget(QtWidgets.QLabel("Name:"), 0, 0)
+        grid.addWidget(self.get_key_widget(), 0, 1)
+
+        grid.addWidget(QtWidgets.QLabel("Value:"), 1, 0)
```
dnadlinger, line 97: Maybe add `" (PYON):"` to the label, or at least as a tool tip?

artiq/dashboard/datasets.py
```diff
@@ -81,6 +82,63 @@ def get_edit_widget(self, initial_value):
...
+
+    def accept(self):
+        try:
+            key = self.get_create_widget_text()
+            value = self.get_create_widget_value()
+            asyncio.ensure_future(self.dataset_ctl.set(key, pyon.decode(value)))
+            QtWidgets.QDialog.accept(self)
+        except:
+            logger.error("Cannot create dataset: "
+                                 "type is not supported",)
```
dnadlinger, line 118: Indentation weird. Also, isn't "type not supported" potentially a bit misleading for what is most likely a user typo?

artiq/dashboard/datasets.py
```diff
@@ -81,6 +82,63 @@ def get_edit_widget(self, initial_value):
...
+            return
+
+    def get_key_widget(self):
+        self.key_widget = QtWidgets.QLineEdit()
```
dnadlinger, line 122: I find a `get_*` method that overwrites a member to be rather confusing. I'd either just use `self.key_widget` throughout, or make the `get_*` methods lazily create the line edits (i.e. something like `if self.key_widget is None: … return self.key_widget`).

dnadlinger (changes requested):
Perhaps the most intuitive UI choice would be to highlight an invalid value (or non-allowable key) somehow, and disallow form submission then. A red border on the line edit, perhaps. This would make the UI rather self-documenting.

The UI should also have a way to specify whether the dataset is persistent (probably just a tick box).

I've left a few inline code comments as well.

dhslichter:
this is fantastic @SidaChen1999!  I agree with @dnadlinger's comments, as well as his suggestion for a "persist" tickbox.

sbourdeauducq:
> It will show "N/A" if the input cannot be recognized and if the user still hits enter it will not proceed but show a log error.

What about greying out the OK button and ignoring Enter while the input doesn't parse?

SidaChen1999:
Thanks for all the reviews. Now the window will look like the first picture when it is first opened. And when something invalid is input, it will show as the second one.
image
image

.gitignore
```diff
@@ -16,7 +16,7 @@ __pycache__/
 /.coverage
 /artiq/test/lit/*/Output/
 /artiq/binaries
-/artiq/firmware/target/
+/artiq/firmware/**/target/
```
sbourdeauducq, line 19: This change does not belong in this PR.

artiq/dashboard/datasets.py
```diff
@@ -81,6 +82,89 @@ def get_edit_widget(self, initial_value):
...
+        grid.addWidget(self.buttons, 4, 0, 1, 3)
+        self.buttons.accepted.connect(self.accept)
+        self.buttons.rejected.connect(self.reject)
+
+    def accept(self):
+        try:
+            key = self.get_create_widget_text()
+            value = self.get_create_widget_value()
+            persist = self.get_persist_widget_value()
+            asyncio.ensure_future(self.dataset_ctl.set(key, pyon.decode(value), persist))
```
sbourdeauducq, line 124: Use this to wrap the asyncio task and log errors immediately: https://github.com/m-labs/artiq/blob/master/artiq/tools.py#L107
Otherwise the asyncio task exceptions are only retrieved when the GUI exits and error messages printed then. This makes debugging tedious.

artiq/dashboard/datasets.py
```diff
@@ -81,6 +82,89 @@ def get_edit_widget(self, initial_value):
...
+        return self.box_widget
+
+    def get_create_widget_text(self):
+        return self.name_widget.text()
+
+    def get_create_widget_value(self):
+        return self.value_widget.text()
+    
+    def get_persist_widget_value(self):
+        return self.box_widget.isChecked()
```
sbourdeauducq, line 152: Those methods can be removed IMO. The ones that create the widgets are also not very well-written.

artiq/dashboard/datasets.py
```diff
@@ -81,6 +82,89 @@ def get_edit_widget(self, initial_value):
...
+        except:
+            self.value_widget.setStyleSheet("border: 1px solid red;"
+                                            "height: 28px;"
+                                            "border-radius: 2px;")
+            self.data_type.setText('N/A')
+            self.ok.setEnabled(False)
+        else:
+            self.value_widget.setStyleSheet("")
+            self.ok.setEnabled(True)
+        return
```
sbourdeauducq, line 167: This ``return`` is not needed, remove.

artiq/dashboard/datasets.py
```diff
@@ -146,6 +235,10 @@ def set_model(self, model):
         self.table_model_filter.setSourceModel(self.table_model)
         self.table.setModel(self.table_model_filter)
 
+    def create_clicked(self):
+        dialog_cls = Creator
+        dialog_cls(self, self.dataset_ctl).open()
```
sbourdeauducq, line 240: The ``dialog_cls`` variable is not needed and should be removed.

artiq/dashboard/datasets.py
```diff
@@ -81,6 +82,89 @@ def get_edit_widget(self, initial_value):
...
+        self.buttons.rejected.connect(self.reject)
+
+    def accept(self):
+        try:
+            key = self.get_create_widget_text()
+            value = self.get_create_widget_value()
+            persist = self.get_persist_widget_value()
+            asyncio.ensure_future(self.dataset_ctl.set(key, pyon.decode(value), persist))
+            QtWidgets.QDialog.accept(self)
+        except:
```
sbourdeauducq, line 126: Remove.

artiq/dashboard/datasets.py
```diff
@@ -81,6 +82,89 @@ def get_edit_widget(self, initial_value):
...
+            logger.error("Cannot create dataset: "
+                         "input is not valid",)
+            return
```
sbourdeauducq, line 129: Remove.

artiq/dashboard/datasets.py
```diff
@@ -82,6 +83,69 @@ def get_edit_widget_value(self):
...
+            key, pyon.decode(value), persist)))
+        QtWidgets.QDialog.accept(self)
+
+    def dtype(self):
+        txt = self.value_widget.text()
+        try:
+            self.data_type.setText(type(pyon.decode(txt)).__name__)
+        except:
+            self.value_widget.setStyleSheet("border: 1px solid red;"
+                                            "height: 28px;"
```
dnadlinger, line 140: Why the explicit `height`? Is this compatible across OSes/UI toolkits?

SidaChen1999: I tested on KDE and XFce, style sheet of `border: 1px solid red` will change the height differently from the default. I fixed the `height` to be the same in KDE but it is indeed different from XFce. So I'll look for another solution.

SidaChen1999: I also tested on GNOME. Changing the border properties of the style sheet will also affect padding and some other properties that make it inconsistent and unaesthetic across OSes. Therefore, I propose to change the background color of the line edit box instead as shown below
image

sbourdeauducq: I'm not sure if a hardcoded color is a good idea, since the user-defined theme colors may change. One obvious problem: You are using the dark theme, but most people use the default with black text and it will be hard to read on a red background.
What about displaying a warning icon at the right of the text box (possibly inside it) when the input is incorrect? Or instead of the N/A text?

SidaChen1999: I changed the `N/A` text into `MessageBoxWarning`. The icon and its size will change according to the OS.
image

artiq/dashboard/datasets.py
```diff
@@ -82,6 +83,67 @@ def get_edit_widget_value(self):
...
+        persist = self.box_widget.isChecked()
+        asyncio.ensure_future(exc_to_warning(self.dataset_ctl.set(
+            key, pyon.decode(value), persist)))
+        QtWidgets.QDialog.accept(self)
+
+    def dtype(self):
+        txt = self.value_widget.text()
+        try:
+            self.data_type.setText(type(pyon.decode(txt)).__name__)
+            self.value_widget.setStyleSheet("")
```
sbourdeauducq, line 138: remove

artiq/dashboard/datasets.py
```diff
@@ -82,6 +83,67 @@ def get_edit_widget_value(self):
...
+        value = self.value_widget.text()
+        persist = self.box_widget.isChecked()
+        asyncio.ensure_future(exc_to_warning(self.dataset_ctl.set(
+            key, pyon.decode(value), persist)))
+        QtWidgets.QDialog.accept(self)
+
+    def dtype(self):
+        txt = self.value_widget.text()
+        try:
+            self.data_type.setText(type(pyon.decode(txt)).__name__)
```
sbourdeauducq, line 137: limit the try/except to ``pyon.decode``

SidaChen1999: I think it is already limited by the `textChanged` signal. Shouldn't we validate the input as soon as new input is detected? I tested on the old pc and it won't take too much resource.

sbourdeauducq: What I meant is right now you are catching exceptions raised by e.g. ``setText`` or ``setEnabled``. This has nothing to do with resource consumption.

artiq/dashboard/datasets.py
```diff
@@ -134,14 +134,15 @@ def accept(self):
     def dtype(self):
         txt = self.value_widget.text()
         try:
-            self.data_type.setText(type(pyon.decode(txt)).__name__)
-            self.value_widget.setStyleSheet("")
-            self.ok.setEnabled(True)
+            pyon.decode(txt)
```
sbourdeauducq, line 137: Just store the result and decode once. You don't need to do anything special regarding try/else:
```python
>>> try:
...    x = 2
... except:
...    pass
... else:
...    print(x)
... 
2
```

Please do such research on the Python language next time.

PR #1718, opened 2021-07-12
applets: add warning message on plot for plot_hist

### LINKED ISSUE #1702: applets.plot_hist: raise Exception on incorrect data lengths
Current behavior: if this test fails the applet silently ignores all data

https://github.com/m-labs/artiq/blob/094a346974642150f9df4d89197207d711e12088/artiq/applets/plot_hist.py#L26

Desired behavior: a clear exception (`ValueError`?) so users have a clue as to why their plot isn't showing anything.

SidaChen1999:
# ARTIQ Pull Request

## Description of Changes
Display a warning message on the plot window if the length of dataset X is not 1 longer than that of Y after a timeout of 0.5s. However, if the change in the dataset is less than the timeout, it will not display anything unless we change the dataset again. The display is as shown below:
image

### Related Issue
Closes #1702 

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

### All Pull Requests

- [x] Close/update issues.

### Code Changes

### Git Logistics

  ```
  
  ```

### Licensing

### DIFF
--- a/artiq/applets/plot_hist.py
+++ b/artiq/applets/plot_hist.py
@@ -11,7 +11,6 @@ class HistogramPlot(pyqtgraph.PlotWidget):
     def __init__(self, args):
         pyqtgraph.PlotWidget.__init__(self)
         self.args = args
-        self.flag = True
         self.timer = QTimer()
         self.timer.setSingleShot(True)
         self.timer.timeout.connect(self.length_warning)
@@ -29,21 +28,18 @@ class HistogramPlot(pyqtgraph.PlotWidget):
             x = list(range(len(y)+1))
 
         if len(y) and len(x) == len(y) + 1:
-            self.flag = True
             self.timer.stop()
             self.clear()
             self.plot(x, y, stepMode=True, fillLevel=0,
                       brush=(0, 0, 255, 150))
             self.setTitle(title)
         else:
-            if self.flag:
-                self.flag = False
-                self.clear()
-                self.timer.start(1000)
+            self.timer.start(1000)
 
     def length_warning(self):
-        text = '''⚠️ dataset lengths mismatch:\n
-                    BIN_BOUNDARIES should be one more than COUNTS'''
+        self.clear()
+        text = "⚠️ dataset lengths mismatch:\n"\
+               "Bin boundaries should be one more than Y values"
         self.addItem(pyqtgraph.TextItem(text))
 
 
--- a/artiq/applets/plot_hist.py
+++ b/artiq/applets/plot_hist.py
@@ -12,6 +12,9 @@ class HistogramPlot(pyqtgraph.PlotWidget):
         self.args = args
 
     def data_changed(self, data, mods, title):
+        if not hasattr(self, 'flag'):
+            self.flag = True
+
         try:
             y = data[self.args.y][1]
             if self.args.x is None:
@@ -24,10 +27,19 @@ class HistogramPlot(pyqtgraph.PlotWidget):
             x = list(range(len(y)+1))
 
         if len(y) and len(x) == len(y) + 1:
+            self.flag = True
             self.clear()
             self.plot(x, y, stepMode=True, fillLevel=0,
                       brush=(0, 0, 255, 150))
             self.setTitle(title)
+        else:
+            if self.flag:
+                self.flag_time = pyqtgraph.ptime.time()
+            self.flag = False
+            if pyqtgraph.ptime.time() - self.flag_time > 0.5:
+                self.clear()
+                text = '⚠️ The length of dataset X is not Y+1'
+                self.addItem(pyqtgraph.TextItem(text))
 
 
 def main():
--- a/artiq/applets/plot_hist.py
+++ b/artiq/applets/plot_hist.py
@@ -39,7 +39,7 @@ class HistogramPlot(pyqtgraph.PlotWidget):
     def length_warning(self):
         self.clear()
         text = "⚠️ dataset lengths mismatch:\n"\
-               "Bin boundaries should be one more than Y values"
+            "There should be one more bin boundaries than there are Y values"
         self.addItem(pyqtgraph.TextItem(text))
 
 
--- a/artiq/applets/plot_hist.py
+++ b/artiq/applets/plot_hist.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
-import PyQt5  # make sure pyqtgraph imports Qt5
+import PyQt5    # make sure pyqtgraph imports Qt5
+from PyQt5.QtCore import QTimer
 import pyqtgraph
 
 from artiq.applets.simple import TitleApplet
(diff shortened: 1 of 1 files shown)


artiq/applets/plot_hist.py
```diff
@@ -12,6 +12,9 @@ def __init__(self, args):
         self.args = args
 
     def data_changed(self, data, mods, title):
+        if not hasattr(self, 'flag'):
+            self.flag = True
+
```
sbourdeauducq, line 17: Set ``flag`` to ``True`` in ``__init__`` instead.

artiq/applets/plot_hist.py
```diff
@@ -24,10 +27,19 @@ def data_changed(self, data, mods, title):
...
             self.plot(x, y, stepMode=True, fillLevel=0,
                       brush=(0, 0, 255, 150))
             self.setTitle(title)
+        else:
+            if self.flag:
+                self.flag_time = pyqtgraph.ptime.time()
+            self.flag = False
+            if pyqtgraph.ptime.time() - self.flag_time > 0.5:
+                self.clear()
+                text = '⚠️ The length of dataset X is not Y+1'
```
sbourdeauducq, line 41: This message is a bit confusing.

sbourdeauducq:
AFAICT this code does not work if you initialize the datasets with incorrect lengths and then do not touch them afterwards. The warning message is not displayed at all. You probably need to use QTimer and another event.

artiq/applets/plot_hist.py
```diff
@@ -10,24 +11,38 @@ class HistogramPlot(pyqtgraph.PlotWidget):
...
+
+    def warning(self):
+        if len(self.y) and len(self.x) == len(self.y) + 1:
+            return
+        else:
+            self.timer.stop()
+            self.clear()
+            text = '''⚠️ dataset lengths mismatch:\n
+                      BIN_BOUNDARIES should be one more than COUNTS'''
+            self.addItem(pyqtgraph.TextItem(text))
```
sbourdeauducq, line 45: How is the text removed?

artiq/applets/plot_hist.py
```diff
@@ -10,24 +11,38 @@ class HistogramPlot(pyqtgraph.PlotWidget):
...
+        if len(self.y) and len(self.x) == len(self.y) + 1:
             self.clear()
-            self.plot(x, y, stepMode=True, fillLevel=0,
+            self.plot(self.x, self.y, stepMode=True, fillLevel=0,
                       brush=(0, 0, 255, 150))
             self.setTitle(title)
+        else:
+            self.timer.start(1000)
+
+    def warning(self):
```
sbourdeauducq, line 37: length_warning

artiq/applets/plot_hist.py
```diff
@@ -10,24 +11,38 @@ class HistogramPlot(pyqtgraph.PlotWidget):
...
+        if len(self.y) and len(self.x) == len(self.y) + 1:
+            return
```
sbourdeauducq, line 39: What about stopping the timer in ``data_changed`` instead, so you don't have to store the data and duplicate this test?

artiq/applets/plot_hist.py
SidaChen1999: Ahead of `self.plot` there is a `self.clear` that does the job

artiq/applets/plot_hist.py
```diff
@@ -10,24 +11,38 @@ class HistogramPlot(pyqtgraph.PlotWidget):
...
+        else:
+            self.timer.stop()
+            self.clear()
+            text = '''⚠️ dataset lengths mismatch:\n
+                      BIN_BOUNDARIES should be one more than COUNTS'''
```
sbourdeauducq, line 44: They are called differently in the command line arguments.

sbourdeauducq:
After thinking about it, I would solve this one with a timer that continuously ticks every 200ms, and a finite state machine that responds to "timer tick", "mismatch appears" and "mismatch disappears" events.
This way we can also handle the text in only the timer tick callback and avoid making a lot of pyqtgraph calls when the data is modified.

artiq/applets/plot_hist.py
SidaChen1999: Should I change the name to `Y values` and `Bin boundaries`?

sbourdeauducq: Would be ok.

dnadlinger:
> After thinking about it, I would solve this one …

I'm sure my two cents are entirely unnecessary here (just saw this scroll by), but I would have solved that with a one-shot timer set in the data callback that adds the text, where it is also cancelled (and text removed) if the lengths match. Just make sure the same function is used for initial rendering too.

artiq/applets/plot_hist.py
```diff
@@ -10,6 +11,10 @@ class HistogramPlot(pyqtgraph.PlotWidget):
     def __init__(self, args):
         pyqtgraph.PlotWidget.__init__(self)
         self.args = args
+        self.flag = True
```
sbourdeauducq, line 14: ``flag`` -> ``timer_triggered`` (with the opposite True/False values)?
Though:
1. This could be replaced with ``timer.isActive()``
2. Is it necessary to have this check at all?

artiq/applets/plot_hist.py
```diff
@@ -24,10 +29,22 @@ def data_changed(self, data, mods, title):
...
+            self.flag = True
+            self.timer.stop()
             self.clear()
             self.plot(x, y, stepMode=True, fillLevel=0,
                       brush=(0, 0, 255, 150))
             self.setTitle(title)
+        else:
+            if self.flag:
+                self.flag = False
+                self.clear()
```
sbourdeauducq, line 41: Why clear here?

SidaChen1999: If there is a successful plot already, simply add the `TextItem` without clear will make the text appear below the plot. Is that what we want?
image

SidaChen1999: What's more important, if another wrong dataset is submitted, the warning message will be rendered again and looks not good.

sbourdeauducq: What about clearing in ``length_warning`` then?

SidaChen1999: ok

artiq/applets/plot_hist.py
```diff
@@ -24,10 +28,19 @@ def data_changed(self, data, mods, title):
...
             self.plot(x, y, stepMode=True, fillLevel=0,
                       brush=(0, 0, 255, 150))
             self.setTitle(title)
+        else:
+            self.timer.start(1000)
+
+    def length_warning(self):
+        self.clear()
+        text = "⚠️ dataset lengths mismatch:\n"\
+               "Bin boundaries should be one more than Y values"
```
sbourdeauducq, line 42: I think the correct grammar would be:
"There should be one more bin boundary than there are Y values."

PR #1720, opened 2021-07-16
applets: add warning message on plot for `plot_xy` and `plot_xy_hist`

### LINKED ISSUE #1702: applets.plot_hist: raise Exception on incorrect data lengths

SidaChen1999:
# ARTIQ Pull Request

## Description of Changes
Display a warning message on the plot window if the datasets do not satisfy some conditions after a timeout of 1s. 
For the `plot_xy` applet, the length of X values, error bars, fit values should have the same length as Y values. Because error bars and fit values are optional, the xy plot will still be displayed if they are wrong. The display is as shown below:
image
For the `plot_xy_hist` applet, the length of bin boundaries should be the same as the first dimension of histogram counts, and the length of point abscissas should be the same as the second dimension of histogram counts. Either failure will cause the xy plot to fail to display. The display is as shown below:
image
Potential bug fix: previously the histogram plot will crash in all my testing cases. Now, it will show `click on the data point at left to see histogram` if the xy plot is shown.

### Related Issue
Closes #1702

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

### All Pull Requests

- [x] Close/update issues.

### Code Changes

### Git Logistics

  ```
  
  ```

### Licensing

### DIFF
--- a/artiq/applets/plot_xy.py
+++ b/artiq/applets/plot_xy.py
@@ -2,6 +2,7 @@
 
 import numpy as np
 import PyQt5  # make sure pyqtgraph imports Qt5
+from PyQt5.QtCore import QTimer
 import pyqtgraph
 
 from artiq.applets.simple import TitleApplet
@@ -11,6 +12,12 @@ class XYPlot(pyqtgraph.PlotWidget):
     def __init__(self, args):
         pyqtgraph.PlotWidget.__init__(self)
         self.args = args
+        self.timer = QTimer()
+        self.timer.setSingleShot(True)
+        self.timer.timeout.connect(self.length_warning)
+        self.mismatch = {'X values': False,
+                         'Error bars': False,
+                         'Fit values': False}
 
     def data_changed(self, data, mods, title):
         try:
@@ -24,31 +31,56 @@ class XYPlot(pyqtgraph.PlotWidget):
         fit = data.get(self.args.fit, (False, None))[1]
 
         if not len(y) or len(y) != len(x):
-            return
+            self.mismatch['X values'] = True
+            self.timer.start(1000)
+        else:
+            self.mismatch['X values'] = False
+            if not sum(self.mismatch.values()):
+                self.timer.stop()
         if error is not None and hasattr(error, "__len__"):
             if not len(error):
                 error = None
             elif len(error) != len(y):
-                return
+                self.mismatch['Error bars'] = True
+                self.timer.start(1000)
+            else:
+                self.mismatch['Error bars'] = False
+                if not sum(self.mismatch.values()):
+                    self.timer.stop()
         if fit is not None:
             if not len(fit):
                 fit = None
             elif len(fit) != len(y):
-                return
+                self.mismatch['Fit values'] = True
+                self.timer.start(1000)
+            else:
+                self.mismatch['Fit values'] = False
+                if not sum(self.mismatch.values()):
+                    self.timer.stop()
 
         self.clear()
+        if self.mismatch['X values']:
+            return
         self.plot(x, y, pen=None, symbol="x")
         self.setTitle(title)
-        if error is not None:
+        if error is not None and not self.mismatch['Error bars']:
             # See https://github.com/pyqtgraph/pyqtgraph/issues/211
             if hasattr(error, "__len__") and not isinstance(error, np.ndarray):
                 error = np.array(error)
             errbars = pyqtgraph.ErrorBarItem(
                 x=np.array(x), y=np.array(y), height=error)
             self.addItem(errbars)
-        if fit is not None:
+        if fit is not None and not self.mismatch['Fit values']:
             xi = np.argsort(x)
             self.plot(x[xi], fit[xi])
+
+    def length_warning(self):
+        text = "⚠️ dataset lengths mismatch:\n"
+        for key in self.mismatch:
+            if self.mismatch[key]:
+                text += key + ', '
+        text += "should have the same length as Y values"
+        self.addItem(pyqtgraph.TextItem(text))
 
 
 def main():
--- a/artiq/applets/plot_xy_hist.py
+++ b/artiq/applets/plot_xy_hist.py
@@ -2,6 +2,7 @@
 
 import numpy as np
 from PyQt5 import QtWidgets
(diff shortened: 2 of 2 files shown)


artiq/applets/plot_xy.py
```diff
@@ -24,32 +31,57 @@ def data_changed(self, data, mods, title):
         fit = data.get(self.args.fit, (False, None))[1]
 
         if not len(y) or len(y) != len(x):
-            return
+            self.mismatch['X values'] = True
+            self.timer.start(1000)
+        else:
+            self.mismatch['X values'] = False
+            if not sum(self.mismatch.values()):
```
sbourdeauducq, line 38: any()

artiq/applets/plot_xy.py
```diff
@@ -24,32 +31,57 @@ def data_changed(self, data, mods, title):
...
             if not len(error):
                 error = None
             elif len(error) != len(y):
-                return
+                self.mismatch['Error bars'] = True
+                self.timer.start(1000)
+            else:
+                self.mismatch['Error bars'] = False
+                if not sum(self.mismatch.values()):
+                    self.timer.stop()
```
sbourdeauducq, line 49: Move below and do not duplicate.

artiq/applets/plot_xy.py
```diff
@@ -24,32 +31,57 @@ def data_changed(self, data, mods, title):
...
+                self.mismatch['Fit values'] = True
+                self.timer.start(1000)
+            else:
+                self.mismatch['Fit values'] = False
+                if not sum(self.mismatch.values()):
+                    self.timer.stop()
 
         self.clear()
+        if self.mismatch['X values']:
+            return
```
sbourdeauducq, line 63: This deviates from the original behavior AFAICT. Why?

artiq/applets/plot_xy.py
```diff
@@ -24,32 +31,57 @@ def data_changed(self, data, mods, title):
...
         self.plot(x, y, pen=None, symbol="x")
         self.setTitle(title)
-        if error is not None:
+        if error is not None and not self.mismatch['Error bars']:
```
sbourdeauducq, line 66: Why?

artiq/applets/plot_xy.py
```diff
@@ -24,32 +31,57 @@ def data_changed(self, data, mods, title):
...
-        if fit is not None:
+        if fit is not None and not self.mismatch['Fit values']:
             xi = np.argsort(x)
             self.plot(x[xi], fit[xi])
 
+    def length_warning(self):
+        text = "⚠️ dataset lengths mismatch:\n"
+        for key in self.mismatch:
+            if self.mismatch[key]:
+                text += key + ', '
```
sbourdeauducq, line 81: ``", ".join()``
The output is otherwise also a bit strange with only one mismatch.

artiq/applets/plot_xy.py
```diff
@@ -24,32 +31,57 @@ def data_changed(self, data, mods, title):
         fit = data.get(self.args.fit, (False, None))[1]
 
         if not len(y) or len(y) != len(x):
-            return
+            self.mismatch['X values'] = True
+            self.timer.start(1000)
```
sbourdeauducq, line 35: This restarts the timer, so regular data input that constantly produces mismatches will not trigger the warning.
I think your previous PR also has this bug.

artiq/applets/plot_xy.py
SidaChen1999: I think `not all()` is what you mean. The timer will stop only when all the mismatches are `False`

artiq/applets/plot_xy.py
SidaChen1999: If the length of X and Y does not match, `self.plot` at the next line will cause the applet to crash and has to be reopened.

artiq/applets/plot_xy.py
SidaChen1999: The same as the checking ahead of `self.plot`, the applet window will crash if the length of error bars is mismatched with the length of Y values. And also the check of fit values.

artiq/applets/plot_xy.py
SidaChen1999: If use any() I think it would be `not any(list(self.mismatch.values()))` and it is not simpler than `not sum(self.mismatch.values())`

pca006132: Just `not any(self.mismatch.values())`. You don't need `list`.

artiq/applets/plot_xy.py
sbourdeauducq: That would be a separate bug, can you fix it first in a separate PR?

artiq/applets/plot_xy.py
SidaChen1999: > flag -> timer_active (with the opposite True/False values)?
Though:
> This could be replaced with timer.isActive()
Is it necessary to have this check at all?

In the last PR, I remove the flag checking. But the purpose of it is to make sure that the warning is shown right after the timeout of the happening of the first failure.

sbourdeauducq: With your previous PR, if you regularly send data that produces a mismatch every time, it will not trigger the warning AFAICT (but it should). This is due to the timer being constantly restarted before it can fire.

artiq/applets/plot_xy.py
SidaChen1999: My fault, the previous explanation is incorrect. In the original file, it will not show anything if the `error bars` or `fit values` do not have the same length as `Y`. But I would like to change it. I want to plot whatever is possible and show the error message below. Should I do that or stick with the original behavior.

sbourdeauducq: Probably easier to keep the original behavior.

sbourdeauducq:
This PR can be closed, right?

PR #1612, opened 2021-02-12
added max_kernel_size argument to core driver

lriesebos:
# ARTIQ Pull Request

## Description of Changes

Add an option to the core driver to limit the maximum (static) kernel size before it is loaded to the core device.

I do not know if there is any interest in such a feature, feel free to close this PR if there is not.

### Related Issue

A way to prevent loading kernels that are too large while waiting for #544.

## Type of Changes

|   | Type |
| ------------- | ------------- |
| ✓  | :sparkles: New feature |

### All Pull Requests

- [x] Use correct spelling and grammar.
- [ ] Update [RELEASE_NOTES.rst](../RELEASE_NOTES.rst) if there are noteworthy changes, especially if there are changes to existing APIs.
- [x] Close/update issues.
- [x] Check the copyright situation of your changes and sign off your patches (`git commit --signoff`, see [copyright](../CONTRIBUTING.rst#copyright-and-sign-off)).

### Code Changes

- [x] Run `flake8` to check code style (follow PEP-8 style). `flake8` has issues with parsing Migen/gateware code, ignore as necessary.
- [x] Test your changes or have someone test them. Mention what was tested and how.
- [x] Add and check docstrings and comments
- [x] Check, test, and update the [unittests in /artiq/test/](../artiq/test/) or [gateware simulations in /artiq/gateware/test](../artiq/gateware/test)

### Documentation Changes

- [x] Check, test, and update the documentation in [doc/](../doc/). Build documentation (`cd doc/manual/; make html`) to ensure no errors.

### Git Logistics

- [x] Split your contribution into logically separate changes (`git rebase --interactive`). Merge/squash/fixup commits that just fix or amend previous commits. Remove unintended changes & cleanup. See [tutorial](https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase).
- [x] Write short & meaningful commit messages. Review each commit for messages (`git show`). Format:
  ```
  topic: description. < 50 characters total.
  
  Longer description. < 70 characters per line
  ```

### Licensing

See [copyright & licensing for more info](https://github.com/m-labs/artiq/blob/master/CONTRIBUTING.rst#copyright-and-sign-off).
ARTIQ files that do not contain a license header are copyrighted by M-Labs Limited and are licensed under LGPLv3+.

PR #1764, opened 2021-10-08
Stack overflow protection

### LINKED ISSUE #544: Trivial memory protection unit (stack overflow protection)
On both kernel and comms CPUs, we mostly use memory-safe code (or will use, once Rust on comms CPU ships), so it is largely unnecessary to have a full-blown MPU. However, memory-safe code relies on detection of stack overflows, and currently that would not happen--indeed, overflows on both CPUs would silently corrupt data and crash some time later.

I propose a trivial MPU to alleviate this problem. It would work as follows:
- There are _n_ registers, each containing a base address for a region 4k long and aligned on a 4k boundary.
- When the region is accessed, an exception is raised.

_n_ would be configurable; on kernel CPU we need one, on comms CPU we need the same _n_ as number of threads.

It is only necessary to trap on a region 4k long because LLVM and Rust support stack probing, intended to be safe in presence of guard pages; when allocating a stack frame, LLVM will write a dummy value to stack every 4k bytes.

occheung:
# ARTIQ Pull Request

## Description of Changes
This patch implements stack overflow protection using guard page, with the aid of physical memory protection (PMP) units from RISC-V.

## Linker scripts (Except bootloader)
A symbol `_sstack_guard` is added to each of the linker script (except bootloader). It specifies the start of the stack guard.

Both stacks and guard pages are 4k aligned. Guard pages of these stacks are always active. PMP region 2 implements these guard pages.

## Thread stacks
When spawning a new thread, `Io` now inflates it stack size by 4k to fit the guard page of the thread. Stacks requested by libfringe are 4k aligned, so the guard page will be 4k aligned as well.

The guard page of a thread stack is active only if the context is switched to the thread. Both PMP region 0 & 1 implement these guard pages.

### Satman/Kernel
A PMP region will be enabled to initialize their stack guard during their startup.

### Runtime
Same as satman/kernel. In addition, runtime will then transfer control from machine privilege to user privilege. It will remain in user privilege unless PMP regions need to be modified.

## Context switch
An environment call will be invoked during context switching (see `arch::swap()` implemented in libfringe). The exception handler then adjusts the PMP regions in machine mode. Finally, it resumes context switching with user privilege.

### Main stack guard
Adding the following code after initializing the main stack guard will trigger a CPU load fault.
```Rust
        extern {
            static mut _sstack_guard: u8;
        }

        unsafe {
            println!("stack border: {:?}", *((&_sstack_guard as *const u8).offset(4095)));
        }
```
Important serial output:
```
panic at runtime/main.rs:338:13: exception LoadFault at PC 0x4000301c, trap value 0x4015cfff
```

### Thread stack guard
Spawning a task that initializes a large array will cause an access fault. For example,
```rust
    io.spawn(4096, |_io| {
        let large_array: [u32; 1024] = [0; 1024];
        println!("Array[0]: {}", large_array[0]);
    });
```
It resulted in the following serial output:
```
Trap frame: TrapFrame { ra: 40008214, t0: deaddead, t1: deaddead, t2: 7, t3: 7, t4: 40052ba4, t5: 4004dc00, t6: 0, a0: 4016cce0, a1: 4016cb60, a2: 4016d084, a3: 40181ff4, a4: 2000, a5: 4017fff4, a6: 201c, a7: 40180000 }
@ 0x400211f8
+0000: 00c12423 02010693 00452403 00000513
+0010: 00058613 00000893 808e70ef 00b12623
+0020: 00c10513 00a12823 00812503 00a12a23
+0030: 00812c23 00001637 02010413 00040513
panic at runtime/main.rs:328:13: exception StoreFault at PC 0x400211f8, trap value 0x40180f78
backtrace for software version 7.unknown.beta;nist_clock:
0x4002e63c
0x40004e00
0x4002dba0
halting.
use `artiq_coremgmt config write -s panic_reset 1` to restart instead
```
The PC value corresponds to the following assembly dump.
```
400211b0 <_ZN6fringe9generator39Generator$LT$Input$C$Output$C$Stack$GT$10unsafe_new17generator_wrapper17h7482ebcb48eab63eE>:
    unsafe extern "C" fn generator_wrapper<Input, Output, Stack, F>(env: usize, stack_ptr: StackPointer) -> !
400211b0:	81010113          	addi	sp,sp,-2032
400211b4:	7e112623          	sw	ra,2028(sp)
400211b8:	7e812423          	sw	s0,2024(sp)
400211bc:	7e912223          	sw	s1,2020(sp)
400211c0:	7f212023          	sw	s2,2016(sp)
400211c4:	7d312e23          	sw	s3,2012(sp)
400211c8:	7d412c23          	sw	s4,2008(sp)
400211cc:	7d512a23          	sw	s5,2004(sp)
400211d0:	7d612823          	sw	s6,2000(sp)
400211d4:	7d712623          	sw	s7,1996(sp)
400211d8:	7d812423          	sw	s8,1992(sp)
400211dc:	7d912223          	sw	s9,1988(sp)
400211e0:	7da12023          	sw	s10,1984(sp)
400211e4:	7bb12e23          	sw	s11,1980(sp)
400211e8:	00001637          	lui	a2,0x1
400211ec:	89060613          	addi	a2,a2,-1904 # 890 <.Lline_table_start0+0x658>
400211f0:	40c10133          	sub	sp,sp,a2
400211f4:	00052603          	lw	a2,0(a0)
400211f8:	00c12423          	sw	a2,8(sp)
400211fc:	02010693          	addi	a3,sp,32
40021200:	00452403          	lw	s0,4(a0)
40021204:	00000513          	li	a0,0
40021208:	00058613          	mv	a2,a1
4002120c:	00000893          	li	a7,0
40021210:	808e70ef          	jal	ra,40008218 <_ZN6fringe4arch3imp4swap10trampoline17h07dbcd4ec0d537a5E>
      let yielder = Yielder::new(stack_ptr);
40021214:	00b12623          	sw	a1,12(sp)
40021218:	00c10513          	addi	a0,sp,12
```
So the fault was actually triggered during the initialization of thread in libfringe.

### Kernel stack guard
Added the following code after initializing the stack guard in kernel.
```Rust
    println!("Access guard memory addr: {:p}", (_sstack_guard as usize as *const usize));
    core::ptr::read_volatile((_sstack_guard as *const usize).offset(1023));
```
After a library is loaded to the kernel, it panics with error messages like such.
```
[    38.838184s]  INFO(kernel): Access guard memory addr: 0x45061000
[    38.843520s]  INFO(kernel): panic at ksupport/lib.rs:539:5: Exception(LoadFault) at PC 0x450073dc, trap value 0x45061ffc
```

## Related Issue
[libfringe PR](https://git.m-labs.hk/M-Labs/libfringe/pulls/1)
Closes #544

## Type of Changes

|   | Type |
|   | :bug: Bug fix  |
|   | :hammer: Refactoring  |
|   | :scroll: Docs |

### All Pull Requests

- [x] Close/update issues.

### Code Changes

### Git Logistics

  ```
  
  ```

### Licensing

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -18,11 +18,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1631098960,
-        "narHash": "sha256-6j6G/omHEFAnI2x7UvdrviGDNqiq1Fjd1A58Q6uc4sQ=",
+        "lastModified": 1633625029,
+        "narHash": "sha256-Ia3kwnN9DhcskAIElLSKb4u/OK7nZU/P0TkaikAX790=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "12eb1d16ae3b6cbf0ea83e9228bca8ffd7cfe347",
+        "rev": "781b1f8e3a2194e1e233cd62b4f2193e129a07f7",
         "type": "github"
       },
       "original": {
@@ -45,11 +45,11 @@
     "src-migen": {
       "flake": false,
       "locked": {
-        "lastModified": 1628592379,
-        "narHash": "sha256-TQV6p0+Ri9HVge4qUKOCXe96cmcimNZ3R4sVwNY67DA=",
+        "lastModified": 1633615575,
+        "narHash": "sha256-frh+WVOkEKMpVEMhdE/GZKSj82XnSC8OF8SWNluk/Yg=",
         "owner": "m-labs",
         "repo": "migen",
-        "rev": "27dbf03edd75c32dc1706e2a1316950c3a8d452a",
+        "rev": "6e3f8e565704b4293174aedfb15b3470d233f528",
         "type": "github"
       },
       "original": {
@@ -61,11 +61,11 @@
     "src-misoc": {
       "flake": false,
       "locked": {
-        "lastModified": 1631240866,
-        "narHash": "sha256-dal999XLFvS8Ol1hZnQjx7q/UfAXkzSMhAWcZKCDPx4=",
+        "lastModified": 1633679103,
+        "narHash": "sha256-1BPWfwetUZZCCZ+ldNKJvNm7Gls1AkmGYSZapNuM65s=",
         "ref": "master",
-        "rev": "c9572e777febf7abcfbebf624e0323d82600f267",
-        "revCount": 2371,
+        "rev": "7c1f614ed3ad9ce0fd3291feccb1e5e4ff65adb3",
+        "revCount": 2372,
         "submodules": true,
         "type": "git",
         "url": "https://github.com/m-labs/misoc.git"
@@ -95,11 +95,11 @@
     "src-sipyco": {
       "flake": false,
       "locked": {
-        "lastModified": 1623807309,
-        "narHash": "sha256-FTRAS4RjqDOygu6+cP8mKbZHu/YZ7YKpEe2gSzJc9rk=",
+        "lastModified": 1632832039,
+        "narHash": "sha256-GYXXCCOxNZyy6j7qScB3/QWUUCEVX+4tM4bXXVGXty0=",
         "owner": "m-labs",
         "repo": "sipyco",
-        "rev": "20c946aad78872fe60b78d9b57a624d69f3eea47",
+        "rev": "b83d8e5d82b25dba9393f0c12bdc5253f8138545",
         "type": "github"
       },
       "original": {


occheung:
> what are the memory limits for the comm cpu (i.e. max size of uploaded binary)

In `runtime.ld`, 64MB is allocated to store the comm CPU binary (with `NOLOAD` sections, e.g. stack and heap) **and** the kernel support binary (`ksupport_data.o`, the part that performs dynamic linking, RPC with comm, include math library etc). But `ksupport_data.o` should be quite small. I built one with `nist_clock` variant on KC705 and it was < 400KB.

However, there might be some other factors that might limit the binary size. For example, the size of the memory region mapped to the SPI flash/ROM (IIRC 16MB for Kasli), and the gateware and other binaries (e.g. bootloader) is also included.

The actual size of `runtime.fbi` is generally a lot smaller than any of these limits, so most of the 64MB goes to the heap. (see traces from, #1761)

>and a running kernel (i.e. allocation of a large array on the stack in a kernel)?

The stack size of the kernel depends its binary size. First, kernel CPU region is around 175MB large (from 0x45060000 to 0x4FFFFFF0). Then, the page guard is added after every linker section, effectively after the kernel binary (the dynamic library part). The linker for kernel is written as such.
```
/* kernel.ld */

    /* Include every sections before this comment */

    . = ALIGN(0x1000);
    _sstack_guard = .;
} /* EOF */
```
The guard page is also 4096 bytes long, so the size of the kernel stack is whatever remains of it after the stack guard.

I think most of these numbers are the same since #1612. This patch only reduces the kernel stack size by no more than 8192 bytes (stack guard & stack alignment).

PR #1778, opened 2021-11-09
Add parameter attributes to LLVM IR call

### LINKED ISSUE #1599: LLVM IR call lacks parameter attributes
# Bug Report

## One-Line Summary
Call instruction does not contain `byval` parameter attributes for functions requiring it.

Example: call to `dma_record_start` would be lowered into the following call:
```
call void @dma_record_start({ i8*, i32 }* %.9), !dbg !32
```
but the declaration of the function is:
```
declare void @dma_record_start({ i8*, i32 }* byval)
```

### Steps to Reproduce

1. Compile the following code, and dump the IR
```python

class DMAPulses(EnvExperiment):
    def build(self):
        self.setattr_device("core")
        self.setattr_device("core_dma")

    @kernel
    def record(self):
        with self.core_dma.record("pulse"):
            pass

    @kernel
    def run(self):
        self.record()
```

### Expected Behavior

The call contains a `byval` attribute for the argument.

* The attribute is missing. This causes misoptimization in newer LLVM versions, and it may cause other memory corruption issues as it is UB.

* ARTIQ version: current master

sbourdeauducq:
Currently untested

Closes #1599

### DIFF
--- a/artiq/compiler/transforms/llvm_ir_generator.py
+++ b/artiq/compiler/transforms/llvm_ir_generator.py
@@ -1174,26 +1174,32 @@ class LLVMIRGenerator:
         else:
             llfun = self.map(insn.static_target_function)
         llenv     = self.llbuilder.extract_value(llclosure, 0, name="env.fun")
-        return llfun, [llenv] + list(llargs)
+        return llfun, [llenv] + list(llargs), {}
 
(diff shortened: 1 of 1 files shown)


dnadlinger:
Merged myself after adding a test case.

PR #1780, opened 2021-11-12
Almazny support

Spaqin:
# ARTIQ Pull Request

## Description of Changes

This PR adds a driver for Almazny - mezzanine board for Mirny. It allows the user to control the RF output with attenuators on the board or toggling the output.

This works in tandem with quartiq/mirny#5.

Tested with a dedicated script and a logic analyzer; after that worked, tested with BladeRF and modified ``artiq_sinara_tester``.

## Type of Changes

|   | Type |
|   | :bug: Bug fix  |
|   | :hammer: Refactoring  |
|   | :scroll: Docs |

### All Pull Requests

### Code Changes

### Git Logistics

  ```
  
  ```

### Licensing


artiq/frontend/artiq_ddb_template.py
```diff
@@ -294,6 +294,18 @@ def process_mirny(self, rtio_offset, peripheral):
             name=mirny_name,
             refclk=peripheral["refclk"],
             clk_sel=clk_sel)
+        almazny = peripheral.get("almazny")
+        if almazny == True:
```
sbourdeauducq, line 298: Is Almazny on a Mirny simply enabled by adding ``almazny: true`` in the Mirny JSON description? This sounds fine to me.
But don't write ``if x == True:``, just write ``if x:``. And it would be cleaner to write ``peripheral.get("almazny", False)``.
Doesn't the JSON schema need an update?

artiq/coredevice/mirny.py
```diff
@@ -140,3 +140,144 @@ def write_ext(self, addr, length, data):
...
+            (self.REG_LATCH_BASE + (reg_i - 1), 1)
+        ])
+        self._send_mezz_data([
+            (self.REG_LATCH_BASE + (reg_i - 1), 0)
+        ])
+
+    @kernel
+    def put_data(self, pin, d):
+        if d not in [0, 1]:
+            raise ValueError("Data can be only 0 or 1")
```
sbourdeauducq, line 269: Just use a boolean

artiq/coredevice/mirny.py
```diff
@@ -140,3 +140,144 @@ def write_ext(self, addr, length, data):
...
+    :param mirny - Mirny device Almazny is connected to
+    """
+
+    # pin map
+    SER_MOSI = 0
+    SER_CLK = 1
+    REG_LATCH_BASE = 2
+    REG_CLEAR = 6
+
+    mezzio_reg = 0x03  # register addr for Mezz_IO 0~7
```
sbourdeauducq, line 158: I'm not sure if we're going to support those class attributes in NAC3 (@pca006132 ?).
If not it may be better to put them in the module with a ``ALMAZNY_*`` prefix to reduce divergence.

artiq/coredevice/mirny.py
```diff
@@ -140,3 +140,144 @@ def write_ext(self, addr, length, data):
...
+            (self.SER_CLK, 0)
+            ])
+
+    @kernel
+    def reg_set(self, reg_i, attin1=1, attin2=1, attin3=1, attin4=1, attin5=1, attin6=1, output_on=0):
+        """
+        Sets the data on chosen shift register.
+        :param reg_i - index of the register [1-4]
+        :param attin[1-6] - attenuator on {0, 1} 
+        :attin1 - 16dB attenuator, att6 - 0.5dB
```
sbourdeauducq, line 181: Use an integer that represents the raw ("machine units") 6-bit value loaded into the attenuator. Write conversion functions to/from machine units. See how it's done on Mirny and Urukul where the situation is exactly the same.

artiq/frontend/artiq_sinara_tester.py
```diff
@@ -408,7 +414,7 @@ def test_mirnies(self):
         print("Frequencies:")
         for card_n, channels in enumerate(chunker(self.mirnies, 4)):
             for channel_n, (channel_name, channel_dev) in enumerate(channels):
-                frequency = 1000*(card_n + 1) + channel_n * 100 + 8     # Extra 8 Hz for easier observation
+                frequency = 1000*(card_n + 1) + channel_n * 100 + 8  # Extra 8 MHz for easier observation
```
sbourdeauducq, line 417: it doesn't really make the observation easier, remove the comment entirely

artiq/frontend/artiq_sinara_tester.py
```diff
@@ -96,6 +97,8 @@ def build(self):
                     self.suservos[name] = self.get_device(name)
                 elif (module, cls) == ("artiq.coredevice.suservo", "Channel"):
                     self.suschannels[name] = self.get_device(name)
+                elif (module, cls) == ("artiq.coredevice.mirny", "Almazny"):
+                    self.almaznys[name] = self.get_device(name)
```
sbourdeauducq, line 101: Won't this still run the Mirny test for the same card?

artiq/frontend/artiq_sinara_tester.py
```diff
@@ -340,6 +343,63 @@ def mirny_rf_switch_wave(self, channels):
...
+    def test_almaznys(self):
+        print("*** Testing Almaznys.")
+        for name, almazny in sorted(self.almaznys.items(), key=lambda x: x[0]):
+            print(name + "...")
+            self.init_almazny(almazny)
+            print("Testing attenuators. Frequencies:")
+            for card_n, channels in enumerate(chunker(self.mirnies, 4)):
+                for channel_n, (channel_name, channel_dev) in enumerate(channels):
+                    frequency = 1000*(card_n + 2) + channel_n * 100
+                    print("{}\t{}MHz (double is {}MHz)".format(channel_name, frequency, frequency*2))
```
sbourdeauducq, line 376: Only the doubled frequency matters here I think?
The other output should be tested during the Mirny test.

artiq/coredevice/mirny.py
pca006132: > I'm not sure if we're going to support those class attributes in NAC3, which should also be kernel invariants in this case (@pca006132 ?).

We can, but probably not recently as there are other higher priority tasks.

artiq/coredevice/mirny.py
```diff
@@ -140,3 +149,160 @@ def write_ext(self, addr, length, data):
...
+        :param att_mu - attenuation setting in machine units [0-63] 
+        :param output_on - RF output on (bool)
+        """
+        if not 4 >= channel >= 1:
+            raise ValueError("channel must be 1, 2, 3 or 4")
+        if not 63 >= att_mu >= 0:
+            raise ValueError("Invalid Almazny attenuator setting")
+
+        att_ins = [ (att_mu >> i) & 0x01 for i in range(6) ]
+        shift_reg = [0, 1 if output_on else 0] + att_ins
```
sbourdeauducq, line 212: I suggest using bit operators instead of those lists. Then you simply have ``att_ins = att_mu`` which can be removed and ``shift_reg = ((1 << 6) if output_on else 0) | att_mu``

artiq/coredevice/mirny.py
```diff
@@ -140,3 +149,160 @@ def write_ext(self, addr, length, data):
...
+            (ALMAZNY_REG_LATCH_BASE + (channel - 1), 1)
+        ])
+        self._send_mezz_data([
+            (ALMAZNY_REG_LATCH_BASE + (channel - 1), 0)
+        ])
+
+    @kernel
+    def put_data(self, pin, d):
+        self.mezz_data = (self.mezz_data & ~(1 << pin)) | d << pin  # data
+        self.mezz_data |= 1 << pin + 8  # oe
```
sbourdeauducq, line 296: Avoid those side effects. Just build an integer locally and pass it around.

sbourdeauducq: And I suggest having a constant OE mask, possibly defined in the module which is systematically ORed.

sbourdeauducq: If you want to retain state, then retain high-level state e.g. ``self.rf_switches = [False]*4; self.att_mu = [0x3f]*4``. Then at each update build the control word based on the high level state, and send it to the hardware.

artiq/coredevice/mirny.py
```diff
@@ -140,3 +149,160 @@ def write_ext(self, addr, length, data):
...
+        """
+        Convert a digital attenuator setting to dB.
+
+        :param att_mu: attenuator setting in machine units
+        :return: attenuator setting in dB
+        """
+        return att_mu / 2
+
+    @kernel
+    def set_att_mu(self, channel, att_mu, output_on=True):
```
sbourdeauducq, line 199: There should be separate methods for the attenuators and RF switch, like the other cards.

Spaqin: Unfortunately, we cannot set just the RF switch without setting the attenuators (and vice versa) - that's just how the shift register works. ...or I guess I could separate that if I keep the high-level internal state, and reapply one setting on change of the other.

artiq/coredevice/mirny.py
```diff
@@ -140,3 +151,232 @@ def write_ext(self, addr, length, data):
...
+            (ALMAZNY_REG_CLEAR, 1),
+            # rest is 0
+            ])
+
+    @kernel
+    def _reverse_att_mu(self, att_mu):
+        res = 0
+        for i in range(6):
+            res |= (att_mu >> i & 0x01) << (5-i)
+        return res
```
sbourdeauducq, line 180: Just send the data the other direction in ``_update_register`` and remove this method.

artiq/coredevice/mirny.py
```diff
@@ -140,3 +151,232 @@ def write_ext(self, addr, length, data):
...
+    @kernel
+    def _send_mezz_data(self, pins_data):
+        """
+        Sends the raw data to the mezzanine board.
+        :param pins_data - list of tuples in format (pin_number, bit)
+        """
+        # by default data is 0 - put 1s where you need
+        data = ALMAZNY_OE_MASK
+        for pin, d in pins_data:
+            data = (data & ~(1 << pin)) | d << pin
```
sbourdeauducq, line 379: Do we really need this list business? Bitwise operations aren't much more complicated to write than adding list elements and it would make the code a bit simpler and consistent with the way most low-level code is written. The runtime speed could be improved too.

artiq/coredevice/mirny.py
```diff
@@ -140,3 +151,232 @@ def write_ext(self, addr, length, data):
...
+        return self._reverse_att_mu(att_mu) / 2
+
+    @kernel
+    def set_att_mu(self, channel, att_mu):
+        """
+        Sets attenuators on chosen shift register (channel).
+        :param channel - index of the register [1-4]
+        :param att_mu - attenuation setting in machine units [0-63]
+        """
+        if not 4 >= channel >= 1:
```
sbourdeauducq, line 212: All other boards index channels starting at 0.

artiq/coredevice/mirny.py
```diff
@@ -140,3 +151,232 @@ def write_ext(self, addr, length, data):
...
+            raise ValueError("channel must be 1, 2, 3 or 4")
+        if not 63 >= att_mu >= 0:
+            raise ValueError("Invalid Almazny attenuator setting")
+
+        ch = channel - 1
+        self.att_mu[ch] = att_mu
+        self._update_register(ch)
+
+    @kernel
+    def set_att_mu_all(self, att_mu):
```
sbourdeauducq, line 222: Other boards don't have such a method and I'm not sure if there's a use case for it. Very disparate equipment, requiring different power levels, could be connected to each Almazny channel.

Spaqin: Urukul does have a method like that, that's what I based this on.

artiq/coredevice/mirny.py
```diff
@@ -140,3 +151,232 @@ def write_ext(self, addr, length, data):
...
+
+    @kernel
+    def att_to_mu(self, att):
+        """
+        Convert an attenuator setting in dB to machine units.
+
+        :param att: attenuator setting in dB [0-31.5] in 0.5 steps
+        :return: attenuator setting in machine units
+        """
+        if att > 31.5 or att < 0 or (att-int(att) != 0.0 and att-int(att) != 0.5):
```
sbourdeauducq, line 190: Remove the ``.5`` check, rounding is always present one way or another and generally acceptable.

artiq/coredevice/mirny.py
```diff
@@ -140,3 +151,232 @@ def write_ext(self, addr, length, data):
...
+            raise ValueError("Invalid Almazny attenuator settings!")
+        # it must be reversed for 
+        return self._reverse_att_mu(int(att * 2))
```
sbourdeauducq, line 193: round() not int()

artiq/coredevice/mirny.py
sbourdeauducq: Look closer - the Urukul method does something else: it lets you set different values for each attenuator. And please use exactly the same method names.

artiq/coredevice/mirny.py
```diff
@@ -140,3 +151,143 @@ def write_ext(self, addr, length, data):
...
+    """
+
+    def __init__(self, dmgr, host_mirny):
+        self.mirny = dmgr.get(host_mirny)
+        self.att_mu = [0x3f] * 4
+        self.rf_switches = [0] * 4
+
+    @kernel
+    def init(self):
+        self._send_mezz_data(ALMAZNY_REG_CLEAR)  # reset the rest
```
sbourdeauducq, line 170: Is this necessary?
My understanding is that this simply resets the data preloaded in the shift registers (left columns of FFs) which you are systematically overwriting later before pulsing the latch signal that transfers it to the right columns of FFs and the outputs. If this understanding is correct, this reset can and should be removed entirely, unless you also wanted to latch afterwards. But in this case, it is simpler to just load the values of ``att_mu`` and ``rf_switches`` in ``init`` so that there is a single source of truth for the initial values of the attenuators and RF switches.
image

artiq/coredevice/mirny.py
```diff
@@ -140,3 +151,143 @@ def write_ext(self, addr, length, data):
...
+    @kernel
+    def _latch(self, ch):
+        self._send_mezz_data(ALMAZNY_REG_CLEAR | 
+                             ALMAZNY_SER_CLK |
+                             ALMAZNY_REG_LATCH_BASE << ch)
+        self._send_mezz_data(ALMAZNY_REG_CLEAR)  # reset latch
+
+    @kernel
+    def _update_register(self, ch):
+        self._send_mezz_data(ALMAZNY_REG_CLEAR)  # reset all latches
```
sbourdeauducq, line 273: Why? Aren't you loading them completely immediately afterwards? There are only 8 bits inside and you are sending 8.

artiq/coredevice/mirny.py
```diff
@@ -140,3 +151,143 @@ def write_ext(self, addr, length, data):
...
+    @kernel
+    def att_to_mu(self, att):
+        """
+        Convert an attenuator setting in dB to machine units.
+
+        :param att: attenuator setting in dB [0-31.5]
+        :return: attenuator setting in machine units
+        """
+        if att > 31.5 or att < 0:
+            raise ValueError("Invalid Almazny attenuator settings!")
```
sbourdeauducq, line 181: You may want to only check after ``round`` - floating point computations are always imprecise and small rounding errors from the user code may otherwise unnecessarily trigger the error.

artiq/coredevice/mirny.py
```diff
@@ -156,11 +162,11 @@ class Almazny:
     def __init__(self, dmgr, host_mirny):
         self.mirny_cpld = dmgr.get(host_mirny)
         self.att_mu = [0x3f] * 4
+        self.channel_oe = [0] * 4
```
sbourdeauducq, line 165: Elsewhere it's called "switch", "RF switch", "sw", etc. and never OE.

artiq/frontend/artiq_sinara_tester.py
```diff
@@ -381,14 +381,20 @@ def test_almaznys(self):
...
+            for name, cpld in sorted(self.mirny_cplds.items(), key=lambda x: x[0]):
+                print(name + "...")
+                self.init_mirny(cpld)
+            print("...done")
+
             print("Testing attenuators. Frequencies:")
             for card_n, channels in enumerate(chunker(self.mirnies, 4)):
                 for channel_n, (channel_name, channel_dev) in enumerate(channels):
-                    frequency = 1000*(card_n + 2) + channel_n * 100
+                    frequency = 2000 + channel_n * 100
```
sbourdeauducq, line 393: Why that change?

artiq/frontend/artiq_sinara_tester.py
```diff
@@ -406,7 +412,7 @@ def test_almaznys(self):
             input()
             print("RF ON, all attenuators are ON. Press ENTER when done.")
             for i in range(4):
-                self.almazny_set_attenuators_db(almazny, i, 31)
+                self.almazny_set_attenuators_db(almazny, i, 31.5)
```
sbourdeauducq, line 415: Why?

artiq/coredevice/mirny.py
```diff
@@ -132,11 +142,124 @@ def set_att_mu(self, channel, att):
...
+    def set_att(self, channel, att):
+        """
+        Sets attenuators on chosen shift register (channel).
+        :param channel - index of the register [0-3]
+        :param att_mu - attenuation setting in dBm [0-31.5]
+        """
+        if not 3 >= channel >= 0:
+            raise ValueError("channel must be 0, 1, 2 or 3")
+        if not 31.5 >= att >= 0:
+            raise ValueError("Invalid Almazny attenuator setting")
```
sbourdeauducq, line 205: I think the other drivers don't do range checks...

artiq/frontend/artiq_sinara_tester.py
```diff
@@ -352,6 +355,69 @@ def mirny_rf_switch_wave(self, channels):
...
+        almazny.init()
+        almazny.output_toggle(True)
+
+    @kernel
+    def almazny_set_attenuators(self, almazny, ch, atts):
+        self.core.break_realtime()
+        almazny.set_att_mu(ch, atts)
+
+    @kernel
+    def almazny_set_attenuators_db(self, almazny, ch, atts):
```
sbourdeauducq, line 370: Convention is no suffix for the function with usual units and ``_mu`` for the one with hardware/machine units.
So it should be ``almazny_set_attenuators_mu`` and ``almazny_set_attenuators``.

artiq/coredevice/mirny.py
Spaqin: You are correct - the only check for Urukul for example is only in db->mu conversion.

PR #1787, opened 2021-11-24
Remove xbuild and use build-std feature

stevefan1999:
# ARTIQ Pull Request

## Description of Changes

This is a simple PR that removes the need for cargo xbuild, and introduces to use `build-std` that existed since nightly 2020–07–15. In fact, the developer of xbuild also suggested [using the build-std feature instead](https://github.com/rust-osdev/cargo-xbuild#alternative-the-build-std-feature-of-cargo). The only downside of this is a nightly version of Rust is required.

## Type of Changes

|   | Type |

It also logically reverts some of the series of changes in https://github.com/m-labs/artiq/commit/dcbd9f905c3ea9b221eb94fe6c838c375a71ca4b

### DIFF
--- a/artiq/firmware/Cargo.lock
+++ b/artiq/firmware/Cargo.lock
@@ -109,12 +109,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
-name = "compiler_builtins"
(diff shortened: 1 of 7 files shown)


stevefan1999:
Nix enforced to require vendored source. This is blocked by https://github.com/rust-lang/wg-cargo-std-aware/issues/23, otherwise we have to specify each and every dependencies in any of the lock files for them to be vendored-in. Closing this right now as it is not very feasible to find any missing dependencies or until that particular issue is fixed.

PR #1801, opened 2021-12-16
comm_kernel: only send replies when no user-defined RPC function throws in service

### LINKED ISSUE #1655: RPCs corrupt integers in lists with little warning
# Bug Report

## One-Line Summary

RPCs returning a `TList(TIntXX)`, return corrupted data for integers outside the IntXX range

## Issue Details

Retruning a list of integers from an RPC results in data corruption if any integers are outside the range of the integer returned by the RPC. This is the case for both `Int32` and `Int64`. This event is indicated by a message logged on debug level. 

The relevant message reads:
`artiq.coredevice.comm_kernel:rpc service: 3 [<artiq_worker_rpc_ints.RPCIntCorruption object at 0x7f1a7756d940>] {} ! error("'l' format requires -2147483648 <= number <= 2147483647")`

The analagous situation when returning Integers form RPCs is handled much more verbosely. Returning out-of-range integers from RPCs raises an exception:
`root:Terminating with exception (RPCReturnValueError: type mismatch: cannot serialize 2147483648 as 32-bit int`

Related comment from @sbourdeauducq 
https://github.com/m-labs/artiq/pull/1628#issuecomment-816298822

### Steps to Reproduce
Run the Following example:
```Python3


class RPCIntCorruption(EnvExperiment):
    def build(self):
        self.setattr_device("core")

    def int_from_rpc(self) -> TList(TInt32):
        number = 1<<31
        print(number)
        return [number]

    @kernel
    def run(self):
        print(self.int_from_rpc())
```

This prints:
```Python3
print:2147483648
print:[1515870810]
```
### Expected Behavior

The user should be alerted. Raising an exception, or logging a warning seem appropreate.

* ARTIQ version: Latest Artiq

stevefan1999:
# ARTIQ Pull Request

## Description of Changes

This PR fixes an issue when an exception happens after running the service function in RPC service routine. Previously I tried to catch `struct.error` as a quick fix, but it seems like the fundamental issue is not just because of that. 

The problem started with this part, after running the service routine:
https://github.com/m-labs/artiq/blob/bbac4770926048b3c0b434aede371a63d198ecd5/artiq/coredevice/comm_kernel.py#L561-L564

And when an exception happened around that part:

https://github.com/m-labs/artiq/blob/bbac4770926048b3c0b434aede371a63d198ecd5/artiq/coredevice/comm_kernel.py#L568-L572

It is possible that the `RPCException` is being mistaken as part of the `RPCReply`, which is going to confuse the host kernel.

One possible way of fixing this is to discard the write buffer when an exception happens. Another solution is to move the RPC reply after service call was handled and checked without error, and if something is wrong in the RPC reply process, let it propagate properly
 
Check out https://github.com/m-labs/artiq/issues/1655#issuecomment-995455469 as well for extra context.

### Related Issue

Closes #1655

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

### Code Changes

### Licensing

### DIFF
--- a/artiq/coredevice/comm_kernel.py
+++ b/artiq/coredevice/comm_kernel.py
@@ -557,12 +557,6 @@ class CommKernel:
             result = service(*args, **kwargs)
             logger.debug("rpc service: %d %r %r = %r",
                          service_id, args, kwargs, result)
-
-            self._write_header(Request.RPCReply)
-            self._write_bytes(return_tags)
-            self._send_rpc_value(bytearray(return_tags),
-                                 result, result, service)
-            self._flush()
         except RPCReturnValueError as exn:
             raise
         except Exception as exn:
@@ -608,6 +602,12 @@ class CommKernel:
                 self._write_int32(line)
                 self._write_int32(-1)  # column not known
                 self._write_string(function)
+            self._flush()
+        else:
+            self._write_header(Request.RPCReply)
+            self._write_bytes(return_tags)
+            self._send_rpc_value(bytearray(return_tags),
+                                 result, result, service)
             self._flush()
 
     def _serve_exception(self, embedding_map, symbolizer, demangler):
--- a/artiq/coredevice/comm_kernel.py
+++ b/artiq/coredevice/comm_kernel.py
@@ -555,8 +555,6 @@ class CommKernel:
 
         try:
             result = service(*args, **kwargs)
-            logger.debug("rpc service: %d %r %r = %r",
-                         service_id, args, kwargs, result)
         except RPCReturnValueError as exn:
             raise
         except Exception as exn:
@@ -604,6 +602,8 @@ class CommKernel:
                 self._write_string(function)
             self._flush()
         else:
+            logger.debug("rpc service: %d %r %r = %r",
+                         service_id, args, kwargs, result)
             self._write_header(Request.RPCReply)
             self._write_bytes(return_tags)
             self._send_rpc_value(bytearray(return_tags),


sbourdeauducq:
> remove the fast path

What is the impact on performance?

stevefan1999:
> > remove the fast path
> 
> What is the impact on performance?

Having preallocation on list can have some dramatic speed up over some large arrays as shown in [this SO thread](https://stackoverflow.com/questions/9940859/fastest-way-to-pack-a-list-of-floats-into-bytes-in-python). 

However, for our use cases, which is focused around small blocks of data, I would say the performance loss is negligible at this moment.

artiq/test/coredevice/test_embedding.py
```diff
@@ -117,6 +117,89 @@ def test_default_arg(self):
...
+        x = randint(-2 ** 63, (2 ** 63) - 1)
+
+        def h() -> TInt64:
+            return x
+
+        self.assert_roundtrip_value_equal(g)
+        self.assert_roundtrip_value_equal(h)
+
+    def test_array_i32(self):
+        x = randint(-2 ** 31, (2 ** 31) - 1)
```
sbourdeauducq, line 159: Where is randint defined? Did you test this?
Also, deterministic tests are better.

sbourdeauducq:
> focused around small blocks of data

Really?
Lists are typically the way large amounts of data are transferred with ARTIQ and we have a unit test that measures the performance of that. It's even mentioned in the Zynq paper.

> It also fixes an issue in the range checker where there is several off-by-one minor mistakes that is commonly seen.

That should be a separate PR.

artiq/test/coredevice/test_embedding.py
stevefan1999: thanks for the reminder!

stevefan1999:
> Lists are typically the way large amounts of data are transferred with ARTIQ and we have a unit test that measures the performance of that. It's even mentioned in the Zynq paper.

Correct, there are substantial hits:

- Before 
```
| Test                 | Mean (MiB/s) |  std (MiB/s) |
| -------------------- | ------------ | ------------ |
| I32 Array (1MB) H2D  |         2.48 |         0.01 |
| I32 Array (1MB) D2H  |         2.65 |         0.00 |
| I32 Array (1KB) H2D  |         0.40 |         0.01 |
| I32 Array (1KB) D2H  |         0.42 |         0.02 |
| Bytes List (1MB) H2D |         2.44 |         0.02 |
| Bytes List (1MB) D2H |         2.49 |         0.08 |
| Bytes List (1KB) H2D |         0.43 |         0.03 |
| Bytes List (1KB) D2H |         0.47 |         0.03 |
| Bytes (1MB) H2D      |         2.50 |         0.00 |
| Bytes (1MB) D2H      |         2.66 |         0.00 |
| Bytes (1KB) H2D      |         0.45 |         0.05 |
| Bytes (1KB) D2H      |         0.50 |         0.05 |
| I32 List (1MB) H2D   |         2.42 |         0.03 |
| I32 List (1MB) D2H   |         2.60 |         0.02 |
| I32 List (1KB) H2D   |         0.42 |         0.02 |
| I32 List (1KB) D2H   |         0.46 |         0.02 |
```
- After:
```
| Test                 | Mean (MiB/s) |  std (MiB/s) |
| -------------------- | ------------ | ------------ |
| I32 Array (1MB) H2D  |         2.50 |         0.01 |
| I32 Array (1MB) D2H  |         2.67 |         0.00 |
| I32 Array (1KB) H2D  |         0.39 |         0.02 |
| I32 Array (1KB) D2H  |         0.40 |         0.00 |
| Bytes List (1MB) H2D |         0.92 |         0.02 |
| Bytes List (1MB) D2H |         2.46 |         0.07 |
| Bytes List (1KB) H2D |         0.18 |         0.02 |
| Bytes List (1KB) D2H |         0.46 |         0.00 |
| Bytes (1MB) H2D      |         2.49 |         0.00 |
| Bytes (1MB) D2H      |         2.66 |         0.00 |
| Bytes (1KB) H2D      |         0.40 |         0.02 |
| Bytes (1KB) D2H      |         0.44 |         0.02 |
| I32 List (1MB) H2D   |         2.37 |         0.04 |
| I32 List (1MB) D2H   |         2.59 |         0.02 |
| I32 List (1KB) H2D   |         0.24 |         0.01 |
| I32 List (1KB) D2H   |         0.44 |         0.02 |
```
Diff (significant data only):
```diff
1,4c1,4
< | Bytes List (1MB) H2D |         2.44 |         0.02 |
< | Bytes List (1KB) H2D |         0.43 |         0.03 |
< | I32 List (1MB) H2D   |         2.42 |         0.03 |
< | I32 List (1KB) H2D   |         0.42 |         0.02 |
---
> | Bytes List (1MB) H2D |         0.92 |         0.02 |
> | Bytes List (1KB) H2D |         0.18 |         0.02 |
> | I32 List (1MB) H2D   |         2.37 |         0.04 |
> | I32 List (1KB) H2D   |         0.24 |         0.01 |
```
So sending megabytes of data definitely took some hits, and we will need some compromises here.

> That should be a separate PR.
Sure. By the way, do we need to include NaN check for floating point numbers (which means all exponents 1)?

sbourdeauducq:
``struct`` already raises an appropriate exception, please figure out what happens to the exception and why, instead of randomly changing the code (and making it slow) so it works by chance.

```
>>> struct.pack("<1l", 1 << 40)
  File "<stdin>", line 1, in <module>
struct.error: 'l' format requires -2147483648 <= number <= 2147483647
```

stevefan1999:
@sbourdeauducq I think we shouldn't modify `_serve_rpc` to also raise `struct.error` since just raising it is pretty vague for debugging purposes, may we should wrap the list serialization code in another try handler and throw it as a nested RPCReturnValueError instead

sbourdeauducq:
Which exception is raised is not relevant at the moment; the problem is the data corruption on the socket.

pca006132:
For impact on performance, I suggest testing on Zynq, as the speed on or1k is limited by the device side and may not reveal host side performance problems.

artiq/coredevice/comm_kernel.py
```diff
@@ -565,6 +565,8 @@ def service(obj, attr, value): return setattr(obj, attr, value)
             self._flush()
         except RPCReturnValueError as exn:
             raise
+        except struct.error as exn:
+            raise RPCReturnValueError(f"cannot serialize data {result} into bytes, exception: {exn}")
```
sbourdeauducq, line 569: Use ``raise from``, not ``exception: {exn}``.
Why is ``RPCReturnValueError`` the correct exception type to use here?

sbourdeauducq:
Code looks good, @stevefan1999 is this tested carefully?

artiq/coredevice/comm_kernel.py
```diff
@@ -557,12 +557,6 @@ def service(obj, attr, value): return setattr(obj, attr, value)
             result = service(*args, **kwargs)
             logger.debug("rpc service: %d %r %r = %r",
```
sbourdeauducq, line 558: Why is ``logger.debug`` not in the ``else`` clause?

PR #1802, opened 2021-12-16
llvm_ir: move stacksave before lltag alloca in build_rpc

### LINKED ISSUE #1798: RPC-heavy experiment triggers StoreFault exception
# Bug Report

## One-Line Summary

Certain RPC-heavy experiments trigger the following error (in the coredevice UART logs) after a few hours runtime:

```
[    28.437425s]  INFO(runtime::kern_hwreq): resetting RTIO
[  6823.673265s]  INFO(kernel): panic at ksupport/lib.rs:536:5: Exception(StoreFault) at PC 0x450000e0, trap value 0x45062000
[  6823.683309s] ERROR(runtime::session): session aborted: unexpected request RunAborted from kernel CPU
[  6823.692412s]  INFO(runtime::session): no connection, starting idle kernel
[  6823.699487s]  INFO(runtime::session): no idle kernel found
```

## Issue Details

- for a given experiment, the time elapsed between the last RTIO reset and the `StoreFault` exception is quite consistent between reproductions
- no backtrace is given in the logs (see above). The coredevice recovers but the experiment is interrupted
- running `llvm-objdump -CD runtime.elf` gives no addresses that correspond to the reported PC where the exception is raised, contrary to @occheung 's demo in #1764. Is this expected?
- used runtime: [runtime.elf.gz](https://github.com/m-labs/artiq/files/7706222/runtime.elf.gz)

### Steps to Reproduce

Reproduction code is at [airwoodix/artiq-storefault-repro](https://github.com/airwoodix/artiq-storefault-repro). The environment is provided by a Nix flake.

1. Copy the test crate's device db in the repository root. `ttl4-6` are expected to be outputs. Target architecture is `rv32ima` (Kasli v1.1)
2. Start the `rpyc` endpoint: `python rpc.py`
3. Execute the ARTIQ kernel: `artiq_run kernel.py`

The reproduction is on recent ARTIQ-7 with RISC-V softcores. The issue has been discovered on a OR1k system (also ARTIQ-7). Both carriers are Kasli v1.1.

Variations of the reproduction code lead to varying behavior. `msgpack` serialization is used to increase the RPC throughput (raw bytes vs lists) and hence the rate of `rpyc` RPC calls, which *seems* to accelerate the error reproduction. It is not clear whether the error is linked to the use of these two libraries but probably not.

### Expected Behavior

Kernel executes "forever" with output on `ttl4-6`.

* After ca. 2h (for the code at airwoodix/artiq-storefault-repro@7efd5702cccb2df60c7a9dcfd60fdb714270ff8d), the kernel execution stops
* Log message, tracebacks:
  *  `Exception(StoreFault)`: see above. Slight variations of the code have triggered the error after much longer times (6-7h). Some occurences failed at PC `0x450000f0`.
  * At `artiq_run` call-site: `ConnectionResetError`

* Operating System: NixOS, nixpkgs 21.11
* ARTIQ version: 7.0-dev, m-labs/artiq@9bbf7eb48539082ee57e3f9b6faef7b0246a042b
* Version of the gateware and runtime loaded in the core device: same as software
* Hardware involved: Kasli v1.1, DIO_SMA. Crate also contains other modules.

stevefan1999:
Signed-off-by: Steve Fan <sf@m-labs.hk>

# ARTIQ Pull Request

## Description of Changes

There is a minor stack errata that made some pointers dangling after restoring the stack.

https://github.com/m-labs/artiq/blob/9e5e234af31a82d053bc6c1add9eedc8bcf1d99b/artiq/compiler/transforms/llvm_ir_generator.py#L1252-L1257

This creates a potential "use-after-free" scenario (there is no `free` since this is all on stack, however, so it is more like invalid stack pointer), but this is still protected by PMP supposedly.

### Related Issue

Closes #1798

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

### Code Changes

### Licensing

### DIFF
--- a/artiq/compiler/transforms/llvm_ir_generator.py
+++ b/artiq/compiler/transforms/llvm_ir_generator.py
@@ -1249,12 +1249,12 @@ class LLVMIRGenerator:
             self.engine.process(diag)
         tag += ir.rpc_tag(fun_type.ret, ret_error_handler)
 
+        llstackptr = self.llbuilder.call(self.llbuiltin("llvm.stacksave"), [],
+                                         name="rpc.stack")
+
         lltag = self.llconst_of_const(ir.Constant(tag, builtins.TStr()))
         lltagptr = self.llbuilder.alloca(lltag.type)
         self.llbuilder.store(lltag, lltagptr)
-
-        llstackptr = self.llbuilder.call(self.llbuiltin("llvm.stacksave"), [],
-                                         name="rpc.stack")
 
         llargs = self.llbuilder.alloca(llptr, ll.Constant(lli32, len(args)),
                                        name="rpc.args")


dnadlinger:
It isn't obvious to me why this would fix the issue. Whether the `stacksave` is before or after the pointer alloca shouldn't matter, as we `stackrestore` it explicitly anyways.

This whole argument alloca construction doesn't really follow the usual use case of LLVM (in languages with mostly fixed layouts), so perhaps we are tickling some codegen bugs/limitations here?

Did you diff the ASM to see what actually changed? For the issues you observe that are now fixed, why do they work after the change? Is it a codegen bug in the sense of direct memory corruption that is removed, or is it that the change e.g. makes the stack no longer grow unbounded over time?

occheung:
> It isn't obvious to me why this would fix the issue. Whether the `stacksave` is before or after the pointer alloca shouldn't matter, as we `stackrestore` it explicitly anyways.

Memory allocated by `alloca` is released when the LLVM function returns, and the RPC send and/or recv sequences are not wrapped inside a function. So I think there are no implicit `stackrestore` there.

> For the issues you observe that are now fixed, why do they work after the change?

It simply deallocate the RPC `type` argument after RPC send terminates, by restoring to the stack pointer before any parameters of RPC was allocated.

Tested with [this ARTIQ_Python script](https://github.com/m-labs/artiq/issues/1798#issuecomment-993310785). Attached is the unoptimized LLVM before the patch.

[hello_pre_fix_unopt.zip](https://github.com/m-labs/artiq/files/7731866/hello_pre_fix_unopt.zip)

Another unoptimized LLVM (`hello_unopt.ll`) was generated after applying this patch. Ran `diff hello_pre_fix_unopt.ll hello_unopt.ll`.

```
37a38
>   %rpc.stack = call i8* @llvm.stacksave()
40d40
<   %rpc.stack = call i8* @llvm.stacksave()
82a83
>   %rpc.stack = call i8* @llvm.stacksave()
85d85
<   %rpc.stack = call i8* @llvm.stacksave()
```

dnadlinger:
Ah right, thanks – I missed we emit the tag `alloca` directly in-line, rather than at the beginning of the function like we should for locals.

The bit about "dangling pointers"/"use after free" in the PR description is misleading, though; the actual issue fixed here is that each RPC call behaved, in C terms, like it had an accompanying `alloca()`, which would lead to ever-growing stack usage when called in a loop.

sbourdeauducq:
Should this go on the release branches as well?

dnadlinger:
Yep, should be low-risk and a clear bug fix.

pca006132:
Just wondering: Why should we do alloca for tags? Can't we allocate this as a const global value as this is a static string?

dnadlinger:
It probably should be an unnamed_addr global, yes.

dnadlinger:
The whole RPC interface is a bit janky; we should rethink this for NAC3. (Probably just loan out a buffer from the comm CPU to the kernel CPU – faster this way, too.)

pca006132:
Loan out a buffer from the comm CPU might not be that simple as we try to avoid sharing memory between the two CPUs in zynq. Instead, I think it should be possible to compute the total required buffer size on the host (accounting for things like alignment), so we will only need 1 alloca call in the kernel and minimize communications between comm CPU and kernel CPU?

PR #1806, opened 2021-12-20
comm_kernel: fix off-by-one error for numeric value range check

stevefan1999:
# ARTIQ Pull Request

## Description of Changes
This is the spun-off of #1801, where there are several off-by-one minor mistakes in `_send_rpc_value`. 

Unit tests included.

## Explanation

Let w be the number of bits and a w-bit signed integer i in two's complement, we should check if i is in [-2^(w-1), 2^(w-1)). However in the kernel communicator code, it is mistaken as (-2^(w-1), 2^(w-1)-1), which means (-2^(w-1), 2^(w-1)-2]. 

Equivalently, if you have an unsigned w-bit integer, check [0, 2^w) 

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

### Code Changes

### Licensing

### DIFF
--- a/artiq/coredevice/comm_kernel.py
+++ b/artiq/coredevice/comm_kernel.py
@@ -476,11 +476,27 @@ class CommKernel:
             if tag_element == "b":
                 self._write(bytes(value))
             elif tag_element == "i":
-                self._write(struct.pack(self.endian + "%sl" %
-                                        len(value), *value))
+                try:
+                    data = struct.pack(self.endian + "%sl" % len(value), *value)
+                except struct.error:
+                    raise RPCReturnValueError(
+                        "type mismatch: cannot serialize {value} as {type}"
+                        " ({function} has returned {root})".format(
+                            value=repr(value), type="32-bit integer list",
+                            function=function, root=root))
+                else:
+                    self._write(data)
             elif tag_element == "I":
-                self._write(struct.pack(self.endian + "%sq" %
-                                        len(value), *value))
+                try:
+                    data = struct.pack(self.endian + "%sq" % len(value), *value)
+                except struct.error:
+                    raise RPCReturnValueError(
+                        "type mismatch: cannot serialize {value} as {type}"
+                        " ({function} has returned {root})".format(
+                            value=repr(value), type="64-bit integer list",
+                            function=function, root=root))
+                else:
+                    self._write(data)
             elif tag_element == "f":
                 self._write(struct.pack(self.endian + "%sd" %
                                         len(value), *value))
--- a/artiq/coredevice/comm_kernel.py
+++ b/artiq/coredevice/comm_kernel.py
@@ -476,27 +476,11 @@ class CommKernel:
             if tag_element == "b":
                 self._write(bytes(value))
             elif tag_element == "i":
-                try:
-                    data = struct.pack(self.endian + "%sl" % len(value), *value)
-                except struct.error:
-                    raise RPCReturnValueError(
-                        "type mismatch: cannot serialize {value} as {type}"
-                        " ({function} has returned {root})".format(
-                            value=repr(value), type="32-bit integer list",
-                            function=function, root=root))
-                else:
-                    self._write(data)
+                self._write(struct.pack(self.endian + "%sl" %
+                                        len(value), *value))
             elif tag_element == "I":
-                try:
-                    data = struct.pack(self.endian + "%sq" % len(value), *value)
-                except struct.error:
-                    raise RPCReturnValueError(
-                        "type mismatch: cannot serialize {value} as {type}"
-                        " ({function} has returned {root})".format(
-                            value=repr(value), type="64-bit integer list",
-                            function=function, root=root))
-                else:
-                    self._write(data)
+                self._write(struct.pack(self.endian + "%sq" %
+                                        len(value), *value))
             elif tag_element == "f":
                 self._write(struct.pack(self.endian + "%sd" %
                                         len(value), *value))
--- a/artiq/test/coredevice/test_embedding.py
+++ b/artiq/test/coredevice/test_embedding.py
@@ -168,7 +168,6 @@ class RPCHostRoundtripTypeTest(ExperimentCase):
         self.assert_roundtrip_value_equal(i64_maxmin)
 
     def test_i32_bad(self):
-        # Notice: the constant itself is valid if casted as u32
         def invalid_as_i32() -> TInt32:
             return 2 ** 31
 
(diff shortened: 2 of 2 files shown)


artiq/test/coredevice/test_embedding.py
```diff
@@ -1,6 +1,8 @@
 import numpy
 import unittest
 from time import sleep
+from random import randint, seed
+seed(0xdeadbeef)
```
sbourdeauducq, line 5: Please avoid global state which still has poor reproducibility - create local ``Random`` objects instead. Or just don't use pseudorandom numbers at all.

artiq/coredevice/comm_kernel.py
```diff
@@ -436,13 +436,15 @@ def check(cond, expected):
                   lambda: "bool")
             self._write_bool(value)
         elif tag == "i":
+            # notice: let b be the number of bits, check [-2^(b-1), 2^(b-1))
+            # equivalent: check [0, 2^b)
```
sbourdeauducq, line 440: Not sure what this comment means or if it adds anything to the code.

artiq/test/coredevice/test_embedding.py
stevefan1999: Okay. Of course, if we have something like QuickCheck, it would be better.

artiq/coredevice/comm_kernel.py
stevefan1999: that's the explanation for the code change

sbourdeauducq: Put it in the PR description (if anywhere), not comment. It's quite obvious already what is going on, that was just a silly mistake...

stevefan1999: Okay. This off by one mistake is very common by the way.

artiq/test/coredevice/test_embedding.py
```diff
@@ -117,6 +118,92 @@ def test_default_arg(self):
...
+            return x
+
+        self.assert_roundtrip_value_equal(g)
+        self.assert_roundtrip_value_equal(h)
+
+    def test_i64(self):
+        def g() -> TInt64:
+            return (2 ** 63) - 1
+        rng = Random(0xdeadbeef)
+        x = rng.randint(-2 ** 63, (2 ** 63) - 1)
```
sbourdeauducq, line 152: Same problem as above. Also below.

artiq/test/coredevice/test_embedding.py
```diff
@@ -117,6 +118,92 @@ def test_default_arg(self):
...
+class _RPCOverflowTest(EnvExperiment):
+    def build(self):
+        self.setattr_device("core")
+
+    @kernel
+    def run(self, f, g):
+        f(g())
+
+
+class RPCOverflowTest(ExperimentCase):
```
sbourdeauducq, line 130: There could be many "overflows" with RPC. Rename to RPCIntOverflowTest?

artiq/test/coredevice/test_embedding.py
```diff
@@ -117,6 +118,92 @@ def test_default_arg(self):
...
+    def assert_roundtrip_value_equal(self, g):
+        exp = self.create(_RPCOverflowTest)
+        exp.run(lambda x: self.assertEqual(g(), x), g)
+
+    def test_i32(self):
+        def g() -> TInt32:
+            return (2 ** 31) - 1
+
+        rng = Random(0xdeadbeef)
+        x = rng.randint(-2 ** 31, (2 ** 31) - 1)
```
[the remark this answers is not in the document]
stevefan1999: right, it is so tempting to remove it, but how do I prove the function would be correct between (min int32, max int32)? that's why I chose to use a random value and see if it passes, then we could prove it works on all cases

artiq/test/coredevice/test_embedding.py
stevefan1999: I've change the name to RPCHostRoundtripTypeTest, it should sound a lot better

artiq/test/coredevice/test_embedding.py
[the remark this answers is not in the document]
sbourdeauducq: Not much difference between ``x`` and ``prng(seed, x)`` for only one value, except that the former is simpler.
I don't see why the code would fail for some intermediate value anyway, testing at the limits should be enough.

artiq/coredevice/comm_kernel.py
```diff
@@ -476,11 +476,27 @@ def check(cond, expected):
             if tag_element == "b":
                 self._write(bytes(value))
             elif tag_element == "i":
-                self._write(struct.pack(self.endian + "%sl" %
-                                        len(value), *value))
+                try:
+                    data = struct.pack(self.endian + "%sl" % len(value), *value)
+                except struct.error:
+                    raise RPCReturnValueError(
```
sbourdeauducq, line 482: What problem is that supposed to solve?

stevefan1999: This used to be checked one by one before
but it comes at a great performance penalty cost (see the benchmark above)
and we found the root cause in #1801, so we should translate `struct.error` instead
so as to keeping the same behavior of what the check functions did in serializing the primitive types above, aka throwing an RPCReturnValueError error

sbourdeauducq:
I have manually committed the first part of this PR. The rest looks shaky to me.

artiq/coredevice/comm_kernel.py
sbourdeauducq: I don't understand

stevefan1999: maybe I should move this hunk to #1801 as well?

sbourdeauducq: I don't see why it should be there at all.

stevefan1999: https://github.com/m-labs/artiq/blob/d7dd75e833fc99ad75e10220a5ec8f03d64ee5bd/artiq/coredevice/comm_kernel.py#L438-L447
I was referring to this part of the code, shouldn't we check if all elements in TList[int]/TList[int64] are within bound as well? Otherwise it will just throw `struct.error` on the array roundtrip test, which is weird.

sbourdeauducq: OK, I see. Should be a separate PR anyway.

artiq/test/coredevice/test_embedding.py
```diff
@@ -117,6 +117,87 @@ def test_default_arg(self):
...
+        self.assert_roundtrip_value_equal(i32_maxmin)
+
+    def test_array_i64(self):
+        def i64_maxmin() -> TList(TInt64):
+            return [(2 ** 63) - 1, -2 ** 63]
+
+        self.assert_roundtrip_value_equal(i64_maxmin)
+
+    def test_i32_bad(self):
+        # Notice: the constant itself is valid if casted as u32
```
sbourdeauducq, line 171: and?

PR #1820, opened 2022-01-07
Remove worker_db warning for writing a dataset that is also in the archive

lriesebos:
# ARTIQ Pull Request

## Description of Changes

When using `get_dataset()` to obtain a key that was stored persistent, the value is added to the archive of the dataset manager.
When using `set_dataset()` with a key that was archived earlier, you get the following warning:

```
WARNING:worker(219,artiq_test.py):artiq.master.worker_db:Modifying dataset 'foo' which is in archive, archive will remain untouched
```

I think this warning is overkill as this is the expected behavior of the archive. Nevertheless, the log can have many of these warnings. The warning especially appears when we load calibrated values from the persistent datasets and update some of these calibration values at the end of the experiment.

If others agree, I propose to remove this warning.

## Type of Changes

|   | Type |
| ✓  | :sparkles: New feature? |

### All Pull Requests

### Git Logistics

  ```
  
  ```

### Licensing

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -34,6 +34,7 @@ Highlights:
 * HDF5 options can now be passed when creating datasets with ``set_dataset``. This allows
   in particular to use transparent compression filters as follows:
   ``set_dataset(name, value, hdf5_options={"compression": "gzip"})``.
+* Removed worker DB warning for writing a dataset that is also in the archive
 
 Breaking changes:
 
--- a/artiq/master/worker_db.py
+++ b/artiq/master/worker_db.py
@@ -120,11 +120,6 @@ class DatasetManager:
 
     def set(self, key, value, broadcast=False, persist=False, archive=True,
             hdf5_options=None):
-        if key in self.archive:
-            logger.warning("Modifying dataset '%s' which is in archive, "
-                           "archive will remain untouched",
-                           key, stack_info=True)
-
         if persist:
             broadcast = True
 


hartytp:
❤️ 

Thanks @lriesebos for filing the issue. We also see that the vast majority of our logged warnings are just servos and other bits of automation chuntering away happily in the background. This devalues the logs (people get used to just ignoring all warnings).

At present, one can prevent the warning from being issued by setting `archive=False` when the dataset value is initially fetched. However this tends not ergonomic because the fetching is done by reusable code, which should not need to be aware of the context it's being used in (and hence whether or not to archive).

I raised a similar point on M-Labs chat recently https://chat.m-labs.hk/m-labs/pl/duioum6g4jgbije9rt5mxw43aa It turns out my mental model of dataset archiving wasn't correct. There was some reluctance to remove the warning, but there are alternative suggestions for how to resolve this.

lriesebos:
Thanks for pointing that out in the chat, I totally missed that (I don't look that often at the chat).

In my opinion, `set_dataset(..., archive=False)` is a bad solution as I might want to have the value archived. Coming back to my calibration example:
1. We get the persistent dataset to obtain the latest calibration value. The value is then archived in the `archive/` group in the HDF5 file. This is desired because we can then later see from the HDF5 file what calibration values were obtained from the persistent storage for this experiment.
2. After a calibration we want to store the new value persistently again. At the same time, we also want the value to be archived in the `datasets/` group in the HDF5 file. This allows us again to look at the HDF5 file and understand what the results of the calibration are.

So functionality wise this all works fine with vanilla ARTIQ, except for the  fact that I get this warning. As said above, I don't think this warning is useful. The software works as expected and there is no reason to warn.
The other dataset warning https://github.com/m-labs/artiq/blob/master/artiq/master/worker_db.py#L178 can stay in my opinion. A value in the archive is overwritten, so there is at least a reason to warn the user.

sbourdeauducq:
This needed a documentation update: https://github.com/m-labs/artiq/commit/eaa1505c947c7987cdbd31c24056823c740e84e0

Still, I'm not sure if this is the best solution. What about having three archive options:
* archived with warning (the old behavior)
* archived without warnings
* not archived

Could one of you write, carefully test, and document a patch for that?

lriesebos:
@sbourdeauducq I am wondering, what is the essence of this specific warning? In my opinion, there is no reason for a warning and I have not heard a compelling motivation yet on why it should be there. The warning could be converted to an info, which in my opinion could be more appropriate, but still I do not see value for the user. Adding more (unnecessary) warning messages will only lead to users ignoring them. For that reason, I do not think it is worth adding extra complexity just to get this warning back. Basically I am perceiving the presence of the warning as a bug.

PR #1824, opened 2022-01-11
comm_kernel: check if elements are within bounds for RPC list

stevefan1999:
Signed-off-by: Steve Fan <sf@m-labs.hk>

# ARTIQ Pull Request

## Description of Changes

Adds missing type validator in `_send_rpc_value` when handling list of i32 and i64 types by checking `struct.error` and throw a RPCReturnValueError.  

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

### Licensing


sbourdeauducq:
Try/except/else is not required here since an exception is always re-raised; and ``self._write`` would never raise ``struct.error``. Please simplify this patch.

PR #1832, opened 2022-01-14
Add DHCP support to Core devices

### LINKED ISSUE #1085: core device DHCP support
Support assignment of the core device IP address by DHCP.
See also https://github.com/m-labs/artiq/issues/955

mbirtwell:
# ARTIQ Pull Request

### DHCP changes

Adds a thread to configure the interface based on DHCP events
Adds a `Dhcpv4Socket` as a wrapper around smoltcp's version
Formalises the storage of the IP addresses so that we can update one in
another module.

There's also a workaround for the first DHCP discover packet frequently
going missing.

It's on by default but that's in a separate commit, which we can drop if we 
hate it. 

Tested:
* Booting with a ip setting respects it still
* Booting with no setting and bad one or use_dhcp enables the DHCP client and an address is acquired.

### artiq_upload

artiq_upload.py can be used to upload firmware images to the bootloader.
I've left it with a somewhat generic name because it probably makes sense to
the gateware upload support to the same script. (The bootloader already
implements it's side of this.)

My only interest in creating this was to test the smoltcp bootloader changes.

### Upgrade smoltcp 0.6 -> 0.8

Main changes:
* Deal with interfaces now being generic over mediums, update interface name
and initialisation.
* Interfaces now own their sockets. So we store a reference to the Interface
instead of the SocketSet in Scheduler and IO.
* Sockets are no longer reference counted. We never called the function to
increase the socket's reference count, so now we just remove it where it
was previously released. This will result in the socket being dropped at
a different time, but I think that should be fine.

### Related Issue

Closes #1085

## Type of Changes

|   | Type |

### All Pull Requests

- [ ] Close/update issues.

### Code Changes

### Documentation Changes

### Git Logistics

  ```
  
  ```

### Licensing

### DIFF
--- a/artiq/firmware/libboard_misoc/net_settings.rs
+++ b/artiq/firmware/libboard_misoc/net_settings.rs
@@ -1,13 +1,20 @@
 use core::fmt;
 use core::fmt::{Display, Formatter};
 use core::str::FromStr;
+use smoltcp::iface::{Interface, InterfaceBuilder};
+use smoltcp::phy::Device;
 
-use smoltcp::wire::{EthernetAddress, IpAddress, Ipv4Address};
+use smoltcp::wire::{EthernetAddress, IpAddress, IpCidr, Ipv4Address, Ipv4Cidr};
 
 use config;
 #[cfg(soc_platform = "kasli")]
 use i2c_eeprom;
 
+
+const IPV4_INDEX: usize = 0;
+const IPV6_LL_INDEX: usize = 1;
+const IPV6_INDEX: usize = 2;
+const IP_ADDRESS_STORAGE_SIZE: usize = 3;
 
 pub enum Ipv4AddrConfig {
     UseDhcp,
@@ -100,5 +107,35 @@ pub fn get_adresses() -> NetAddresses {
         ipv4_addr: ipv4_addr,
         ipv6_ll_addr: ipv6_ll_addr,
         ipv6_addr: ipv6_addr
+    }
+}
+
+pub trait InterfaceBuilderEx {
+    fn init_ip_addrs(self, net_addresses: &NetAddresses) -> Self;
+}
+
+impl<'a, DeviceT: for<'d> Device<'d>> InterfaceBuilderEx for InterfaceBuilder<'a, DeviceT> {
+    fn init_ip_addrs(self, net_addresses: &NetAddresses) -> Self {
+        let mut storage = [
+            IpCidr::new(IpAddress::Ipv4(Ipv4Address::UNSPECIFIED), 0);  IP_ADDRESS_STORAGE_SIZE
+        ];
+        if let Ipv4AddrConfig::Static(ipv4) = net_addresses.ipv4_addr {
+            storage[IPV4_INDEX] = IpCidr::new(IpAddress::Ipv4(ipv4), 0);
+        }
+        storage[IPV6_LL_INDEX] = IpCidr::new(net_addresses.ipv6_ll_addr, 0);
+        if let Some(ipv6) = net_addresses.ipv6_addr {
+            storage[IPV6_INDEX] = IpCidr::new(ipv6, 0);
+        }
+        self.ip_addrs(storage)
+    }
+}
+
+pub trait InterfaceEx {
+    fn update_ipv4_addr(&mut self, addr: &Ipv4Cidr);
+}
+
+impl<'a, DeviceT: for<'d> Device<'d>> InterfaceEx for Interface<'a, DeviceT> {
+    fn update_ipv4_addr(&mut self, addr: &Ipv4Cidr) {
+        self.update_ip_addrs(|storage| storage[IPV4_INDEX] = IpCidr::Ipv4(*addr))
     }
 }
--- a/artiq/firmware/libboard_misoc/net_settings.rs
+++ b/artiq/firmware/libboard_misoc/net_settings.rs
@@ -1,10 +1,13 @@
 use core::fmt;
 
-use smoltcp::wire::{EthernetAddress, IpAddress};
+use smoltcp::wire::{EthernetAddress, IpAddress, Ipv4Address};
 
 use config;
 #[cfg(soc_platform = "kasli")]
 use i2c_eeprom;
+
+
+pub const USE_DHCP: IpAddress = IpAddress::Ipv4(Ipv4Address::UNSPECIFIED);
 
 
 pub struct NetAddresses {
@@ -49,7 +52,13 @@ pub fn get_adresses() -> NetAddresses {
     }
 
     let ipv4_addr;
-    match config::read_str("ip", |r| r.map(|s| s.parse())) {
+    match config::read_str("ip", |r| r.map(|s| {
+        if s == "use_dhcp" {
+            Ok(USE_DHCP)
+        } else {
+            s.parse()
+        }
+    })) {
         Ok(Ok(addr)) => ipv4_addr = addr,
         _ => {
             #[cfg(soc_platform = "kasli")]
--- a/artiq/firmware/libboard_misoc/net_settings.rs
+++ b/artiq/firmware/libboard_misoc/net_settings.rs
@@ -60,16 +60,7 @@ pub fn get_adresses() -> NetAddresses {
         }
     })) {
         Ok(Ok(addr)) => ipv4_addr = addr,
-        _ => {
-            #[cfg(soc_platform = "kasli")]
-            { ipv4_addr = IpAddress::v4(192, 168, 1, 70); }
-            #[cfg(soc_platform = "sayma_amc")]
-            { ipv4_addr = IpAddress::v4(192, 168, 1, 60); }
-            #[cfg(soc_platform = "metlino")]
-            { ipv4_addr = IpAddress::v4(192, 168, 1, 65); }
-            #[cfg(soc_platform = "kc705")]
-            { ipv4_addr = IpAddress::v4(192, 168, 1, 50); }
-        }
+        _ => ipv4_addr = USE_DHCP,
     }
 
     let ipv6_ll_addr = IpAddress::v6(
--- a/artiq/firmware/libboard_misoc/net_settings.rs
+++ b/artiq/firmware/libboard_misoc/net_settings.rs
@@ -1,4 +1,6 @@
 use core::fmt;
+use core::fmt::{Display, Formatter};
+use core::str::FromStr;
 
 use smoltcp::wire::{EthernetAddress, IpAddress, Ipv4Address};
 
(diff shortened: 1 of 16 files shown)


sbourdeauducq:
> ### artiq_upload
> 
> artiq_upload.py can be used to upload firmware images to the bootloader.

We already have artiq_netboot. https://git.m-labs.hk/M-Labs/artiq-netboot

mbirtwell:
> We already have artiq_netboot. https://git.m-labs.hk/M-Labs/artiq-netboot

Ok, well I'll drop that commit then. I looked through the Utilities section of the docs when I was looking for something that did this. I wonder if there should be a mention of netboot there. I can open another PR for that.

mbirtwell:
@sbourdeauducq are you the right person to review this or should I ping someone else?

sbourdeauducq:
Let me stage first the smoltcp update alone on the master branch. Then we can test it for a little while and see if there are regressions.

sbourdeauducq:
How much testing have you done? What is the impact, if any, on the network performance metrics?

mbirtwell:
On Friday after reading your message I realised that there was another test I could try that I hadn't done when I submitted this but was now easy. Unfortunately when I tried running this I didn't get any network connectivity at all. Much debugging later I've found the following things:
* The failure mode is that some times when the board comes up it produces vast numbers of rx preamble errors and basically no packets make it through. I once saw one packet get through when it was in this state, but I think we can ignore that.
* I have 3 SFP modules available to me. When I use 2 of them I never see failures like this. When I use the 3rd I do see failures. The 2 that always work have slightly different markings on them to the one that occasionally fails. They're all "1000BASE-T SFP 100m" modules from FS, but they have different serial number prefixes and the ones that always work are labelled "1GT" and the one that fails is labelled "T".
* I see the failure significantly more frequently with my code changes. But I also see the failure occasionally with the current master branch. 
* I see the failure significantly more frequently from a warm start (artiq_flash start), than I do from a cold start (plugging in the power cable).
* Unplugging the ethernet cable and replugging it, almost always fixes the problem.
* Once it's working I've never seen it break with out a reboot.

But given all of that it doesn't seem like it's a good plan to merge this until we can find out what this issue is.

sbourdeauducq:
Those symptoms don't seem related to the TCP/IP stack.

mbirtwell:
I agree that they don't seem directly related to the TCP/IP stack. But there was quite a clear correlation that this happened more frequently with these code changes than without them, which suggests that against expectations these changes are influencing this in some way. I'd speculate that there's some change here (probably in smoltcp) which biases some race condition that we were already occasionally suffering from against us.

sbourdeauducq:
Are you sure you aren't fooling yourself? Could be a bitstream miscompilation (did it meet timing by the way?) or some intermittent hardware issue. I suggest compiling a fresh bitstream, possibly for another board or variant, and testing on another hardware setup.

sbourdeauducq (changes requested):
I tried the smoltcp update on the KC705.
I cannot reproduce the problems you report on Kasli, however the network breaks in other way, e.g. pulling the analyzer buffer freezes, and there are problems like:
```
ERROR: test_dma_trace (test_rtio.DMATest)
  File "/home/sb/artiq/artiq/test/coredevice/test_rtio.py", line 685, in test_dma_trace
    get_analyzer_dump(core_host)  # clear analyzer buffer
  File "/nix/store/gyhdm53q7y126nvidil871kzxlfak95s-python3-3.9.6-env/lib/python3.9/site-packages/artiq/coredevice/comm_analyzer.py", line 42, in get_analyzer_dump
    buf = sock.recv(8192)
ConnectionResetError: [Errno 104] Connection reset by peer

```

mbirtwell:
I was able to reproduce the hanging while pulling the analyzer buffer, this happening because the analyzer server thread just dropped the `TcpStream` which since my changes to upgrade smoltcp just removed the underlying `TcpSocket` immediately with out allowing any pending data to be sent. I've added support for sending pending data in drop. 

I wasn't able to reproduce any connection resets. The test that failed in your example, passes for me. 

As to the preamble errors I tried mixmatching bitstreams and firmware with and with out my changes and I got the following results:

| gateware | firmware | success | failure |
| ------------- | ------------- | ------------- | ------------- |
| dhcp | dhcp | 7 | 3 |
| base | base | 9 | 1 |
| dhcp | base | 10 | 0 |
| base | dhcp | 9 | 1 | 

Which actually was a lot less significant than I was expecting. As I say this only happens with one SFP module that I have, the others all work all the time. So perhaps we can ignore this.

sbourdeauducq:
As I asked before: What is the impact, if any, on the network performance metrics?
Have you run the full test suite?

mbirtwell:
Sorry, I meant to ask about this before. Do you have an existing test suite and performance benchmarks? How have you characterised network performance in the past? It would be nice if I could do something consistent with previous measurements.

sbourdeauducq:
artiq/test/coredevice + Hydra logs of the same
Includes some tests for the network and network speed.

mbirtwell:
With new smoltcp:
```
| Test                 | Mean (MiB/s) |  std (MiB/s) |
| -------------------- | ------------ | ------------ |
| I32 Array (1MB) H2D  |         2.03 |         0.04 |
| I32 Array (1MB) D2H  |         2.24 |         0.12 |
| I32 Array (1KB) H2D  |         0.42 |         0.03 |
| I32 Array (1KB) D2H  |         0.43 |         0.05 |
| Bytes List (1MB) H2D |         1.95 |         0.19 |
| Bytes List (1MB) D2H |         2.23 |         0.11 |
| Bytes List (1KB) H2D |         0.40 |         0.03 |
| Bytes List (1KB) D2H |         0.45 |         0.04 |
| Bytes (1MB) H2D      |         2.01 |         0.04 |
| Bytes (1MB) D2H      |         2.29 |         0.11 |
| Bytes (1KB) H2D      |         0.40 |         0.04 |
| Bytes (1KB) D2H      |         0.46 |         0.03 |
| I32 List (1MB) H2D   |         2.00 |         0.02 |
| I32 List (1MB) D2H   |         2.27 |         0.14 |
| I32 List (1KB) H2D   |         0.38 |         0.03 |
| I32 List (1KB) D2H   |         0.43 |         0.04 |

```

The current master:
```
| Test                 | Mean (MiB/s) |  std (MiB/s) |
| -------------------- | ------------ | ------------ |
| I32 Array (1MB) H2D  |         2.00 |         0.06 |
| I32 Array (1MB) D2H  |         2.30 |         0.00 |
| I32 Array (1KB) H2D  |         0.37 |         0.03 |
| I32 Array (1KB) D2H  |         0.41 |         0.04 |
| Bytes List (1MB) H2D |         2.01 |         0.02 |
| Bytes List (1MB) D2H |         2.23 |         0.05 |
| Bytes List (1KB) H2D |         0.38 |         0.02 |
| Bytes List (1KB) D2H |         0.42 |         0.02 |
| Bytes (1MB) H2D      |         1.99 |         0.04 |
| Bytes (1MB) D2H      |         2.29 |         0.04 |
| Bytes (1KB) H2D      |         0.41 |         0.02 |
| Bytes (1KB) D2H      |         0.46 |         0.03 |
| I32 List (1MB) H2D   |         1.95 |         0.03 |
| I32 List (1MB) D2H   |         2.28 |         0.04 |
| I32 List (1KB) H2D   |         0.37 |         0.03 |
| I32 List (1KB) D2H   |         0.41 |         0.03 |
```

That looks like no change to me. I wasn't able to find anywhere that these tests were run in Hydra. I found a couple of places where the artiq/test suite was run but in each of them all the coredevice tests were skipped.

sbourdeauducq:
The hardware tests are in a special ``extended-tests`` derivation.

artiq/firmware/runtime/main.rs
```diff
@@ -58,6 +59,14 @@ mod session;
 mod moninj;
 #[cfg(has_rtio_analyzer)]
 mod analyzer;
+mod dhcp;
+
+// Fixed indexes for the IP addresses so that they can be modified with some
+// semblance of confidence
```
sbourdeauducq, line 65: Why not use a Rust struct?

artiq/firmware/libboard_misoc/net_settings.rs
```diff
@@ -1,12 +1,15 @@
...
 
-use smoltcp::wire::{EthernetAddress, IpAddress};
+use smoltcp::wire::{EthernetAddress, IpAddress, Ipv4Address};
 
 use config;
 #[cfg(soc_platform = "kasli")]
 use i2c_eeprom;
 
 
+pub const USE_DHCP: IpAddress = IpAddress::Ipv4(Ipv4Address::UNSPECIFIED);
```
sbourdeauducq, line 10: Use ``Option`` or ``Enum``?

artiq/firmware/libboard_misoc/net_settings.rs
```diff
@@ -49,7 +52,13 @@ pub fn get_adresses() -> NetAddresses {
     }
 
     let ipv4_addr;
-    match config::read_str("ip", |r| r.map(|s| s.parse())) {
+    match config::read_str("ip", |r| r.map(|s| {
+        if s == "use_dhcp" {
+            Ok(USE_DHCP)
+        } else {
+            s.parse()
```
sbourdeauducq, line 59: If you use ``Enum`` above, you can just implement ``parse`` on it.

artiq/firmware/runtime/sched.rs
```diff
@@ -250,14 +250,26 @@ impl<'a> Io<'a> {
         })
     }
 
-    pub fn until<F: FnMut() -> bool>(&self, mut f: F) -> Result<(), Error> {
+    pub fn inner_until<F: FnMut() -> bool>(
```
sbourdeauducq, line 253: Wouldn't it be simpler to flush and close the socket in the core analyzer server? This way the ``Io`` abstraction layer would not deviate from the smoltcp API.

artiq/firmware/runtime/main.rs
mbirtwell: These indexes are used for the positions of these IP addresses with in `smoltcp::iface::interface::InterfaceInner::ip_addrs` which is a `ManagedSlice<'a, IpCidr>`. Using a struct unfortunately isn't an option, we would require smoltcp to be generic over the datastructure that it uses to store IP addresses.

sbourdeauducq: What about building the list from the struct just before passing it to smoltcp?

artiq/firmware/runtime/sched.rs
mbirtwell: That would work, I didn't have confidence that that would be the only place where this was needed, or that it wouldn't be a paper cut for others adding more features in future. I was preferring simplicity of use over simplicity of implementation. 

Whilst we're discussing this I think there's a third possible design here: we could have a task that owns closing sockets when a socket is dropped if it hasn't already been closed it could be added to a list of sockets so that that task can flush it and close it down. That'll make the Drop implementation simpler and mean that Drop won't block. But might be considered more complex over all.

sbourdeauducq: I tend to prefer explicit flush in the application - this is also how it is with a number of existing APIs such as Python asyncio.

mbirtwell: Sure, I'll make that change then

sbourdeauducq:
I'm still really not a fan of the ``pub const IPV4_INDEX`` and similar, which look dirty and error-prone - can we at least have a minimum of encapsulation them e.g. make them private and access them only through some functions in the same module?

Otherwise looks good, thanks @mbirtwell

mbirtwell:
> I'm still really not a fan of the `pub const IPV4_INDEX` and similar, which look dirty and error-prone - can we at least have a minimum of encapsulation them e.g. make them private and access them only through some functions in the same module?

I've had a go at that. I put the code in `net_settings` unfortunately I can't trivially use it from the bootloader also, because the bootloader doesn't have alloc and so needs to borrow the ip address storage. There are obviously ways around this but I didn't think it was worth it seen as the bootloader doesn't need to worry about which positions the IP addresses are in.

I've also used extension traits to make these functions look like members of the smoltcp Interface/Builder. This is particularly nice on the builder because it allows us to use the method chaining as before.

mbirtwell:
Oops that prevents that bootloader from compiling even when it's unused. I'll fix shortly.

sbourdeauducq:
There's the merge conflict, otherwise LGTM.

sbourdeauducq:
Merged manually

mbirtwell:
Thank you @sbourdeauducq

PR #1839, opened 2022-01-27
prioritize min as part of default value resolution in dashboard

### LINKED ISSUE #1811: NumberValue does not enforce default value min/max bounds
# Bug Report

## Issue Details

Consider the following simple experiment

```python
import artiq.language as aq


class TestIntParam(aq.EnvExperiment):
    def build(self):
        self.setattr_argument('int_param', aq.NumberValue(type="int",
                                                          ndecimals=0,
                                                          scale=1,
                                                          step=1,
                                                          min=1)
                             )

    def run(self):
        print(self.int_param)

```

### Expected Behavior

1. `int_param` doesn't have a default specified so should default to 1
2. The GUI should show a value of 1
3. The experiment should print "1"

The GUI shows 1, but the experiment prints 0. After clicking on the spin button to increase `int_param` to 2 and then decrease it back to 1, the experiment correctly prints "1". Recomputing arguments then restores the original (incorrect) behaviour, with the GUI showing "1" but the experiment printing "0". 

This behaviour is not exhibited if "1" is explicitly given as a default value for `int_param`

```python
import artiq.language as aq


class TestIntParam(aq.EnvExperiment):
    def build(self):
        self.setattr_argument('int_param', aq.NumberValue(default=1,
                                                          type="int",
                                                          ndecimals=0,
                                                          scale=1,
                                                          step=1,
                                                          min=1)
                             )

    def run(self):
        print(self.int_param)

```

stevefan1999:
# ARTIQ Pull Request

## Description of Changes

This solves an issue where the dashboard value will appear to go out of bound when the default initial value on dashboard refresh is less than the minimum value of the argument (not rendered correctly). This could be solved with a forced re-render, but it is much better to handle it by setting the default value to min if min is positive, and 0 if min is negative

### Related Issue

Closes #1811 

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

### Licensing

### DIFF
--- a/artiq/gui/entries.py
+++ b/artiq/gui/entries.py
@@ -102,10 +102,11 @@ class NumberEntryInt(QtWidgets.QSpinBox):
         else:
             have_max = "max" in procdesc and procdesc["max"] is not None
             have_min = "min" in procdesc and procdesc["min"] is not None
-            if have_max and have_min and procdesc["min"] <= 0 < procdesc["max"]:
-                return 0
+            if have_max and have_min:
+                if procdesc["min"] <= 0 < procdesc["max"]:
+                    return 0
             elif have_min and not have_max:
-                if procdesc["min"] > 0:
+                if procdesc["min"] >= 0:
                     return procdesc["min"]
             elif not have_min and have_max:
                 if procdesc["max"] < 0:
--- a/artiq/gui/entries.py
+++ b/artiq/gui/entries.py
@@ -99,7 +99,7 @@ class NumberEntryInt(QtWidgets.QSpinBox):
     def default_state(procdesc):
         if "default" in procdesc:
             return procdesc["default"]
-        elif "min" in procdesc and procdesc["min"]:
+        elif "min" in procdesc and procdesc["min"] is not None:
             # when the min value is negative, we clamp it to 0 instead
             return max(0, procdesc["min"])
         else:
--- a/artiq/gui/entries.py
+++ b/artiq/gui/entries.py
@@ -100,8 +100,7 @@ class NumberEntryInt(QtWidgets.QSpinBox):
         if "default" in procdesc:
             return procdesc["default"]
         elif "min" in procdesc and procdesc["min"] is not None:
-            # when the min value is negative, we clamp it to 0 instead
-            return max(0, procdesc["min"])
+            return procdesc["min"]
         else:
             return 0
 
--- a/artiq/gui/entries.py
+++ b/artiq/gui/entries.py
@@ -99,6 +99,9 @@ class NumberEntryInt(QtWidgets.QSpinBox):
     def default_state(procdesc):
         if "default" in procdesc:
             return procdesc["default"]
+        elif "min" in procdesc and procdesc["min"]:
+            # when the min value is negative, we clamp it to 0 instead
+            return max(0, procdesc["min"])
         else:
             return 0
 
--- a/artiq/gui/entries.py
+++ b/artiq/gui/entries.py
@@ -99,9 +99,17 @@ class NumberEntryInt(QtWidgets.QSpinBox):
     def default_state(procdesc):
         if "default" in procdesc:
             return procdesc["default"]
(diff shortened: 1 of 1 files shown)


artiq/gui/entries.py
```diff
@@ -99,6 +99,9 @@ def state_to_value(state):
     def default_state(procdesc):
         if "default" in procdesc:
             return procdesc["default"]
+        elif "min" in procdesc and procdesc["min"]:
+            # when the min value is negative, we clamp it to 0 instead
```
sbourdeauducq, line 103: Why?

artiq/gui/entries.py
```diff
@@ -99,6 +99,9 @@ def state_to_value(state):
     def default_state(procdesc):
         if "default" in procdesc:
             return procdesc["default"]
+        elif "min" in procdesc and procdesc["min"]:
```
sbourdeauducq, line 102: Why are you also testing for ``procdesc["min"] != 0``?

artiq/gui/entries.py
sbourdeauducq: Seems not just superfluous but also incorrect - the bug will still be present if both min and max are negative.

artiq/gui/entries.py
stevefan1999: oh, it is actually doing `procdesc["min"] is None`

sbourdeauducq: That's not how Python works. It would also test for ``!= 0``.

artiq/gui/entries.py
stevefan1999: hmm, we can just add an additional check whether 0 is within the range of min & max, but let say we just set the minimum value here then.

sbourdeauducq: What's special about 0? Why not e.g. return ``(min+max)/2`` if we have both, or just ``min`` or ``max`` if we have only one of them?

sbourdeauducq: And please do your research about ``/`` vs. ``//`` in Python and use them appropriately.

stevefan1999: using midpoint as the default value...feels bit odd though, should we do it?

artiq/gui/entries.py
stevefan1999: right. i didn't remember it is an Option[int] here, yet they overlapped when they shouldn't

artiq/gui/entries.py
sbourdeauducq: Can also keep 0 iff min <= 0 < max, then it's also backward compatible.

sbourdeauducq: Or just min instead of midpoint?

artiq/gui/entries.py
```diff
@@ -99,9 +99,17 @@ def state_to_value(state):
     def default_state(procdesc):
         if "default" in procdesc:
             return procdesc["default"]
-        elif "min" in procdesc and procdesc["min"] is not None:
-            return procdesc["min"]
         else:
+            have_max = "max" in procdesc and procdesc["max"] is not None
+            have_min = "min" in procdesc and procdesc["min"] is not None
```
sbourdeauducq, line 104: Are both of these situations (``None`` value and key not present) really legitimately occurring? Why?

artiq/gui/entries.py
```diff
@@ -99,9 +99,17 @@ def state_to_value(state):
...
+            if have_max and have_min and procdesc["min"] <= 0 < procdesc["max"]:
+                return 0
+            elif have_min and not have_max:
+                if procdesc["min"] > 0:
```
sbourdeauducq, line 108: ``min`` is inclusive (and ``max`` is exclusive). It doesn't change the behavior but ``>=`` would be clearer here.

sbourdeauducq: Other than that it looks good to me.

hartytp:
🎆

PR #1841, opened 2022-01-28
Backport of "fixes alignment and size problem" from artiq-zynq

### LINKED ISSUE #1810: RPC tuple size calculation is wrong
# Bug Report

## One-Line Summary

Tuple size calculation for RPC is wrong and can lead to invalid memory access.

### Steps to Reproduce
```python

class TestListTuple(EnvExperiment):
    def build(self):
        self.setattr_device("core")

    @rpc
    def fine(self) -> TList(TTuple([TFloat, TFloat])):
        return [(123.456, 123.456), (123.456, 123.456)]

    @rpc
    def error(self) -> TList(TTuple([TBool, TFloat, TBool])):
        return [(True, 123.456, True), (True, 123.456, True)]

    @kernel
    def run(self):
        print(self.fine()[0]) # OK
        print(self.error()[0]) # Error
        return 0
```

The error case would not finish, as it is accessing an incorrect memory address.

Affects artiq-zynq and riscv. Not tested for or1k but I expect the same issue.

### Cause
https://github.com/m-labs/artiq/blob/bbac4770926048b3c0b434aede371a63d198ecd5/artiq/firmware/libproto_artiq/rpc_proto.rs#L398-L406

The size calculation does not take into account of padding and alignment requirements. The case I use is specifically tailored to trigger this bug, as the alignment of TBool should be 1 byte and the alignment of TFloat should be 4/8 bytes, there should be padding in between.

stevefan1999:
# ARTIQ Pull Request

## Description of Changes

Backporting of https://git.m-labs.hk/M-Labs/artiq-zynq/pulls/160 to master

### Related Issue
Closes #1810

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

### Licensing

### DIFF
--- a/artiq/test/coredevice/test_embedding.py
+++ b/artiq/test/coredevice/test_embedding.py
@@ -538,23 +538,3 @@ class _Alignment(EnvExperiment):
 class AlignmentTest(ExperimentCase):
     def test_tuple(self):
         self.create(_Alignment).run()
-
-class _Alignment(EnvExperiment):
-    def build(self):
-        self.setattr_device("core")
-
-    @rpc
-    def a_tuple(self) -> TTuple([TInt64, TBool, TInt32]):
-        return 1234, True, 5678
-
-    @kernel
-    def run(self):
-        a, b, c = self.a_tuple()
-        assert a == 1234
-        assert b
-        assert c == 5678
-
-
-class AlignmentTest(ExperimentCase):
-    def test_tuple(self):
-        self.create(_Alignment).run()
--- a/artiq/test/coredevice/test_embedding.py
+++ b/artiq/test/coredevice/test_embedding.py
@@ -515,3 +515,23 @@ class NumpyBoolTest(ExperimentCase):
     def test_numpy_bool(self):
         """Test NumPy bools decay to ARTIQ compiler builtin bools as expected"""
         self.create(_NumpyBool).run()
+
+class _Alignment(EnvExperiment):
+    def build(self):
+        self.setattr_device("core")
+
+    @rpc
+    def a_tuple(self) -> TTuple([TInt64, TBool, TInt32]):
+        return 1234, True, 5678
+
+    @kernel
+    def run(self):
+        a, b, c = self.a_tuple()
+        assert a == 1234
+        assert b
+        assert c == 5678
+
+
+class AlignmentTest(ExperimentCase):
+    def test_tuple(self):
+        self.create(_Alignment).run()
--- a/artiq/test/coredevice/test_embedding.py
+++ b/artiq/test/coredevice/test_embedding.py
@@ -516,6 +516,29 @@ class NumpyBoolTest(ExperimentCase):
         """Test NumPy bools decay to ARTIQ compiler builtin bools as expected"""
         self.create(_NumpyBool).run()
 
+
+class _Alignment(EnvExperiment):
+    def build(self):
+        self.setattr_device("core")
+
+    @rpc
+    def a_tuple(self) -> TList(TTuple([TBool, TFloat, TBool])):
+        return [(True, 1234.5678, True)]
+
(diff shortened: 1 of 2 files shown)


sbourdeauducq:
What testing have you done?

pca006132:
Looks fine to me, but maybe we can put `ptr.neg().rem_euclid(alignment)` in a separate function? This may not be obvious to others that this is calculating the alignment requirement.

We can merge it if it is tested with the example in #1810

stevefan1999:
@pca006132 yep, exactly did it, however I think this should be put in the test suite

pca006132:
> @pca006132 yep, exactly did it, however I think this should be put in the test suite

Please put the test into the test suite as well, in the test portability script. Thanks.

stevefan1999:
This is interesting:
```py3
class _Alignment(EnvExperiment):
    def build(self):
        self.setattr_device("core")

    @rpc
    def a_list_of_tuple(self) -> TList(TTuple([TBool, TFloat, TBool])):
        return [(True, 123.456, True), (True, 123.456, True)]

    @kernel
    def run(self):
        a,_,_ = self.a_list_of_tuple()[0]
        assert a


class AlignmentTest(ExperimentCase):
    def test_tuple(self):
        """Test NumPy bools decay to ARTIQ compiler builtin bools as expected"""
        self.create(_Alignment).run()
```

```
<artiq>/test/coredevice/test_embedding.py:529:17-529:42: error: cannot unify list(elt=(bool, float, float)) with list(elt=(bool, float, bool)): float is incompatible with bool
        a,_,_ = self.a_list_of_tuple()[0]
                ^^^^^^^^^^^^^^^^^^^^^^^^^
<artiq>/test/coredevice/test_embedding.py:529:17-529:42: note: expression of type list(elt=(bool, float, float))
        a,_,_ = self.a_list_of_tuple()[0]
                ^^^^^^^^^^^^^^^^^^^^^^^^^
<artiq>/test/coredevice/test_embedding.py:529:17-529:39: note: expression of type list(elt=(bool, float, bool))
        a,_,_ = self.a_list_of_tuple()[0]
                ^^^^^^^^^^^^^^^^^^^^^^   
```
The compiler might be confused too, but just access the value separately works:
```py3
    @kernel
    def run(self):
        x = self.a_list_of_tuple()
        a,b,c = x[0]
        assert a
```
cc @pca006132

pca006132:
> In a case pattern within a match statement, `_` is a soft keyword that denotes a wildcard.
> 
> Separately, the interactive interpreter makes the result of the last evaluation available in the variable `_`. (It is stored in the builtins module, alongside built-in functions like print.)
> 
> Elsewhere, `_` is a regular identifier. It is often used to name “special” items, but it is not special to Python itself.

So, it is just a regular identifier. This is not a bug.

pca006132:
Are you sure the test case will fail without your patch? Your test case is not what I wrote in the issue, so I am not sure without actually running it. If yes, this can be merged directly.

stevefan1999:
> Are you sure the test case will fail without your patch? Your test case is not what I wrote in the issue, so I am not sure without actually running it. If yes, this can be merged directly.

Oh yes, it is problematic, you need to generate a case of misalignment which I didn't and is probably optimized by the compiler

artiq/test/coredevice/test_embedding.py
```diff
@@ -515,3 +515,46 @@ class NumpyBoolTest(ExperimentCase):
...
+        assert b == e
+        assert c == f
+        return 0
+
+
+class AlignmentTest(ExperimentCase):
+    def test_tuple(self):
+        self.create(_Alignment).run()
+
+class _Alignment(EnvExperiment):
```
pca006132, line 542: Why there are classes with the same name?

pca006132 (approved):

PR #1849, opened 2022-02-16
latexpdf: add some line breaks to change poor formatting

### LINKED ISSUE #1846: poor PDF formatting
image

ciciwu213:
closes #1846

### DIFF
--- a/artiq/frontend/artiq_flash.py
+++ b/artiq/frontend/artiq_flash.py
@@ -13,7 +13,6 @@ from collections import defaultdict
 from sipyco import common_args
 
 from artiq import __version__ as artiq_version
-from artiq import __artiq_dir__ as artiq_dir
 from artiq.remoting import SSHClient, LocalClient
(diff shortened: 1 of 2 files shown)


sbourdeauducq:
Why was this not word-wrapped like the other text in the first place and why is the solution to add ``|``? Isn't that table syntax?

ciciwu213:
> Why was this not word-wrapped like the other text in the first place and why is the solution to add `|`? Isn't that table syntax?

I think the reason why is there are so many words in the first line that it treated ``*`` in the next line as an end-string of the whole paragraph, then all the words are emphasized. Then I tried to do some changes to the file (if deleting ``*`` in the second line, the poor formatting in the first line will disappear) and found the real solution is actually the blank line between the first line and the second line so that ``*`` can be distinguished, and ``|`` is only used for keeping line breaks. I saw that usage from line blocks in Sphinx Documentation and StackOverflow website.

PR #1854, opened 2022-02-22
gateware: expand monitorable value to at least 32-bit

stevefan1999:
# ARTIQ Pull Request

## Description of Changes

Set monitorable value to at least 32-bit 

### Related Issue

Urukul monitor achieves monitoring by sniffing SPI bus, which is 32-bit in general. We should make it explicit that we monitor at least 32-bits. 

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

### Licensing

### DIFF
--- a/artiq/firmware/runtime/moninj.rs
+++ b/artiq/firmware/runtime/moninj.rs
@@ -18,7 +18,7 @@ mod local_moninj {
             csr::rtio_moninj::mon_chan_sel_write(channel as _);
             csr::rtio_moninj::mon_probe_sel_write(probe);
             csr::rtio_moninj::mon_value_update_write(1);
-            csr::rtio_moninj::mon_value_read() as u32
+            csr::rtio_moninj::mon_value_read()
         }
     }
 
@@ -55,7 +55,7 @@ mod remote_moninj {
     use sched::{Io, Mutex};
 
     pub fn read_probe(io: &Io, aux_mutex: &Mutex, linkno: u8, destination: u8, channel: u16, probe: u8) -> u32 {
-        let reply = drtio::aux_transact(io, aux_mutex, linkno, &drtioaux::Packet::MonitorRequest { 
+        let reply = drtio::aux_transact(io, aux_mutex, linkno, &drtioaux::Packet::MonitorRequest {
             destination: destination,
             channel: channel,
             probe: probe
--- a/artiq/firmware/runtime/moninj.rs
+++ b/artiq/firmware/runtime/moninj.rs
@@ -18,7 +18,7 @@ mod local_moninj {
             csr::rtio_moninj::mon_chan_sel_write(channel as _);
             csr::rtio_moninj::mon_probe_sel_write(probe);
             csr::rtio_moninj::mon_value_update_write(1);
-            csr::rtio_moninj::mon_value_read()
+            csr::rtio_moninj::mon_value_read() as _
         }
     }
 
--- a/artiq/firmware/runtime/moninj.rs
+++ b/artiq/firmware/runtime/moninj.rs
@@ -13,7 +13,7 @@ use board_artiq::drtio_routing;
 mod local_moninj {
     use board_misoc::csr;
 
-    pub fn read_probe(channel: u16, probe: u8) -> u32 {
+    pub fn read_probe(channel: u16, probe: u8) -> u64 {
         unsafe {
             csr::rtio_moninj::mon_chan_sel_write(channel as _);
             csr::rtio_moninj::mon_probe_sel_write(probe);
@@ -41,7 +41,7 @@ mod local_moninj {
 
 #[cfg(not(has_rtio_moninj))]
 mod local_moninj {
-    pub fn read_probe(_channel: u16, _probe: u8) -> u32 { 0 }
+    pub fn read_probe(_channel: u16, _probe: u8) -> u64 { 0 }
 
     pub fn inject(_channel: u16, _overrd: u8, _value: u8) { }
 
@@ -54,7 +54,7 @@ mod remote_moninj {
     use rtio_mgt::drtio;
     use sched::{Io, Mutex};
 
-    pub fn read_probe(io: &Io, aux_mutex: &Mutex, linkno: u8, destination: u8, channel: u16, probe: u8) -> u32 {
+    pub fn read_probe(io: &Io, aux_mutex: &Mutex, linkno: u8, destination: u8, channel: u16, probe: u8) -> u64 {
         let reply = drtio::aux_transact(io, aux_mutex, linkno, &drtioaux::Packet::MonitorRequest {
             destination: destination,
             channel: channel,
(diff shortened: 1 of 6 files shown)


sbourdeauducq:
What's the plan for monitoring the AD9912 and its 48-bit FTW?
Why not u64?

sbourdeauducq:
> Urukul monitor achieves monitoring by sniffing SPI bus, which is 32-bit in general

You need to look at the commands that write the FTW, decode the FTW, and buffer it with IO_UPDATE.

See https://github.com/m-labs/artiq/blob/master/artiq/gateware/rtio/phy/dds.py for an example.
FUD (Frequency UpDate) on the NIST AD9914 cards that this PHY targets is the equivalent of IO_UPDATE.
By the way:
Why is this patch necessary?
AD9914 monitoring with a 32-bit FTW works fine without it.

stevefan1999:
> What's the plan for monitoring the AD9912 and its 48-bit FTW? Why not u64?

IRRC from @b-bondurant, who did both AD9910 and AD9912 monitoring, also implemented AD9912 monitoring by sniffing SPI bus twice. (also, the monitor implementation is shared with AD9910)

sbourdeauducq:
His implementation isn't necessarily free from bugs, in fact he said it wasn't finished. My question still stands, in particular wrt atomicity.

b-bondurant:
@stevefan1999 yeah I believe that's what I did.

@sbourdeauducq IIRC, I had monitoring working fairly well, but it was far from well-tested (I think I pretty much just tweaked things by hand and then confirmed that the value showed up in the dashboard, and had no DRTIO testing). The real trouble I ran into was when I tried to implement injection, which I foolishly didn't put in a separate branch. Expanding the rust runtime to handle 32-bit injection words rather than 8-bit words proved too much for me to handle at the time.

stevefan1999:
> You need to look at the commands that write the FTW, decode the FTW, and buffer it with IO_UPDATE.

By that you mean this right?
https://github.com/m-labs/artiq/blob/17ecd355309600b997e303b41efb3d33af33c3dc/artiq/coredevice/ad9912.py#L47-L63

And if we trace a little up:
https://github.com/m-labs/artiq/blob/17ecd355309600b997e303b41efb3d33af33c3dc/artiq/coredevice/ad9912.py#L158-L178

Seems like we are trapped with a 4 bytes here and on, that's why for the last few days when I switch it to 48-bit monitor size it does nothing but just fills 32-bit. I have yet to find a better strategy to handle this.

sbourdeauducq:
It's really not that complicated. Snoop on the multiple SPI transfers, assemble/store the 48 bits in a backbuffer in gateware (like the AD9914 monitor does), and transfer the backbuffer (atomically) to the 48-bit probe when IO_UPDATE is pulsed. The backbuffer cannot and does not need to update atomically.

sbourdeauducq:
This is just the same way the DDS chip works, obviously we don't want output glitches due to non-atomic FTW transfer...

stevefan1999:
> It's really not that complicated. Snoop on the multiple SPI transfers, assemble/store the 48 bits in a backbuffer in gateware (like the AD9914 monitor does), and transfer the backbuffer (atomically) to the 48-bit probe when IO_UPDATE is pulsed. The backbuffer cannot and does not need to update atomically.

That sounds like what I planned to do before I scrapped it. Okay, I'll try it again.

sbourdeauducq:
> By the way:
> Why is this patch necessary?
> AD9914 monitoring with a 32-bit FTW works fine without it.

@stevefan1999 You didn't answer this question when your patch was for 32 bits. This does not inspire confidence that you are doing the right thing.

What tests did you do for this new "48-bit" patch?

artiq/firmware/runtime/moninj.rs
```diff
@@ -55,7 +55,7 @@ mod remote_moninj {
     use sched::{Io, Mutex};
 
     pub fn read_probe(io: &Io, aux_mutex: &Mutex, linkno: u8, destination: u8, channel: u16, probe: u8) -> u32 {
-        let reply = drtio::aux_transact(io, aux_mutex, linkno, &drtioaux::Packet::MonitorRequest { 
+        let reply = drtio::aux_transact(io, aux_mutex, linkno, &drtioaux::Packet::MonitorRequest {
```
sbourdeauducq, line 58: Remove this change.

artiq/firmware/runtime/moninj.rs
```diff
@@ -18,7 +18,7 @@ mod local_moninj {
             csr::rtio_moninj::mon_chan_sel_write(channel as _);
             csr::rtio_moninj::mon_probe_sel_write(probe);
             csr::rtio_moninj::mon_value_update_write(1);
-            csr::rtio_moninj::mon_value_read() as u32
+            csr::rtio_moninj::mon_value_read()
```
sbourdeauducq, line 21: I can't see how this possibly works. ``read_probe()`` stll returns ``u32``.

artiq/gateware/rtio/moninj.py
```diff
@@ -12,7 +12,7 @@ def __init__(self, channels):
         self.chan_sel = CSRStorage(bits_for(len(chan_probes)-1))
         self.probe_sel = CSRStorage(bits_for(max_chan_probes-1))
         self.value_update = CSR()
-        self.value = CSRStatus(max_probe_len)
+        self.value = CSRStatus(max(48, max_probe_len))
```
sbourdeauducq, line 15: Nonsense.

sbourdeauducq: The ``max_probe_len`` computation already takes care of large probes.

stevefan1999:
> > By the way:
> > Why is this patch necessary?
> > AD9914 monitoring with a 32-bit FTW works fine without it.
> 
> @stevefan1999 You didn't answer this question when your patch was for 32 bits. This does not inspire confidence that you are doing the right thing.
> 
> What tests did you do for this new "48-bit" patch?

I know that this size is influenced by `max_probe_len` which in turn is influenced by the max size of all monitor probe signals, so this way we can be sure the CSR size won't suddenly shrink down causing a firmware compilation error (it is possible for end user to remove some modules if they deemed that unnecessary, that would also touch the probe CSR size). It also served as a conservative move to make sure that all the original stuff works when monitorable values are expanded to at least 48-bit, for example, I spotted some issue when I expanded the injection probe size to 32-bit regarding endianness.

Another way though: always convert the CSR monitor probe value to 64-bit in software (it should be zero-cost on LE), so this way we won't need to touch moninj max probe anymore. Let's see if it works.

sbourdeauducq:
It wouldn't cause a compilation error if you had not removed the ``as uXX`` type conversions from the code. The situation is the same with systems with AD9914 vs. without AD9914.

sbourdeauducq:
> I spotted some issue when I expanded the injection probe size to 32-bit regarding endianness

Why would you do that? Any injection is outside the scope of your task.

And you still have not answered the question.

artiq/firmware/runtime/moninj.rs
stevefan1999: seems like this line is not formatted properly, the space shouldn't be there

sbourdeauducq: As I told you multiple times already, style-only changes should not be bundled with other things.

sbourdeauducq:
Done and also for Zynq.

PR #1872, opened 2022-03-15
urukul: ad9910 & ad9912 monitoring

stevefan1999:
# ARTIQ Pull Request

## Description of Changes

This patch implements SPI sniffing of RTIO commands of AD9910 and AD9912. 

This exports four sets of two probes (register, value/ftw) linearly. This can be inspected using https://github.com/m-labs/artiq/pull/1875. Register is flatten first then value. As an example the layout for the probes of 4 channels is therefore:
```
Probe 0: Channel 0 value
Probe 1: Channel 1 value
Probe 2: Channel 2 value
Probe 3: Channel 3 value
...
Probe n: Channel n value
```
Respectively.

(Works on AD9912 so far)

## Type of Changes

|   | Type |

### All Pull Requests

- [ ] Close/update issues.

### Code Changes

### Documentation Changes

### Git Logistics

  ```
  
  ```

### Licensing


artiq/coredevice/comm_moninj.py
```diff
@@ -84,9 +84,8 @@ async def _receive_cr(self):
                 if not ty:
                     return
                 if ty == b"\x00":
-                    payload = await self._reader.readexactly(9)
-                    channel, probe, value = struct.unpack(
-                        self.endian + "lbl", payload)
+                    payload = await self._reader.readexactly(13)
+                    channel, probe, value = struct.unpack(self.endian + "lbq", payload)
```
sbourdeauducq, line 88: How does this belong in this PR?

sbourdeauducq: I thought we had agreed a long time ago on how to structure this.

artiq/gateware/rtio/phy/urukul_monitor.py
```diff
@@ -0,0 +1,133 @@
...
+        )
+
+        self.cs = Signal(8)
+        self.length = Signal(8)
+        self.flags = Signal(8)
+
+        self.sync.rio += If(
+            self.rtlink.o.stb & (self.current_address == SPI_CONFIG_ADDR),
+            self.cs.eq(self.current_data[24:]),
+            # self.div.eq(self.current_data[16:24] + 2),
```
sbourdeauducq, line 26: Remove commented-out code.

artiq/gateware/rtio/phy/urukul_monitor.py
```diff
@@ -0,0 +1,133 @@
...
+            self.length.eq(self.current_data[8:16] + 1),
+            self.flags.eq(self.current_data[0:8])
+        )
+
+    def selected(self, c):
+        if self.cs == CS_DDS_MULTI:
+            return True
```
sbourdeauducq, line 33: I suspect this doesn't do what you expect. ``if`` is always executed in the interpreter. You may need ``If``.

artiq/gateware/rtio/phy/urukul_monitor.py
```diff
@@ -0,0 +1,133 @@
...
+    def __init__(self, spi_rtlink):
+        self.rtlink = spi_rtlink
+
+        self.current_address = Signal.like(self.rtlink.o.address)
+        self.current_data = Signal.like(self.rtlink.o.data)
+        self.sync.rio += If(
+            self.rtlink.o.stb,
+            self.current_address.eq(self.rtlink.o.address),
+            self.current_data.eq(self.rtlink.o.data)
+        )
```
sbourdeauducq, line 17: What is the purpose of this?

artiq/gateware/rtio/phy/urukul_monitor.py
```diff
@@ -0,0 +1,133 @@
+from migen import *
+
+from artiq.coredevice.spi2 import SPI_CONFIG_ADDR, SPI_DATA_ADDR, SPI_END, SPI_INPUT
+from artiq.coredevice.urukul import CS_DDS_CH0, CS_DDS_MULTI
+
+
+class AD99XXMonitorGeneric(Module):
+    def __init__(self, spi_rtlink):
+        self.rtlink = spi_rtlink
```
sbourdeauducq, line 9: consistency

artiq/coredevice/comm_moninj.py
stevefan1999: hmm, since the byte sizes are expanded in firmware, without this line artiq_rtiomon and subsequently artiq_dashboard will not work. Should we further cherrypick this away?

sbourdeauducq: It's OK for a PR to depend on another one. And anyway I was expecting the probe width expansion one to be straightforward and easily merged but obviously that was wrong...

artiq/gateware/rtio/phy/urukul_monitor.py
```diff
@@ -0,0 +1,133 @@
...
+                                state.eq(1)
+                            ])
+                        ],
+                        1: [
+                            If(self.flags & SPI_END, buffer[i]['value'].eq(self.current_data))
+                        ]
+                    })
+                ])
+
+        self.sync.rio_phy += If(phy.rtlink.o.stb, [
```
sbourdeauducq, line 77: Isn't stb asserted twice for one pulse?

artiq/gateware/rtio/phy/urukul_monitor.py
```diff
@@ -0,0 +1,133 @@
...
+            return True
+        else:
+            return c == self.cs
+
+    def master_is_data_and_not_input(self):
+        return (self.current_address == SPI_DATA_ADDR) & ~(self.flags & SPI_INPUT)
+
+
+class AD9910Monitor(AD99XXMonitorGeneric):
+    def __init__(self, spi_rtlink, phy, nchannels=4):
```
sbourdeauducq, line 42: Can't you see anything wrong with those first two parameters?

artiq/gateware/rtio/phy/urukul_monitor.py
```diff
@@ -86,7 +78,7 @@ def __init__(self, spi_rtlink, phy, nchannels=4):
 
 
 class AD9912Monitor(AD99XXMonitorGeneric):
-    def __init__(self, spi_rtlink, phy, nchannels=4):
+    def __init__(self, rtlink, phy, nchannels=4):
```
sbourdeauducq, line 81: Try harder.

artiq/firmware/runtime/moninj.rs
```diff
@@ -13,12 +13,12 @@ use board_artiq::drtio_routing;
...
     use board_misoc::csr;
 
-    pub fn read_probe(channel: u16, probe: u8) -> u32 {
+    pub fn read_probe(channel: u16, probe: u8) -> u64 {
         unsafe {
             csr::rtio_moninj::mon_chan_sel_write(channel as _);
             csr::rtio_moninj::mon_probe_sel_write(probe);
             csr::rtio_moninj::mon_value_update_write(1);
-            csr::rtio_moninj::mon_value_read() as u32
+            csr::rtio_moninj::mon_value_read()
```
sbourdeauducq, line 21: As mentioned before ``as u64`` is necessary. Are you paying attention?

sbourdeauducq: And, one more time: this should be in the other PR.

artiq/gateware/rtio/phy/urukul_monitor.py
stevefan1999: One for high and one for low, so should we look for high?

stevefan1999: added a check for high, but this works only if `io_update` is TTL based. Need to handle `cfg_reg` based `io_update` as well.

sbourdeauducq:
Is this tested?

sbourdeauducq (changes requested):
Doesn't use FTW therefore obviously broken.

stevefan1999:
> I don't see the address of the FTW register. How is this supposed to work?

You don't really see that. It is extracted from the SPI command when the user sends using the API like here:

https://github.com/m-labs/artiq/blob/f91e1065862e38009a6430343f01c54faddaea42/artiq/coredevice/ad9912.py#L158-L178

and here:

https://github.com/m-labs/artiq/blob/f91e1065862e38009a6430343f01c54faddaea42/artiq/coredevice/ad9910.py#L643-L649

https://github.com/m-labs/artiq/blob/f91e1065862e38009a6430343f01c54faddaea42/artiq/coredevice/ad9910.py#L234-L257

The rest will be handled by user. You can use the rtiomon patch to query the probes:
```bash
$ artiq_rtiomon 192.168.1.70 0 --monitors 0-10
[monitor] 0x000000[0]: 0
[monitor] 0x000000[1]: 429
[monitor] 0x000000[2]: 0
[monitor] 0x000000[3]: 0
[monitor] 0x000000[4]: 0
[monitor] 0x000000[5]: 38280596832649
[monitor] 0x000000[6]: 0
[monitor] 0x000000[7]: 0
[monitor] 0x000000[5]: 84442493013197
[monitor] 0x000000[5]: 83879543059775
[monitor] 0x000000[5]: 83316593106354
[monitor] 0x000000[5]: 82753643152933
[monitor] 0x000000[5]: 82190693199512
[monitor] 0x000000[5]: 81627743246090
[monitor] 0x000000[5]: 81064793292669
[monitor] 0x000000[5]: 80501843339248
[monitor] 0x000000[5]: 79938893385826
[monitor] 0x000000[5]: 79375943432405
[monitor] 0x000000[5]: 78812993478984
[monitor] 0x000000[5]: 78250043525562
[monitor] 0x000000[5]: 77687093572141
[monitor] 0x000000[5]: 77124143618720
[monitor] 0x000000[5]: 76561193665298
[monitor] 0x000000[5]: 75998243711877
[monitor] 0x000000[5]: 75435293758456
[monitor] 0x000000[5]: 74872343805035
[monitor] 0x000000[5]: 74309393851613
[monitor] 0x000000[5]: 73746443898192
[monitor] 0x000000[5]: 73183493944771
[monitor] 0x000000[5]: 72620543991349
[monitor] 0x000000[5]: 72057594037928
[monitor] 0x000000[5]: 71494644084507
[monitor] 0x000000[5]: 70931694131085
[monitor] 0x000000[5]: 70368744177664
[monitor] 0x000000[5]: 69805794224243
[monitor] 0x000000[5]: 69242844270821
[monitor] 0x000000[5]: 68679894317400
[monitor] 0x000000[5]: 68116944363979
[monitor] 0x000000[5]: 67553994410557
[monitor] 0x000000[5]: 66991044457136
[monitor] 0x000000[5]: 66428094503715
```
We could be more specific to monitor certain registers only, for example in this case 429 (aka 0x1ad, AD9912_POW1), but this should better be handled by software.

sbourdeauducq:
> this should better be handled by software.

I don't think so. Storing the register address and/or more than one register wastes FPGA resources, and additionally it will not work if you write more registers than there are storage slots. Currently you only have one storage slot per channel, so a simple frequency control command followed by e.g. amplitude control in every channel (which is a common use case) will break your naive code, as the FTW information is lost. Software will not be able to capture it reliably before it is overwritten. Your proposal uses lots of FPGA resources and cannot work reliably; bottom line: it is broken.

I had told you, on multiple occasions, to follow the AD9914 monitor which does [filter FTW in gateware](https://github.com/m-labs/artiq/blob/master/artiq/gateware/rtio/phy/dds.py#L41). You should have studied it carefully and paid attention to that. Just like most of ARTIQ, every line of code there does not come from random decisions and is written like that for good reasons.

stevefan1999:
> > this should better be handled by software.
> 
> I don't think so. Storing the register address and/or more than one register wastes FPGA resources, and additionally it will not work if you write more registers than there are storage slots. Currently you only have one storage slot per channel, so a simple frequency control command followed by e.g. amplitude control in every channel (which is a common use case) will break your naive code, as the FTW information is lost. Software will not be able to capture it reliably before it is overwritten. Your proposal uses lots of FPGA resources and (unless you store every register at greatly increased resource usage) cannot work reliably; bottom line: it is broken.
> 
> I had told you, on multiple occasions, to follow the AD9914 monitor which does [filter FTW in gateware](https://github.com/m-labs/artiq/blob/master/artiq/gateware/rtio/phy/dds.py#L41). You should have studied it carefully and paid attention to that. Just like most of ARTIQ, every line of code there does not come from random decisions and is written like that for good reasons.

I do studied all DDS monitor code, and I thought the way AD9914 did would be inelegant...Okay, although adding a filter for AD9910 would be trivial, but for AD9912 in the general case:

https://github.com/m-labs/artiq/blob/f91e1065862e38009a6430343f01c54faddaea42/artiq/coredevice/ad9912.py#L171-L177

Seems like we need to find whether the command would overlap-write over there (in this case it write back from AD9912_POW1/0x1ad to AD9912_FTW0/0x1a6).

sbourdeauducq:
> I do studied all DDS monitor code, and I thought the way AD9914 did would be inelegant...Okay, although adding a filter for AD9910 would be trivial

Since you were unable to come up with a correct/working solution, perhaps those things are neither "inelegant" nor "trivial".

stevefan1999:
> amplitude control in every channel (which is a common use case) will break your naive code

I assume this would only happen on AD9910, what else should I test for it?

artiq/gateware/rtio/phy/urukul_monitor.py
```diff
@@ -0,0 +1,125 @@
...
+        self.sync.rio_phy += If(self.is_io_update(), [
+            If(self.current_address == SPI_DATA_ADDR, [update_probe_data(i) for i in range(nchannels)]),
+            [state[i].eq(0) for i in range(nchannels)]
+        ])
+
+
+class AD9912Monitor(AD99XXMonitorGeneric):
+    def __init__(self, spi_phy, io_update_phy, nchannels=4):
+        data = [Signal(48) for i in range(nchannels)]
+        buffer = [Signal(64) for i in range(nchannels)]
```
sbourdeauducq, line 89: 64?

artiq/test/coredevice/test_moninj.py
```diff
@@ -1,8 +1,28 @@
...
+import numpy
+from contextlib import asynccontextmanager
+from unittest import IsolatedAsyncioTestCase
 
 from artiq.coredevice.comm_moninj import *
 from artiq.test.hardware_testbench import ExperimentCase
+from artiq.experiment import *
+
+
+def async_test(coro):
```
sbourdeauducq, line 13: For the 238th time please follow existing coding patterns. If you want to introduce (justified) new ones do it consistently throughout the entire codebase.

artiq/test/coredevice/test_moninj.py
```diff
@@ -1,8 +1,28 @@
...
+        loop = asyncio.new_event_loop()
+        try:
+            return loop.run_until_complete(coro(*args, **kwargs))
+        finally:
+            loop.close()
+
+    return wrapper
+
+
+def get_last_integers(str):
```
sbourdeauducq, line 24: Use the channel numbers in the device instead of relying on an arbitrary device db naming scheme.

artiq/test/coredevice/test_moninj.py
```diff
@@ -66,3 +86,175 @@ def injection_status_cb(channel, override, value):
...
+    def setUp(self):
+        super().setUp()
+        try:
+            self.core = self.device_mgr.get_desc("core")
+            self.urukuls = self.get_urukuls()
+            self.kernel = self.create(_UrukulExperiment)
+        except KeyError as e:
+            # skip if ddb does not match requirements
+            raise unittest.SkipTest(
+                "test device not available: `{}`".format(*e.args))
```
sbourdeauducq, line 145: In what situations would this trigger? Are all of them appropriate? What problem does this solve?

artiq/gateware/rtio/phy/urukul_monitor.py
```diff
@@ -0,0 +1,125 @@
...
+        # shifted 8 bits left for 32-bit bus
+        dest_cfg_reg_and_io_update_bit_set = (self.cs == CS_CFG) & self.current_data[8 + CFG_IO_UPDATE]
+        if self.io_update_phy is not None and self.io_update_phy.rtlink is not None:
+            io_update_strobe_set_and_high = self.io_update_phy.rtlink.o.stb & self.io_update_phy.rtlink.o.data
+            return io_update_strobe_set_and_high | dest_cfg_reg_and_io_update_bit_set
+        return dest_cfg_reg_and_io_update_bit_set
+
+
+class AD9910Monitor(AD99XXMonitorGeneric):
+    _AD9910_REG_FTW = 0x07
```
sbourdeauducq, line 42: Just import it from ``coredevice.ad9910`` like you did for AD9912.

artiq/gateware/rtio/phy/urukul_monitor.py
```diff
@@ -0,0 +1,125 @@
...
+                data[i].eq(buffer[i])
+            ])
+
+        # 0 -> init, 1 -> start reading
+        state = [Signal(1) for i in range(nchannels)]
+        for i in range(nchannels):
+            self.sync.rio_phy += If(self.selected(i + CS_DDS_CH0) & (self.current_address == SPI_DATA_ADDR), [
+                Case(state[i], {
+                    0: [
+                        # Bit 7: read op
```
sbourdeauducq, line 63: What does that mean?

artiq/gateware/rtio/phy/urukul_monitor.py
```diff
@@ -0,0 +1,125 @@
...
+            ])
+        ])
+
+    def selected(self, c):
+        return (self.cs == CS_DDS_MULTI) | (self.cs == c)
+
+    def is_io_update(self):
+        # shifted 8 bits left for 32-bit bus
+        dest_cfg_reg_and_io_update_bit_set = (self.cs == CS_CFG) & self.current_data[8 + CFG_IO_UPDATE]
+        if self.io_update_phy is not None and self.io_update_phy.rtlink is not None:
```
sbourdeauducq, line 35: Why are you testing for ``self.io_update_phy.rtlink is not None``?

artiq/gateware/rtio/phy/urukul_monitor.py
```diff
@@ -0,0 +1,125 @@
...
+                self.cs.eq(self.current_data[24:]),
+                self.length.eq(self.current_data[8:16] + 1),
+                self.flags.eq(self.current_data[0:8])
+            ])
+        ])
+
+    def selected(self, c):
+        return (self.cs == CS_DDS_MULTI) | (self.cs == c)
+
+    def is_io_update(self):
```
sbourdeauducq, line 32: Reorganize that method, it's a total mess for what it does.

artiq/test/coredevice/test_moninj.py
stevefan1999: hmm, i suppose you mean this channel number in `arguments`:
```python
device_db["ttl_urukul0_sw0"] = {
    "type": "local",
    "module": "artiq.coredevice.ttl",
    "class": "TTLOut",
    "arguments": {"channel": 0x000002}
}
```
It doesn't seem accurate so that's why I used a regex to extract the last sequence of numbers instead, or should we take account of this offset in the gateware?

artiq/test/coredevice/test_moninj.py
stevefan1999: If there is no core device, we probably can't connect to the moninj and open a session. So it should be skipped, but seems like the next two lines after `self.core` should be separated from the exception block

artiq/test/coredevice/test_moninj.py
sbourdeauducq: Of course it is accurate. Nothing would work if it weren't.

artiq/test/coredevice/test_moninj.py
sbourdeauducq: The entire core device tests are skipped if ``ARTIQ_ROOT`` is not set. Assuming a core device is present when you are running the core device tests sounds reasonable. Stop copy-pasting code without understanding its context.

artiq/test/coredevice/test_moninj.py
```diff
@@ -66,3 +86,175 @@ def injection_status_cb(channel, override, value):
...
+            notifications.append((channel, probe, value))
+
+        moninj_comm = CommMonInj(monitor_cb, lambda x, y, z: None)
+        try:
+            await moninj_comm.connect(core_host)
+            if auto_monitor:
+                self.setup_monitor(moninj_comm)
+            yield moninj_comm
+        finally:
+            self.setup_monitor(moninj_comm, teardown=True)
```
sbourdeauducq, line 164: This teardown business is unnecessary. Had you been able to write the proxy correctly, you would have known that both the proxy and the core device already stop monitoring the relevant probes when a connection is closed.

artiq/test/coredevice/test_moninj.py
```diff
@@ -66,3 +86,175 @@ def injection_status_cb(channel, override, value):
...
+            await moninj_comm._writer.drain()
+            await asyncio.sleep(0.5)
+            await moninj_comm.close()
+
+    def setup_monitor(self, moninj_comm, *, teardown=False, reset_zero=True):
+        for name, dev in self.urukuls.items():
+            moninj_comm.monitor_probe(not teardown, dev.bus.channel, get_last_integers(name))
+            if reset_zero and not teardown:
+                self.kernel.write_raw(dev, 0)
+                ftw, _ = self.kernel.read_raw(dev)
```
sbourdeauducq, line 174: What is that for?

artiq/test/coredevice/test_moninj.py
```diff
@@ -66,3 +86,175 @@ def injection_status_cb(channel, override, value):
...
+        moninj_comm = CommMonInj(monitor_cb, lambda x, y, z: None)
+        try:
+            await moninj_comm.connect(core_host)
+            if auto_monitor:
+                self.setup_monitor(moninj_comm)
+            yield moninj_comm
+        finally:
+            self.setup_monitor(moninj_comm, teardown=True)
+            await moninj_comm._writer.drain()
+            await asyncio.sleep(0.5)
```
sbourdeauducq, line 166: What the hell?

stevefan1999: ikr but this is derived from `MonInjTest` above...

sbourdeauducq: Where?
If you are talking about this:
```python
moninj_comm.monitor_injection(True, loop_out_channel, TTLOverride.level.en.value)
loop.run_until_complete(asyncio.sleep(0.5))
moninj_comm.inject(loop_out_channel, TTLOverride.level.value, 0)
moninj_comm.inject(loop_out_channel, TTLOverride.level.oe.value, 1)
moninj_comm.inject(loop_out_channel, TTLOverride.level.en.value, 1)
loop.run_until_complete(asyncio.sleep(0.5))
...
```
you are completely misunderstanding what this code does and what the delays are for.

artiq/test/coredevice/test_moninj.py
stevefan1999: Resets the frequency tuning word to 0, sometimes if it has the same data the monitor probe notifier from firmware would simply not update. This also ensures the probe can be tested in a predictable way with an initial sequence.

artiq/test/coredevice/test_moninj.py
sbourdeauducq: Just the fact that there is no delay in the original code between ``drain()`` and ``close()`` should have made alarm bells go off, if you were paying attention.

artiq/test/coredevice/test_moninj.py
sbourdeauducq: Sure, the firmware only reports changes. This is correct and consistent behavior, not something that happens "sometimes".
I was asking about ``read_raw`` though.

artiq/gateware/rtio/phy/urukul_monitor.py
stevefan1999: In AD9910:
```
R/W—Bit 7 of the instruction byte determines whether a read
or write data transfer occurs after the instruction byte write.
Logic 1 indicates a read operation. Cleared indicates a write
operation.
```
In AD9912:
```
The MSB of the instruction word is R/W, which indicates
whether the instruction is a read or a write. The next two bits,
[W1:W0], are the transfer length in bytes. The final 13 bits are
the address ([A12:A0]) at which to begin the read or write
operation.
```
It seems like it would always be MSB, so just addressing the 31-th bit would be fine, but it would better be mentioned somewhere to not cause any confusion.

artiq/test/coredevice/test_moninj.py
stevefan1999: I see...let's get rid of that

artiq/test/coredevice/test_moninj.py
stevefan1999: actually I added a sleep there to yield to the asyncio loop for other tasks to continue specifically the receive task, it did not receive anything and failed the test. I changed it to a `asyncio.gather` call on the drain task, should there be better ways?

artiq/test/coredevice/test_moninj.py
```diff
@@ -66,3 +82,209 @@ def injection_status_cb(channel, override, value):
...
+            for name, urukul in self.urukuls_all().items():
+                idx = urukul.chip_select - 4
+                self.kernel.write(urukul, target_freqs[idx], 123.4)
+                freqs[idx] = self.kernel.read(urukul)[0]
+        for key, value in freqs.items():
+            assert numpy.isclose(value, target_freqs[key], rtol=1e-06)
+
+    @async_test
+    async def test_ad9910_set_all(self):
+        self.ensure_ad9910_only()
```
sbourdeauducq, line 274: only?

artiq/test/coredevice/test_moninj.py
sbourdeauducq: Please pay close attention to the code I copied above and understand it instead of this cargo cult nonsense. Calling ``asyncio.gather`` on a single object is pointless.

artiq/test/coredevice/test_moninj.py
stevefan1999: okay, removed

artiq/test/coredevice/test_moninj.py
stevefan1999: yeah, it is no better than just `await`ing it. it should be solved now.

artiq/test/coredevice/test_moninj.py
stevefan1999: changed its name to reflect what is it supposed to do

artiq/gateware/rtio/phy/urukul_monitor.py
```diff
@@ -55,23 +55,38 @@ def update_probe_data(i):
 
         # 0 -> init, 1 -> start reading
         state = [Signal(1) for i in range(nchannels)]
```
sbourdeauducq, line 57: FSM()

artiq/coredevice/ad9910.py
```diff
@@ -64,6 +65,9 @@
 # Default profile for RAM mode
 _DEFAULT_PROFILE_RAM = 0
 
+AD9910_REG_FTW = _AD9910_REG_FTW
+AD9910_REG_PROFILE0 = _AD9910_REG_PROFILE0
+AD9910_REG_PROFILE7 = _AD9910_REG_PROFILE7
```
sbourdeauducq, line 70: Remove the ``_`` in the first place (in a separate PR of course) or just import with ``_``.

artiq/gateware/eem.py
```diff
@@ -222,13 +222,13 @@ def io_qspi(eem0, eem1, iostandard):
         return ios
 
     @classmethod
-    def add_std(cls, target, eem, eem_aux, ttl_out_cls, sync_gen_cls=None, iostandard=default_iostandard):
+    def add_std(cls, target, eem, eem_aux, ttl_out_cls, sync_gen_cls=None, iostandard=default_iostandard, dds=None):
```
sbourdeauducq, line 225: Why is ``None`` an appropriate default value?
Also I think all the other methods have ``iostandard`` last.

artiq/gateware/rtio/phy/urukul_monitor.py
```diff
@@ -0,0 +1,141 @@
+from migen import *
+
+from artiq.coredevice.ad9910 import AD9910_REG_FTW, AD9910_REG_PROFILE0, AD9910_REG_PROFILE7
+from artiq.coredevice.ad9912_reg import AD9912_FTW0, AD9912_POW1
+from artiq.coredevice.spi2 import SPI_CONFIG_ADDR, SPI_DATA_ADDR, SPI_END
+from artiq.coredevice.urukul import CS_DDS_CH0, CS_DDS_MULTI, CS_CFG, CFG_IO_UPDATE
+
+
+class _AD9910_AD9912MonitorGeneric(Module):
```
sbourdeauducq, line 9: AD991x

artiq/gateware/rtio/phy/urukul_monitor.py
```diff
@@ -0,0 +1,141 @@
...
+        if self.io_update_phy is not None:
+            io_update_strobe_set_and_high = self.io_update_phy.rtlink.o.stb & self.io_update_phy.rtlink.o.data
+            return io_update_strobe_set_and_high | dest_cfg_reg_and_io_update_bit_set
+        return dest_cfg_reg_and_io_update_bit_set
+
+
+class AD9910Monitor(_AD9910_AD9912MonitorGeneric):
+    def __init__(self, spi_phy, io_update_phy, nchannels=4):
+        data = [Signal(32) for i in range(nchannels)]
+        buffer = [Signal(32) for i in range(nchannels)]
```
sbourdeauducq, line 45: Use more descriptive names, e.g. ``ftw_current`` / ``ftw_pending``

artiq/gateware/rtio/phy/urukul_monitor.py
```diff
@@ -0,0 +1,141 @@
...
+
+        # Flatten register first, then data
+        self.probes = Array(data)
+        super().__init__(spi_phy, io_update_phy)
+
+        def update_probe_data(i):
```
sbourdeauducq, line 51: I don't think you need a function definition for this.

artiq/test/coredevice/test_moninj.py
```diff
@@ -66,3 +71,237 @@ def injection_status_cb(channel, override, value):
...
+                    urukuls[cls][name] = self.device_mgr.get(name)
+        return urukuls
+
+    def urukuls_all(self):
+        ret = {}
+        for x in self.urukuls.values():
+            ret.update(x)
+        return ret
+
+    def ensure_ad9910_only(self):
```
sbourdeauducq, line 137: Poorly named function.

artiq/test/coredevice/test_moninj.py
```diff
@@ -66,3 +71,237 @@ def injection_status_cb(channel, override, value):
...
+
+        loop = asyncio.new_event_loop()
+        try:
+            with self.open_comm_session(loop, notifications_out):
+                for name, urukul in self.urukuls["AD9910"].items():
+                    idx = urukul.chip_select - 4
+                    self.kernel.write_with_amp(urukul, target_freqs[idx], 123.4, 0.25)
+                    freqs[idx] = self.kernel.read(urukul)[0]
+            for key, value in freqs.items():
+                assert numpy.isclose(value, target_freqs[key], rtol=1e-06)
```
sbourdeauducq, line 305: Use the ``unittest`` methods not ``assert``

artiq/test/coredevice/test_moninj.py
```diff
@@ -66,3 +71,237 @@ def injection_status_cb(channel, override, value):
...
+            if reset_zero:
+                self.kernel.write_raw(dev, 0)
+
+    def test_double_read(self):
+        target_ftw = 0xff00ff00
+        notifications_out = []
+
+        loop = asyncio.new_event_loop()
+        try:
+            with self.open_comm_session(loop, notifications_out):
```
sbourdeauducq, line 185: Doesn't unittest provide features to deal with that boilerplate? Look at hardware_testbench.py. You may also want to extend it and convert the other moninj tests to that (in a separate PR).

artiq/test/coredevice/test_moninj.py
```diff
@@ -66,3 +71,237 @@ def injection_status_cb(channel, override, value):
...
+        channel.set(freq, phase)
+
+    @kernel
+    def write_with_amp(self, channel, freq: TFloat, phase: TFloat = 0.0, amplitude: TFloat = 0.0):
+        self.init_channel(channel)
+        channel.set(freq, phase, amplitude)
+
+
+class AD991XMonitorTest(ExperimentCase, IsolatedAsyncioTestCase):
+    def get_urukuls(self):
```
sbourdeauducq, line 119: The coding pattern for every other HITL test is to look for a device with a particular name in the device db, not autodectect by driver class. Again, please keep things consistent and rewrite this to be exactly like the other tests.

artiq/test/coredevice/test_moninj.py
```diff
@@ -1,7 +1,12 @@
 import unittest
 import asyncio
+import numpy
+from collections import defaultdict
+from contextlib import contextmanager
+from unittest import IsolatedAsyncioTestCase
```
sbourdeauducq, line 6: Not used anywhere else in ARTIQ (it didn't exist when most of those tests were written), Needs to be introduced consistently in a separate PR or not used at all.

artiq/gateware/rtio/phy/urukul_monitor.py
stevefan1999: Not a good name since this mean AD9914 is part of it

sbourdeauducq: In a file named urukul_monitor it isn't.

stevefan1999: okay, let's split it into two files then

artiq/test/coredevice/test_moninj.py
stevefan1999: `IsolatedAsyncioTestCase` would allow the use of async in each test case function by doing the same thing. I just forgot to remove it, since we removed `async` support and used asyncio loop instead.

artiq/gateware/rtio/phy/urukul_monitor.py
sbourdeauducq: There's nothing wrong with urukul_monitor.py... _AD991xMonitor in urukul_monitor.py is fine, short and not ambiguous.

artiq/coredevice/ad9910.py
stevefan1999: done

artiq/gateware/rtio/phy/urukul_monitor.py
stevefan1999: done

artiq/test/coredevice/test_moninj.py
stevefan1999: this is now removed since i separated the test cases into two different class which implies only this kind of device would run this test.

artiq/test/coredevice/test_moninj.py
stevefan1999: done, let's be easy and put it in setUp/tearDown which is what every unit test runs before it start/stop

artiq/test/coredevice/test_moninj.py
stevefan1999: removed, done

artiq/gateware/rtio/phy/urukul_monitor.py
stevefan1999: done

artiq/test/coredevice/test_moninj.py
stevefan1999: done, but there's a slightly little issue on using assertAlmostEqual

artiq/gateware/rtio/phy/urukul_monitor.py
stevefan1999: done

artiq/gateware/eem.py
stevefan1999: done

artiq/gateware/rtio/phy/urukul_monitor.py
stevefan1999: Let's not change it now since I can't find ways to fix it, for some reason it doesn't work:
```py
        for i in range(nchannels):
            fsm = ClockDomainsRenamer("rio_phy")((FSM(reset_state="IDLE")))
            fsm.act("IDLE", [
                If(self.selected(i + CS_DDS_CH0) & (self.current_address == SPI_DATA_ADDR), [
                    If(~self.current_data[31], [
                        If((AD9910_REG_PROFILE0 <= reg) & (reg <= AD9910_REG_PROFILE7), [
                            # assume it is write64, which means the last command in the SPI sequence (SPI_END)
                            # will be FTW
                            NextState("READ_NEXT")
                        ]),
                        If(reg == AD9910_REG_FTW, [
                            If((self.length == 24) & (self.flags & SPI_END), [
                                # write16
                                buffer[i][:16].eq(self.current_data[8:24]),
                            ]),
                            If(self.length == 8, [
                                # write32
                                NextState("READ_NEXT")
                            ])
                        ])
                    ]),
                ])
            ])

            fsm.act("READ_NEXT", [
                If(self.selected(i + CS_DDS_CH0) & (self.current_address == SPI_DATA_ADDR), [
                    If(self.flags & SPI_END, [
                        buffer[i][:32].eq(self.current_data),
                        NextState("IDLE")
                    ])
                ])
            ])

            self.submodules += ClockDomainsRenamer("rio_phy")(fsm)

```

artiq/test/coredevice/test_moninj.py
stevefan1999: done

PR #1889, opened 2022-05-10
Forward async errors from satellites to the artiq master

mbirtwell:
# ARTIQ Pull Request

## Description of Changes

When async errors, like RTIO sequence errors, happen on the master kasli they are reported in the artiq master logs against the experiment RID, which is useful for seeing if any these types of errors potentially affected your experiment. However if these types of error occur on an satellite, then although the satellite reports the error to the master kasli the master kasli wouldn't forward that error to the artiq master. Instead it was just logged. 

This change adds forwarding in the above case. 

In order to make sure that async errors are likely to have been forwarded to the kasli when errors are reported to the artiq master the DRTIO destination survey is explicitly triggered and waited for. Otherwise it's quite common for the errors not to have been reported from the satellite yet. 

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

### All Pull Requests

### Code Changes

### Git Logistics

  ```
  
  ```

### Licensing

### DIFF
--- a/artiq/firmware/runtime/rtio_mgt.rs
+++ b/artiq/firmware/runtime/rtio_mgt.rs
@@ -1,4 +1,5 @@
 use core::cell::RefCell;
+use core::sync::atomic::{AtomicU8, Ordering};
 use urc::Urc;
 use board_misoc::csr;
 #[cfg(has_drtio)]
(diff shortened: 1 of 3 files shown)


sbourdeauducq:
In any case this would need to be ported to Zynq as well to avoid major divergence in behavior.

sbourdeauducq:
> However if these types of error occur on an satellite, then although the satellite reports the error to the master kasli the master kasli wouldn't forward that error to the artiq master. Instead it was just logged.

I don't understand this. Async RTIO errors are always reported through logging and are difficult to associate to a RID, since the error can occur after the experiment has terminated (and had left problematic events into the RTIO FIFOs). If you try hard enough, several experiments can even be involved in an async error.

For example, one experiment can schedule an event 10 seconds into the future and terminate. Then the next experiment schedules a conflicting event at exactly the same timestamp and terminates. About 10 seconds later, you get a collision error even though the two responsible experiments are no longer executing.

mbirtwell:
If I understand all this correctly, then it applies equally to collision errors that occur on the master kasli and yet we do send a summary of errors from the master kasli to the artiq master at the end of an experiment. Is there some reason this applies differently for the master and satellite that I've missed. 

Whilst what you say is clearly true in the general case, I think that most of the time we a running experiments that wait for all the events they've scheduled to occur and many of them clear the RTIO FIFO at the start. So the approximation that errors occurred during the kernel that triggered them is not unreasonable.

sbourdeauducq:
> yet we do send a summary of errors from the master kasli to the artiq master at the end of an experiment

It's not a summary of errors, it just says that some async error had been reported in the log since the previous kernel terminated. It's not precise or anything and is mostly intended to draw attention to the log, which is often missed by beginner users.

True - ``destination_survey()`` should also update ``SEEN_ASYNC_ERRORS`` but does not. Could you send a patch that fixes *only* that, without refactoring or other features?

sbourdeauducq:
> Could you send a patch that fixes _only_ that, without refactoring or other features?

@mbirtwell ping

mbirtwell:
Sorry, I'm not really sure what I want to do with this. We are, and will continue to use the version posted. I understand that there's some race conditions here, whatever we do, so people can't rely entirely on this message being reported at the most useful time. So there's a trade-off between the accuracy of the message and the complexity of the code. Also that there isn't a complete fix makes this trade-off unsatisfying. But I'm happy that this is worth it for us because the structure of our experiments is such that this will provide accurate information in the majority of cases. I realise this might not be true for everyone, but I think it is for us. And we have the code now so it seems a shame to go back. 

So, I'm left trying to decide whether to do an alternate version just for upstream, or not to bother. Which is a philosophical/ethical question as much as anything. And I've had other things to keep me busy. Also I'm going to be on holiday over most of the next few weeks. So I don't expect to resolve it soon, but I plan to think more about it when I'm back.

Issue #1142, opened 2018-08-30
Urukul monitoring/injection


jordens:
* monitoring is doable (tricky corners are the hidden and mutable state of the ref clock frequency, pll multiplier etc)
* injection is more difficult due to the shared resources (SPI bus and shared IO_UPDATE), exact requirements and options need to be discussed

sbourdeauducq:
One option for DDS "injection": the dashboard submits an experiment into the scheduler that simply sets the frequency.
That would require implementing support for submission-by-content to the master; this feature would play nice with a "sketchpad" feature that lets you write and test quick-and-dirty experiment code in the dashboard.

dhslichter:
Honestly I am not sure how necessary this is as an ARTIQ feature; we have already written simple "set DDS"-type experiments that achieve the desired injection, and I think that this is a cleaner way of doing it than having implicit experiment submissions coming from the DDS panel, especially given the potential for latency (e.g. if another experiment is already running or in the queue, if priorities are not properly set, etc.) and the need to hit a "submit new frequency" button anyway so that you're not firing off a new experiment each time a digit is changed.

sbourdeauducq:
It's just a nicer UI.
The experiment submissions are not so "implicit", they would appear in the scheduler panel, and pipeline/priority/etc. could be set in the options of the DDS panel. It would become pretty clear to the user what is happening.
There is no need for a "submit new frequency" button with this design: double click on the frequency to pop up a small text editor, then press Enter after editing the frequency to close the pop-up editor and submit the experiment.

klickverbot:
In terms of injection, what would be higher priority in my opinion than a nice UI would be working override functionality similar to what we have for TTLs, where one can pin the channel to a fixed frequency/phase setting regardless of what the experiment is doing. (This is motivated by diagnosing issues with both DDS hardware and the experiment attached to it over the last few months.)

There are a few tricks one could play in that regard, but I don't have a good design for this yet which doesn't involve shifting management of a considerable portion of state into the FPGA.

dhslichter:
@klickverbot yes this would be nice, but definitely requires more mucking around with the guts on the FPGA in a serious way.  @sbourdeauducq is there a reasonable way to maintain the DDS state in gateware or is this a nightmare?

sbourdeauducq:
[...]

PR #1902, opened 2022-06-06
[WIP] Dashboard: DDS Injection

Spaqin:
# ARTIQ Pull Request

## Description of Changes

This PR adds ability to set DDS channel frequency in the dashboard. Works with AD9914, AD9912, AD9910.

It works by generating a kernel with given frequency and sending it by content to the device.

The widget for each channel displays the frequency as before. If you move your cursor over one, OVR button will pop up:

Screenshot from 2022-06-06 12-11-13

And if you press it, you will be able to change the frequency, and either apply or cancel the changes. Using Enter keys to apply and Esc to cancel works as well.

Screenshot from 2022-06-06 12-11-40

Tested with kc705 (looking into logs, kernel is run, not verified AD9914 functionality), and a Kasli with an AD9912. The core functionality works - frequency can be easily changed - but has some limitations:
- new frequency is not displayed in the menu (moninj does not report it back, I suppose?). I suppose displaying the user value doesn't make sense as it doesn't have to exactly match the reality (device can be restarted, and also see below).
- since they're all newly generated kernels, if you set one channel of an Urukul card, the rest are re-initialized and become 0 again. This would require changes in the Urukul driver to retrieve config and keep previous state somehow, instead of starting from scratch every time. Thus at this point only one channel per card can be really used.

Thus before merging I'd like to address these issues and discuss if and what should be done within this one (or maybe another one, to keep issues separate?) - that's why even though the feature itself feels complete, I'm marking it as WIP.

### Related Issue

Related #1142

## Type of Changes

|   | Type |

### All Pull Requests

### Code Changes

### Git Logistics

  ```
  
  ```

### Licensing

### DIFF
--- a/artiq/dashboard/experiments.py
+++ b/artiq/dashboard/experiments.py
@@ -703,6 +703,26 @@ class ExperimentManager:
             scheduling["priority"], scheduling["due_date"],
             scheduling["flush"]))
 
+    def submit_by_content(self, content, class_name):
+        expid = {
+            "log_level": logging.WARNING,
+            "content": content,
+            "class_name": class_name,
+            "arguments": []
+        }
+        scheduling = {
+            "pipeline_name": "main",
+            "priority": 0,
+            "due_date": None,
+            "flush": False
+        }
+        asyncio.ensure_future(self._submit_task(
+            "Content",
+            scheduling["pipeline_name"],
+            expid,
+            scheduling["priority"], scheduling["due_date"],
+            scheduling["flush"]))
+
     async def _request_term_multiple(self, rids):
         for rid in rids:
             try:
--- a/artiq/dashboard/experiments.py
+++ b/artiq/dashboard/experiments.py
@@ -703,26 +703,6 @@ class ExperimentManager:
             scheduling["priority"], scheduling["due_date"],
             scheduling["flush"]))
 
-    def submit_by_content(self, content, class_name, description=None):
-        expid = {
-            "log_level": logging.WARNING,
-            "content": content,
-            "class_name": class_name,
-            "arguments": []
-        }
-        scheduling = {
-            "pipeline_name": "main",
-            "priority": 0,
-            "due_date": None,
-            "flush": False
-        }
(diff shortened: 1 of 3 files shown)


Spaqin:
Changed the interface a bit:
Screenshot from 2022-06-07 13-27-19
Screenshot from 2022-06-07 13-27-27

(should we add an "off" button next to set? although it would be Urukul only)

Also using ``core_cache`` to write urukul CPLD config now, so multiple channels can be set and they won't switch others off.

sbourdeauducq:
> should we add an "off" button next to set? although it would be Urukul only

Sounds like a good idea.

sbourdeauducq:
Make sure it works in 1-EEM and 2-EEM modes.

artiq/dashboard/moninj.py
```diff
@@ -168,15 +169,146 @@ def sort_key(self):
...
+class _DDSWidget(QtWidgets.QFrame):
+    class CancellableLineEdit(QtWidgets.QLineEdit):
+        def escapePressedConnect(self, cb):
+            self.esc_cb = cb
+
+        def keyPressEvent(self, event):
+            key = event.key()
+            if key == QtCore.Qt.Key_Escape:
+                self.esc_cb(event)
+            QtWidgets.QLineEdit.keyPressEvent(self, event)
```
sbourdeauducq, line 181: This doesn't need to be defined inside ``_DDSWidget`` - move out?

artiq/dashboard/experiments.py
```diff
@@ -703,6 +703,26 @@ def submit(self, expurl):
             scheduling["priority"], scheduling["due_date"],
             scheduling["flush"]))
 
+    def submit_by_content(self, content, class_name, description=None):
```
sbourdeauducq, line 706: The experiment manager is dealing with the experiment GUI docks, the repos, etc.
It is not the place to define unrelated internal APIs.
Pass the scheduler connection to moninj instead and move that function there. ``_submit_task`` is a 2-line function that is not interesting to share.

PR #1912, opened 2022-06-17
master/worker_db: Support lists of strings when writing to HDF5

dnadlinger:
This does the obvious thing of writing list of strings as datasets
of variable-length UTF string datasets. h5py does this automatically
for dataset attributes, but for whatever reason not for datasets
themselves.

### DIFF
--- a/artiq/master/worker_db.py
+++ b/artiq/master/worker_db.py
@@ -5,8 +5,10 @@ standalone command line tools).
 """
 
 from operator import setitem
+import h5py
 import importlib
 import logging
+import numpy
 
 from sipyco.sync_struct import Notifier
 from sipyco.pc_rpc import AutoTarget, Client, BestEffortClient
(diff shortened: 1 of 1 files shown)


sbourdeauducq:
Could this be a bug in h5py?

dnadlinger:
This appears to be the case, as I could no longer manually reproduce the issue on 3.6.0. I regret not taking the time to add a regression test for this back in 2019…

dnadlinger:
Ah, this was likely fixed in h5py 3.0. From the [release notes](https://docs.h5py.org/en/stable/whatsnew/3.0.html):

> In previous versions, creating a dataset from a list of bytes objects would choose a fixed length string datatype to fit the biggest item. It will now use a variable length string datatype. To store fixed length strings, use a suitable dtype from [h5py.string_dtype()](https://docs.h5py.org/en/stable/special.html#h5py.string_dtype).

PR #1850, opened 2022-02-16
(WIP) feat: urukul monitoring support

stevefan1999:
# ARTIQ Pull Request

## Description of Changes

This is a rebase of https://github.com/stevefan1999/artiq/tree/patch-urukul, which is based on the refactored moninj branch, which is #1753. The entire file is a big mess to maintain sure, but I realized it is less of a concern since we could always do the code split later. Still, I can just port the code from the modular base to monolithic base quite easily, as I can just reattach the code almost out of the box.

This patch adds Urukul support. For monitoring it just works, for injection however would involve adding a new type of hidden experiment that would be submitted to master (it is yet to be finalized however)

### Related Issue

Closes #1142

## Type of Changes

|   | Type |
| ------------- | ------------- |
| ✓  | :sparkles: New feature |

### All Pull Requests

- [x] Use correct spelling and grammar.
- [ ] Update [RELEASE_NOTES.rst](../RELEASE_NOTES.rst) if there are noteworthy changes, especially if there are changes to existing APIs.
- [ ] Close/update issues.
- [ ] Check the copyright situation of your changes and sign off your patches (`git commit --signoff`, see [copyright](../CONTRIBUTING.rst#copyright-and-sign-off)).

### Code Changes

- [ ] Run `flake8` to check code style (follow PEP-8 style). `flake8` has issues with parsing Migen/gateware code, ignore as necessary.
- [ ] Test your changes or have someone test them. Mention what was tested and how.
- [ ] Add and check docstrings and comments
- [ ] Check, test, and update the [unittests in /artiq/test/](../artiq/test/) or [gateware simulations in /artiq/gateware/test](../artiq/gateware/test)

### Documentation Changes

- [ ] Check, test, and update the documentation in [doc/](../doc/). Build documentation (`cd doc/manual/; make html`) to ensure no errors.

### Git Logistics

- [ ] Split your contribution into logically separate changes (`git rebase --interactive`). Merge/squash/fixup commits that just fix or amend previous commits. Remove unintended changes & cleanup. See [tutorial](https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase).
- [ ] Write short & meaningful commit messages. Review each commit for messages (`git show`). Format:
  ```
  topic: description. < 50 characters total.
  
  Longer description. < 70 characters per line
  ```

### Licensing

See [copyright & licensing for more info](https://github.com/m-labs/artiq/blob/master/CONTRIBUTING.rst#copyright-and-sign-off).
[...]

PR #1921, opened 2022-06-21
Urukul monitoring

### LINKED ISSUE #1142: Urukul monitoring/injection

Spaqin:
# ARTIQ Pull Request

## Description of Changes

Added support for Urukul monitoring in the dashboard.

Code mostly lifted from #1850 - with few changes, mostly on the dashboard side not to put everything together.

Tested with AD9910 and AD9912. 

Screenshot from 2022-06-21 17-33-32

Since it monitors what is sent to the DDS themselves but has no control over the CPLD or rf_sw, it won't know if the channel is turned off or on, however. But the frequencies will be displayed.

### Related Issue

Closes #1142

## Type of Changes

|   | Type |

### All Pull Requests

### Code Changes

### Git Logistics

  ```
  
  ```

### Licensing


artiq/dashboard/moninj.py
```diff
@@ -545,6 +620,7 @@ def build(self):
                 
             @kernel
             def run(self):
+                self.core.reset()
                 self.core.break_realtime()
```
sbourdeauducq, line 624: You should not need break_realtime after reset.

artiq/coredevice/ad9910.py
```diff
@@ -65,6 +66,22 @@
...
+
+
+def AD9910_REG_PROFILE0():
+    return _AD9910_REG_PROFILE0
+
+
+def AD9910_REG_PROFILE7():
+    return _AD9910_REG_PROFILE7
+
+
```
sbourdeauducq, line 84: A bit weird and long. Just import the underscore names instead - Python allows it, and using them for the monitoring core is arguably internal use.

artiq/dashboard/moninj.py
```diff
@@ -179,14 +181,81 @@ def sort_key(self):
...
+        if self.dds_type == "AD9910":
+            self.cur_reg = (reg >> 24) & 0xff
+        else: # AD9912
+            self.cur_reg = ((reg >> 16) & ~(3 << 13)) & 0xffff
+
+    def _update_data_low(self, value):
+        if self.dds_type == "AD9910":
+            if (AD9910_REG_PROFILE0() <= self.cur_reg <= AD9910_REG_PROFILE7() or
+                    self.cur_reg == AD9910_REG_FTW()):
+                self.cur_frequency_low = self._ftw_to_freq(value)
```
sbourdeauducq, line 224: Those non-atomic updates aren't nice.
If atomic is difficult then ignore the lower FTW bits on AD9912 (with corresponding loss of resolution) - it doesn't sound worse than those non-atomic updates, and it is simpler.

Spaqin: AD9912 values without the low part don't look great:

Screenshot from 2022-06-22 12-31-22

Atomic updates kind of don't exactly make sense either - sometimes only the high or only the low part would be updated (e.g. from 10.075 to 10.070MHz - only low part will get updated; surely there's a combination in which only the high part updated as well). So, if low part or high part comes by itself - these are both valid options. If there was a way for moninj to indicate end of probe updates for given channel, then that could be used as an indicator for the update - but adding one would be too much for this tiny issue.

artiq/gateware/rtio/phy/urukul_monitor.py
```diff
@@ -0,0 +1,50 @@
...
+        self.probes = Array([Signal(32) for i in range(3 * nchannels)])
+        self.rtlink = spi_rtlink
+
+        current_address = Signal.like(self.rtlink.o.address)
+        current_data = Signal.like(self.rtlink.o.data)
+        self.sync.rio += If(
+            self.rtlink.o.stb,
+            current_address.eq(self.rtlink.o.address),
+            current_data.eq(self.rtlink.o.data)
+        )
```
sbourdeauducq, line 19: What is this for?

artiq/gateware/rtio/phy/urukul_monitor.py
```diff
@@ -0,0 +1,50 @@
...
+
+        cs = Signal(8)
+        length = Signal(8)
+        end = Signal()
+        self.sync.rio += If(
+            self.rtlink.o.stb & (current_address == SPI_CONFIG_ADDR),
```
sbourdeauducq, line 25: rtlink address is only valid in the cycle at which ``stb`` is asserted. Here they are shifted by one clock cycle (with the pipeline delay on ``current_address``) and this looks incorrect.
Stefan's code is full of problems and generally not an example to follow.

artiq/dashboard/moninj.py
sbourdeauducq: > AD9912 values without the low part don't look great:

That's surprising - if you ignore the lower 16 FTW bits (or round the result instead of just truncating), you should still get 32 bits of precision and it should be as precise as the AD9910.
Are you sure you are truncating and interpreting the truncated FTW correctly?

sbourdeauducq: FYI, the AD9914 driver has a mode with 63-bit FTW, of which the lower 31 bits are ignored by moninj, and it does not have issues like that.

artiq/gateware/rtio/phy/urukul_monitor.py
```diff
@@ -0,0 +1,50 @@
+from migen import *
+
+from artiq.coredevice.spi2 import SPI_CONFIG_ADDR, SPI_DATA_ADDR, SPI_END
+from artiq.coredevice.urukul import CS_DDS_CH0, CS_DDS_MULTI
+
+
+class UrukulMonitor(Module):
+    def __init__(self, spi_rtlink, nchannels=4):
+        # [0:3] = register address, [4:7] = data_high, [8:11] = data_low (or just data for 32-bit transfers)
+        self.probes = Array([Signal(32) for i in range(3 * nchannels)])
```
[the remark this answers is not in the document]
sbourdeauducq: The AD9914 monitoring core decodes FTW transactions in gateware and should be the reference design to follow here.

artiq/dashboard/moninj.py
sbourdeauducq: > Atomic updates kind of don't exactly make sense either - sometimes only the high or only the low part would be updated (e.g. from 10.075 to 10.070MHz - only low part will get updated; surely there's a combination in which only the high part updated as well)

Transmitting only the changed bytes is a small and unimportant optimization. But if you don't do atomic updates, and the CPU looks at the probes at the wrong time while the FTW bits are completely changed by the RTIO/SPI system, then the frequency display would glitch (as only half the FTW took the new value) and show a completely incorrect value until the next moninj scan.

Truncating the FTW like the AD9914 does (when operated in extended resolution mode) is an easy way out of this problem.

artiq/gateware/rtio/phy/dds.py
```diff
@@ -140,12 +141,12 @@ def __init__(self, current_data, data_length, flags, ch_sel):
...
                 If(flags & SPI_END,
                     # lower 16 bits (16-32 from 48-bit transfer)
-                    NextValue(self.ftw[0:16], current_data[16:32]),
+                    NextValue(self.ftw[:16], current_data[16:]),
                     NextState("START")
-                ).Else(
-                    # upper 16 bits (32-48 from 48-bit transfer)
-                    NextValue(self.ftw[16:32], current_data[0:16]),
-            ))
+                ).Elif(current_data != 0,
```
Spaqin, line 146: Since ``current_data`` is connected to ``rtlink.o.data`` through comb instead of getting latched, the ftw value would get overwritten with zeroes. Checking the value like that is a bit of an ugly workaround and while it works, I'd rather do it properly, with a register.

artiq/dashboard/moninj.py
```diff
@@ -653,7 +691,6 @@ async def mi_connector(self):
             try:
                 await new_mi_connection.connect(self.mi_addr, self.mi_port)
             except asyncio.CancelledError:
-                logger.info("cancelled connection to moninj")
```
sbourdeauducq, line 656: Why is this PR removing this?

artiq/dashboard/moninj.py
```diff
@@ -545,8 +583,8 @@ def build(self):
                 
             @kernel
             def run(self):
-                self.core.break_realtime()
-                delay(2*ms)
+                self.core.reset()
+                delay(3*ms)
```
sbourdeauducq, line 587: Please move this to other PR

sbourdeauducq:
Looking good except for minor details noted

PR #1927, opened 2022-06-29
language.environment: Add check_unused_arguments

### LINKED ISSUE #1641: scheduler: check for mistyped / non-present arguments

Deepskyhunter:
# ARTIQ Pull Request

## Description of Changes
**Problem**: scheduler device might have added incorrect arguments as stated in issue #1641

**Solution**: Check the unused arguments in expid in unprocessed_arguments after build

**Details**: remove argument in unprocessed_arguments after get() during build. Call check_unused_arguments() after build to see what argument is left. Raise exception and state those arguments.

**Test result**:
Screenshot from 2022-06-29 15-24-48
Based on the example in #1641 
After submitting RID 40, exception occur and stating what arguments were not being used.
After correcting the typo, submit it as RID 42 and successfully changed the text output.

## Type of Changes

|   | Type |
|   | :bug: Bug fix  |
|   | :hammer: Refactoring  |
|   | :scroll: Docs |

### All Pull Requests

- [x] Close/update issues.

### Code Changes

### Documentation Changes

### Git Logistics

  ```
  
  ```

### Licensing

### DIFF
--- a/artiq/language/environment.py
+++ b/artiq/language/environment.py
@@ -213,18 +213,21 @@ class TraceArgumentManager:
 class ProcessArgumentManager:
     def __init__(self, unprocessed_arguments):
         self.unprocessed_arguments = unprocessed_arguments
+        self._processed_arguments = []
 
     def get(self, key, processor, group, tooltip):
         if key in self.unprocessed_arguments:
             r = processor.process(self.unprocessed_arguments[key])
-            self.unprocessed_arguments.pop(key)
+            self._processed_arguments.append(key)
         else:
             r = processor.default()
         return r
 
     def check_unprocessed_arguments(self):
-        if self.unprocessed_arguments:
-            args = ", ".join(self.unprocessed_arguments.keys())
+        unprocessed = set(self.unprocessed_arguments.keys()) -\
+                      set(self._processed_arguments)
+        args = ", ".join(list(unprocessed))
+        if unprocessed:
             raise AttributeError("Unprocessed argument(s): " + args)
 
 class HasEnvironment:
--- a/artiq/language/environment.py
+++ b/artiq/language/environment.py
@@ -215,19 +215,19 @@ class TraceArgumentManager:
 class ProcessArgumentManager:
     def __init__(self, unprocessed_arguments):
         self.unprocessed_arguments = unprocessed_arguments
-        self._processed_arguments = []
+        self._processed_arguments = set()
 
     def get(self, key, processor, group, tooltip):
         if key in self.unprocessed_arguments:
             r = processor.process(self.unprocessed_arguments[key])
-            self._processed_arguments.append(key)
+            self._processed_arguments.add(key)
         else:
             r = processor.default()
         return r
 
     def check_unprocessed_arguments(self):
         unprocessed = set(self.unprocessed_arguments.keys()) -\
-                      set(self._processed_arguments)
+                      self._processed_arguments
         if unprocessed:
             raise AttributeError("Unprocessed argument(s): " +
                                  ", ".join(unprocessed))
--- a/artiq/language/environment.py
+++ b/artiq/language/environment.py
@@ -222,15 +222,10 @@ class ProcessArgumentManager:
             r = processor.default()
         return r
 
-    def check_unused_arguments(self):
-        if len(self.unprocessed_arguments) > 0:
-            args = ""
-            for key in self.unprocessed_arguments.keys():
-                if args == "":
-                    args += str(key)
-                else:
-                    args += (", "+str(key))
-            raise Exception("Unused argument(s): " + args)
+    def check_unprocessed_arguments(self):
+        if self.unprocessed_arguments:
+            args = ", ".join(self.unprocessed_arguments.keys())
+            raise AttributeError("Unprocessed argument(s): " + args)
 
 class HasEnvironment:
     """Provides methods to manage the environment of an experiment (arguments,
@@ -253,7 +248,7 @@ class HasEnvironment:
         self.build(*args, **kwargs)
         self.__in_build = False
         if isinstance(self.__argument_mgr, ProcessArgumentManager):
-            self.__argument_mgr.check_unused_arguments()
+            self.__argument_mgr.check_unprocessed_arguments()
 
     def register_child(self, child):
         self.children.append(child)
--- a/artiq/language/environment.py
+++ b/artiq/language/environment.py
@@ -229,7 +229,7 @@ class ProcessArgumentManager:
                 if args == "":
                     args += str(key)
                 else:
-                    args += (" ,"+str(key))
+                    args += (", "+str(key))
             raise Exception("Unused argument(s): " + args)
 
 class HasEnvironment:
--- a/artiq/language/environment.py
+++ b/artiq/language/environment.py
@@ -229,7 +229,7 @@ class ProcessArgumentManager:
         unprocessed = set(self.unprocessed_arguments.keys()) -\
                       self._processed_arguments
         if unprocessed:
-            raise AttributeError("Unprocessed argument(s): " +
+            raise AttributeError("Invalid argument(s): " +
                                  ", ".join(unprocessed))
 
 class HasEnvironment:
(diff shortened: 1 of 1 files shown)


artiq/language/environment.py
```diff
@@ -217,10 +217,20 @@ def __init__(self, unprocessed_arguments):
     def get(self, key, processor, group, tooltip):
         if key in self.unprocessed_arguments:
             r = processor.process(self.unprocessed_arguments[key])
+            self.unprocessed_arguments.pop(key)
         else:
             r = processor.default()
         return r
 
+    def check_unused_arguments(self):
+        if len(self.unprocessed_arguments) > 0:
```
sbourdeauducq, line 226: ``if self.unprocessed_arguments:``

artiq/language/environment.py
```diff
@@ -217,10 +217,20 @@ def __init__(self, unprocessed_arguments):
...
+            args = ""
+            for key in self.unprocessed_arguments.keys():
+                if args == "":
+                    args += str(key)
+                else:
+                    args += (", "+str(key))
+            raise Exception("Unused argument(s): " + args)
```
sbourdeauducq, line 233: Use a more appropriate exception type.

artiq/language/environment.py
```diff
@@ -217,10 +217,20 @@ def __init__(self, unprocessed_arguments):
...
         return r
 
+    def check_unused_arguments(self):
+        if len(self.unprocessed_arguments) > 0:
+            args = ""
+            for key in self.unprocessed_arguments.keys():
+                if args == "":
+                    args += str(key)
+                else:
+                    args += (", "+str(key))
```
sbourdeauducq, line 232: ``str.join()``

artiq/language/environment.py
```diff
@@ -242,6 +252,8 @@ def __init__(self, managers_or_parent, *args, **kwargs):
         self.__in_build = True
         self.build(*args, **kwargs)
         self.__in_build = False
+        if isinstance(self.__argument_mgr, ProcessArgumentManager):
+            self.__argument_mgr.check_unused_arguments()
```
sbourdeauducq, line 256: Is this the best place to check for this? Why?

artiq/language/environment.py
```diff
@@ -217,10 +217,20 @@ def __init__(self, unprocessed_arguments):
     def get(self, key, processor, group, tooltip):
         if key in self.unprocessed_arguments:
             r = processor.process(self.unprocessed_arguments[key])
+            self.unprocessed_arguments.pop(key)
```
sbourdeauducq, line 220: Is there no side effect to modifying ``self.unprocessed_arguments`` here?

Deepskyhunter: `ProcessArgumentManager` is created in every run, so there will be new unprocessed_arguments every time.

Deepskyhunter: I think no because each argument only called this function once during build stage.

artiq/language/environment.py
Deepskyhunter: I think the best place to check is right after setting all arguments, which should be after build stage.

artiq/language/environment.py
sbourdeauducq: Can you avoid this kind of mutation anyway? It's quite a footgun that ``ProcessArgumentManager`` modifies the dictionary it was given at creation when you call ``get()`` later. Especially since the creation doesn't even occur in this file.

For example, you could add a ``_processed_arguments`` private set and add to it instead.

artiq/language/environment.py
```diff
@@ -213,14 +213,22 @@ def get(self, key, processor, group, tooltip):
...
             r = processor.process(self.unprocessed_arguments[key])
+            self._processed_arguments.append(key)
         else:
             r = processor.default()
         return r
 
+    def check_unprocessed_arguments(self):
+        unprocessed = set(self.unprocessed_arguments.keys()) -\
+                      set(self._processed_arguments)
+        args = ", ".join(list(unprocessed))
```
sbourdeauducq, line 229: Move this into the ``if`` obviously

artiq/language/environment.py
```diff
@@ -242,6 +250,8 @@ def __init__(self, managers_or_parent, *args, **kwargs):
         self.__in_build = True
         self.build(*args, **kwargs)
         self.__in_build = False
+        if isinstance(self.__argument_mgr, ProcessArgumentManager):
```
sbourdeauducq, line 253: Why the isinstance check here?
Wouldn't an empty method in the other object types be better?

Deepskyhunter: Oh right, I didn't think of this solution.

artiq/language/environment.py
```diff
@@ -209,18 +209,28 @@ def get(self, key, processor, group, tooltip):
...
+            self._processed_arguments.append(key)
         else:
             r = processor.default()
         return r
 
+    def check_unprocessed_arguments(self):
+        unprocessed = set(self.unprocessed_arguments.keys()) -\
+                      set(self._processed_arguments)
+        if unprocessed:
+            args = ", ".join(list(unprocessed))
```
sbourdeauducq, line 232: Small details:
1. ``list()`` is unnecessary
2. ``args`` is a poor variable name...
3. ...which is not necessary anyway, you can just combine this line with the next.

artiq/language/environment.py
```diff
@@ -209,18 +209,28 @@ def get(self, key, processor, group, tooltip):
...
     def get(self, key, processor, group, tooltip):
         if key in self.unprocessed_arguments:
             r = processor.process(self.unprocessed_arguments[key])
+            self._processed_arguments.append(key)
         else:
             r = processor.default()
         return r
 
+    def check_unprocessed_arguments(self):
+        unprocessed = set(self.unprocessed_arguments.keys()) -\
```
sbourdeauducq, line 229: ``\`` is not necessary, is it?

Deepskyhunter: It is needed

sbourdeauducq: Really?
```
sb@vulcan ~> cat foo.py                                                         
x = set({1, 2} -
        {1, 4})
print(x)
sb@vulcan ~> python foo.py                                                      
{2}
```

artiq/language/environment.py
```diff
@@ -209,18 +209,28 @@ def get(self, key, processor, group, tooltip):
...
+                      set(self._processed_arguments)
```
sbourdeauducq, line 230: Why not just make ``_processed_arguments`` a set in the first place?

artiq/language/environment.py
```diff
@@ -209,18 +209,28 @@ def get(self, key, processor, group, tooltip):
...
+        if unprocessed:
+            raise AttributeError("Unprocessed argument(s): " +
```
sbourdeauducq, line 232: This is not a very user-friendly message, since the "argument processor" is not documented or exposed. Something like "Invalid arguments: ..." would be better

PR #1950, opened 2022-08-10
GUI log: Apply level and text filter to the log messages in panel

### LINKED ISSUE #851: searching log messages in dashboard panels
Currently, the filters only applies to the log messages as they arrive. There is currently no way in the dashboard to search for a previously received log message by keyword (and, with lesser importance, by priority). I see two ways of doing this:
1. also applying the filters to the log messages already in the panel's buffer.
2. adding another search feature.

I'm more leaning towards option 1.

Deepskyhunter:
# ARTIQ Pull Request

## Description of Changes
Apply level and text filter to the log messages in panel
Incoming message filtering is removed

## Type of Changes

|   | Type |

### All Pull Requests

- [x] Close/update issues.

### Code Changes

### Documentation Changes

### Git Logistics

  ```
  
  ```

### Licensing

### DIFF
--- a/artiq/gui/log.py
+++ b/artiq/gui/log.py
@@ -276,7 +276,7 @@ class LogDock(QDockWidgetCloseDetect):
         self.model.rowsInserted.connect(self.rows_inserted_after)
         self.model.rowsRemoved.connect(self.rows_removed)
 
-        self.filter_freetext.textChanged.connect(self.apply_text_filter)
+        self.filter_freetext.returnPressed.connect(self.apply_text_filter)
         self.filter_level.currentIndexChanged.connect(self.apply_level_filter)
 
     def apply_text_filter(self):
--- a/artiq/gui/log.py
+++ b/artiq/gui/log.py
@@ -18,17 +18,18 @@ class _ModelItem:
         self.children_by_row = []
 
 
-class _ProxyModel(QtCore.QSortFilterProxyModel):
+class _FilterProxyModel(QtCore.QSortFilterProxyModel):
     def __init__(self):
         super().__init__()
         self.setFilterCaseSensitivity(QtCore.Qt.CaseInsensitive)
+        self.setRecursiveFilteringEnabled(True)
         self.filter_level = 0
 
     def filterAcceptsRow(self, source_row, source_parent):
         source = self.sourceModel()
         index_0 = source.index(source_row, 0, source_parent)
 
-        entry = source.data(index_0, None)
+        entry = source.data(index_0, UserRole)
         if entry[0] >= self.filter_level:
             regex = self.filterRegExp()
             return (regex.indexIn(entry[1]) != -1 
@@ -191,8 +192,8 @@ class _Model(QtCore.QAbstractItemModel):
             return (log_level_to_name(v[0]) + ", " +
                 time.strftime("%m/%d %H:%M:%S", time.localtime(v[2])) +
                 "\n" + v[3][lineno])
-        elif role == None:
-            return self.entries[msgnum]
+        elif role == QtCore.Qt.UserRole:
+            return self.entries[msgnum][0]
 
 
 class LogDock(QDockWidgetCloseDetect):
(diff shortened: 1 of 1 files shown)


artiq/gui/log.py
```diff
@@ -168,6 +191,8 @@ def data(self, index, role):
             return (log_level_to_name(v[0]) + ", " +
                 time.strftime("%m/%d %H:%M:%S", time.localtime(v[2])) +
                 "\n" + v[3][lineno])
+        elif role == None:
```
sbourdeauducq, line 194: Why not pass a valid role in ``filterAcceptsRow()`` instead?

artiq/gui/log.py
```diff
@@ -18,6 +18,29 @@ def __init__(self, parent, row):
         self.children_by_row = []
 
 
+class _ProxyModel(QtCore.QSortFilterProxyModel):
```
sbourdeauducq, line 21: Class name needs to be more specific...

artiq/gui/log.py
```diff
@@ -240,11 +265,24 @@ def __init__(self, manager, name):
...
         self.model.rowsAboutToBeInserted.connect(self.rows_inserted_before)
         self.model.rowsInserted.connect(self.rows_inserted_after)
         self.model.rowsRemoved.connect(self.rows_removed)
 
+        self.filter_freetext.textChanged.connect(self.apply_text_filter)
+        self.filter_level.currentIndexChanged.connect(self.apply_level_filter)
+
+    def apply_text_filter(self):
+        text = self.filter_freetext.text()
+        self.proxy_model.setFilterRegExp(text)
```
sbourdeauducq, line 281: In the function just below you wrote a single line.

sbourdeauducq:
What are the performance consequences of this? Qt expects MVC models to be written in C++ and be fast, and calls them redundantly without any kind of caching - PyQt's Python models can very easily result in very sluggish UIs.
Please try this code on the old computer in the annex.

artiq/gui/log.py
Deepskyhunter: The `DisplayRole` is used in line 172 to get the message of each row and it will not be shown if its parents does not fit the text filter. Therefore it has to return the original entry for checking.

sbourdeauducq: I don't understand. Where does ``_Model`` know about that filter?

Deepskyhunter: `_ProxyModel` is acting as the filter and `_Model` just responsible for passing the data

sbourdeauducq: Exactly. So just pass a valid role to ``_Model`` in ``filterAcceptsRow`` instead of abusing Python typing. What is the problem?

Deepskyhunter:
image

It has around 200 log message

sbourdeauducq:
Doesn't seem too bad. But we probably want to remove the incoming message filter if we have this.

artiq/gui/log.py
```diff
@@ -18,6 +18,32 @@ def __init__(self, parent, row):
...
+    def __init__(self):
+        super().__init__()
+        self.setFilterCaseSensitivity(QtCore.Qt.CaseInsensitive)
+        self.setRecursiveFilteringEnabled(True)
+        self.filter_level = 0
+
+    def filterAcceptsRow(self, source_row, source_parent):
+        source = self.sourceModel()
+        src = source.index(source_row, 0, source_parent)
+        msg = source.index(source_row, 1, source_parent)
```
sbourdeauducq, line 31: Call those index0 and index1 (like in the Qt example). Why are you passing ``src`` on the line below otherwise?

Deepskyhunter: > Why are you passing `src` on the line below otherwise?

To get the severity of that row. Either `src` or `msg` is needed.

sbourdeauducq: Yes I understand this. My point is if it's called ``src`` instead of ``index0``, it misleadingly suggests the logging level has to do with the source in particular.

PR #2026, opened 2022-12-09
Store non-archive datasets in memory

### LINKED ISSUE #2023: `set_dataset` with `archive=False` throws `KeyError` when accessing
Running 
```python
class Test(EnvExperiment):
    def prepare(self):
        self.set_dataset("x", 1, archive=False)

    def run(self):
        print(self.get_dataset("x"))
```
throws a `KeyError`
```python
root:Terminating with exception (KeyError: "'x'")
  File "C:\tools\miniconda3\envs\artiq-7\lib\site-packages\artiq\master\worker.py", line 229, in _handle_worker_requests
    data = func(*obj["args"], **obj["kwargs"])
  File "C:\tools\miniconda3\envs\artiq-7\lib\site-packages\artiq\master\databases.py", line 62, in get
    return self.data.raw_view[key][1]
KeyError: "'x'"
```
If I run the same code with `archive=True`, which is the default everything works as expected.

I am running ARTIQ-7 version 7.8123.3038639.

Edit: Tested this also on 6.7665.83af3b75 with the same result.

thomasfire:
# ARTIQ Pull Request

## Description of Changes

Store non-archive datasets in memory. Without this change `set_dataset(broadcast=False, persist=False, archive=False)` would not set any new values.

### Related Issue

Closes #2023 

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

## Licensing


artiq/master/worker_db.py
```diff
@@ -119,6 +119,10 @@ def set(self, key, value, broadcast=False, persist=False, archive=True):
         if persist:
             broadcast = True
 
+        if not (broadcast or archive):
+            logger.warning("No dataset will be stored. Both 'broadcast' and 'archive' are set to False.",
```
jordens, line 123: Fix the wording. "No dataset will be stored" means that there will not be any dataset stored.
Also no need to print the trace and the logging message is missing the info which dataset is affected.

PR #2060, opened 2023-03-06
coredevice: support Almazny v1.2

jordens:
This does not affect the legacy Almazny v1.1/v1.0 support.

### DIFF
--- a/artiq/coredevice/almazny.py
+++ b/artiq/coredevice/almazny.py
@@ -1,0 +1,68 @@
+from artiq.language.core import kernel, portable
+
+from numpy import int32
+
+
+class AlmaznyChannel:
+    """
+    One Almazny channel
+
+    Almazny is a mezzanine for the Quad PLL RF source Mirny that exposes and
+    controls the frequency-doubled outputs.
+
+    This driver requires Almazny hardware revision v1.2 or later
+    and Mirny CPLD gateware v0.3 or later.
+
+    Use :class:`artiq.coredevice.mirny.Almazny` for Almazny hardware v1.1 and earlier.
+
+    :param cpld_device: Mirny CPLD device name
+    :param channel: channel index (0-3)
+    """
+
+    def __init__(self, dmgr, cpld_device, channel):
+        self.channel = channel
+        self.cpld = dmgr.get(cpld_device)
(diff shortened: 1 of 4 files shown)


Spaqin:
Well, it may break compatibility with previous device_dbs and JSON files a little bit - but I would suggest calling the new ``AlmaznyChannel`` just ``Almazny``, and renaming the old one to ``AlmaznyLegacy``. I think the new name is a bit confusing and on the first sight doesn't really explain the difference between the two.

Maybe keep the old Almazny code as-is for release-7 as stability is priority there, but break the compatibility and bring in a better name for 8+?

artiq/coredevice/almazny.py
```diff
@@ -0,0 +1,68 @@
...
+
+        :param mu: channel state in machine units.
+        """
+        self.cpld.write_ext(
+            addr=0xc + self.channel, length=8, data=mu, ext_div=32)
+
+    @kernel
+    def set(self, att, enable, led=False):
+        """
+        Set attenuation, RF switch, and LET state (SI units).
```
Spaqin, line 62: LED state

artiq/coredevice/coredevice_generic.schema.json
```diff
@@ -511,6 +511,10 @@
                         "almazny": {
                             "type": "boolean",
                             "default": false
+                        },
```
Spaqin, line 514: Maybe merge both Almaznys together and add an optional ``hw_rev`` parameter, with default being legacy to keep JSON compatibility?

jordens: That would be pretty complicated. You'd need to define what "legacy" actually means and translate that into code. This here is straight forward and maintains compatibility.

jordens:
> > Maybe keep the old Almazny code as-is for release-7
> 
> I'm not sure if we need this on release-7 at all anyway, Almazny is relatively a niche card and people who need the new version could perhaps just use 8+?

Almazny is not niche at all.

jordens:
My preference would be to tackle cleanup (phasing out, renaming, removing, fixing) of the legacy Almazny code in a different PR and let this stand alone.

PR #2068, opened 2023-03-23
WIP: Draft prototype of custom example EEM module

### LINKED ISSUE #1816: Better Support for Custom EEMs
# ARTIQ Feature Request

There should be a simpler method for adding custom EEM modules, without requiring acceptance into ARTIQ mainline (e.g. plugins for EEMs). Sort of a follow-up to #1403, which was accepted but had a regression in ARTIQ 6 with the requirement of a JSON file to match the schema.

### Use-Case (actual case at Duke)

Group X creates a custom EEM module to interface with a custom HW device Y. Several internal groups want to use the same software, but Y hasn't been open-sourced, so it's not appropriate for ARTIQ. The JSON board description makes sense, b/c most of the hardware is standard Kasli EEMs, but for the Y EEM. Such an issue leads to PRs like #1800, which have the downside that they are slow to be integrated, require outside approval for something that's slightly experiment-specific, and typically require waiting for a new ARTIQ version before being able to be used.

The main issue is that there's not a clear path for adding out-of-tree EEM support, and it feels that adding out-of-tree EEM support is made extra difficult by the requirement that JSON files match the standard ARTIQ Kasli schema.

Proposed solutions (some/any combination of these would be great):
1. Allow disabling checking the JSON schema against file (maybe via environment variable)
2. Allow changing the JSON schema to a different schema (again, via environment variable/option flag to ``kasli_generic``?)
3. Document how/what needs changed to accommodate outside EEMs.

## Additional context

As a partial attempt to do the documentation suggested above:
* Add custom gateware module (optional, depends on use case)
* Add custom EEM class (sub-classing ``EEM``). Signature:
```python
class CustomEEM(artiq.gateware.eem.EEM):
    @staticmethod
    def io(*args, **kwargs, iostandard: migen.IOStandard = artiq.gateware.eem.default_iostandard) -> Sequence["Pad Assignments"]:
         """Define names for EEM IO pins to add to the available pins that the Kasli has available."""

    @classmethod
    def add_std(cls, target, *args, **kwargs) -> None:
        """Add the ARTIQ RTIO PHYs for your (custom) gateware to the target module."""
        cls.add_extension(target, *args, **kwargs) # calls CustomEEM.io(*args, **kwargs) internally
        # request IO pins that were added in CustomEEM.io()
        # add submodules & RTIO channels for each PHY that you're adding
        
```
* Add custom peripheral interpretation function to add the ``CustomEEM`` for a given set of EEM slots. Example
```python
def peripheral_custom(module, peripheral: Dict[str, Any]) -> None:
     # interpret peripheral arguments
     # add CustomEEM to ARTIQ SOC
     CustomEEM.add_std(module, *args, **kwargs)
```
* Add custom peripheral interpretation to kasli EEM peripheral module lookup dictionary: ``artiq.gateware.eem_7series.peripheral_processors["custom_eem"] = peripheral_custom``
* Write JSON schema for your custom peripheral.
* Modify JSON schema used by ``kasli_generic.py`` to validate the Kasli system description file. Example:
```python
import mergedeep
import artiq.coredevice.jsondesc as artiq_jsondesc

mergedeep.merge(artiq_jsondesc.schema, custom_eem_schema, strategy=mergedeep.Strategy.TYPESAFE_ADDITIVE)
```

All of the above can be seen (less well-organized) in https://github.com/drewrisinger/entangler-core/blob/634ed112649dd1e26b48d92700a07321231aa86c/entangler/kasli_generic.py if users would like an example.

thomasfire:
# ARTIQ Pull Request

### Related Issue

Closes #1816 

## Type of Changes

|   | Type |

### Licensing

### DIFF
--- a/modules/example.py
+++ b/modules/example.py
@@ -1,0 +1,92 @@
+"""
+Draft prototype of custom example EEM module
+"""
+
+from artiq.gateware.eem import _EEM, default_iostandard
+from artiq.frontend.artiq_ddb_template import PeripheralManager
+from artiq.language import kernel
+
+
+# to be added to the artiq/gateware/eem.py
+class Example(_EEM):
+    @classmethod
+    def add_std(cls, target, eem, iostandard=default_iostandard):
+        cls.add_extension(target, eem, iostandard=iostandard)
+
+    @staticmethod
+    def io(eem, iostandard):
+        return [((iostandard(eem),),)]
+
+
+# to be added to the artiq/gateware/eem_7series.py
+def peripheral_example(module, peripheral, **kwargs):
+    Example.add_std(module, peripheral, **kwargs)
+
+
+# to be added to the artiq/gateware/eem_7series.py/peripheral_processors dictionary
+peripheral_processors = {
+    "example": peripheral_example
+}
+
+# to be appended to the artiq/coredevice/coredevice_generic.schema.json/definitions/peripheral/allOf array
+device_schema = """
+{
+    "title": "Example",
+    "if": {
+        "properties": {
+            "type": {
+                "const": "example"
+            }
+        }
+    },
+    "then": {
+        "properties": {
+            "ports": {
+                "type": "array",
+                "items": {
+                    "type": "integer"
+                },
+                "minItems": 1,
+                "maxItems": 3
+            }, 
+            "arg1": {
+                "type": "string"
+            }
+        },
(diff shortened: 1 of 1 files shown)


thomasfire:
> This looks more like a tutorial for the forum.

Well for now I just want to know if such format is desirable. From this point the implementation should not be much complex.

PR #2122, opened 2023-06-27
kasli-soc: change initialization of GTX transceivers.

den512is:
# ARTIQ Pull Request

## Description of Changes
Change of enabling IBUFDS_GTE2. 
Separates CPLL and transceiver reset. 

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

### Licensing


den512is:
By default, IBUFDS_GTE2 is in low power mode (CEB = 1). When the CPU starts the clock switching process, IBUFDS_GTE2 is enabled from the system clock domain, but then the system clock domain is reset. IBUFDS_GTE2 must remain enabled and the tx_init.done signal use for this. IBUFDS_GTE2 must remain enabled and the tx_init.done signal is used for this. If the CPLL and the transceiver are not in the reset state while IBUFDS_GTE2 is enabling, then the transceiver cannot enter the TXPHALIGNDONE state, despite a reset applied later. The tx_init.done signal never becomes 1 and IBUFDS_GTE2 turn off after system clock domain reset. 

I think resetting the CPLL with the BruteforceClockAligner module can sometimes cause the same issues.

Why the behavior of the CPLL and the transceiver depends on the internal termination of IBUFDS_GTE2 I don't know.

>But why does this PR involve the CPLL parameters?

In UG476 about CPLL_LOCK_CFG and CPLL_INIT_CFG said "The recommended value from the 7 Series FPGAs Transceivers Wizard should be used."
These values are present in the IBERT example that was used to test the kasli-soc hardware.

SquidDev:
Just to confirm that this fixes the issues that we were seeing last month with the SYS CLK not switching. Thank you very much!

PR #2126, opened 2023-07-05
associate unit, precision, scale metadata with dataset

SimonRenblad:
### Description

See #2115 for description and comments. Resubmitted PR due to minor merge conflict.


SimonRenblad:
Latest updates / clarifications:

+ `set_dataset` API changed `ndecimals` to `precision`. This will be changed in `NumberValue` to keep consistency.
+ `get_dataset_with_dimension` was dropped in favor of a separate `get_dataset_metadata` that returns only the metadata, as suggested by feedback
+ Metadata will no longer be added to HDF5 or lmdb unless explicitly provided
+ default value for `precision` changed from `2` to `None`. Uses the Numpy `unique` option to perform formatting instead. This should be consistent for both arrays and scalars.
+ `scale` will not be generated unless explicitly provided. The formatting function will still infer based on prefix if scaling needs to be performed. Can still override this by providing `scale`.
+ Separated value and metadata in order to make `lmdb` migrations easier and for better backward compat with applets, client and unit tests. 
+ Modified migration script in release notes to migrate persistent dbs.

For details, see the description in #2115

artiq/language/environment.py
```diff
@@ -337,21 +337,33 @@ def setattr_device(self, key):
...
                     broadcast=False, persist=False, archive=True):
         """Sets the contents and handling modes of a dataset.
 
         Datasets must be scalars (``bool``, ``int``, ``float`` or NumPy scalar)
         or NumPy arrays.
 
+        :param unit: A string representing the unit of the value.
+        :param scale: A numerical scaling factor by which the displayed value is
+            multiplied when referenced in the experiment.
+        :param precision: The precision a UI should use.
```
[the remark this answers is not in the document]
SimonRenblad: Let me make it a bit clearer. The precision should be the same as numpy's :+1:

RELEASE_NOTES.rst
```diff
@@ -40,14 +40,16 @@ Highlights:
...
+      txn.put(key.encode(), pyon.encode((value, {})).encode())
   new.close()
 
 Breaking changes:
 
 * ``SimpleApplet`` now calls widget constructors with an additional ``ctl`` parameter for control
   operations, which includes dataset operations. It can be ignored if not needed. For an example usage,
   refer to the ``big_number.py`` applet.
+* Metadata such as units and precision can now be associated with a dataset. 
+  The ``(value, {})`` is needed to enable metadata in persistent datasets.
```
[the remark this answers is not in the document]
SimonRenblad: removing :+1:

artiq/master/databases.py
```diff
@@ -83,13 +90,18 @@ def update(self, mod):
...
             if key in self.data.raw_view:
                 persist = self.data.raw_view[key][0]
             else:
                 persist = False
-        self.data[key] = (persist, value)
+        if metadata is None:
+            if key in self.data.raw_view:
+                metadata = self.data.raw_view[key][2]
+            else:
+                metadata = {}
```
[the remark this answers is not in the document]
SimonRenblad: I will look it over again :+1:

RELEASE_NOTES.rst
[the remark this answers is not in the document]
dnadlinger: For a breaking change like this, it would be very useful to document what the necessary changes in user code are (in this case, to clients that consume the dataset db using the `sync_struct` interface).

artiq/language/environment.py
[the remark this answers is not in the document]
dnadlinger: I'd very much prefer it not to refer to NumPy, though. I am a long-time user of NumPy and have taught courses using it, yet wouldn't know off the top of my head what the exact semantics are there.

dnadlinger:
Reproducing my comments, as I hadn't noticed that the line notes had been updated in the meantime:

In terms of the release notes, for a breaking change like this, it would be very useful to document what the necessary changes in user code are (in this case, to clients that consume the dataset db using the sync_struct interface).

artiq/master/worker_db.py
```diff
@@ -111,19 +111,20 @@ def __init__(self, ddb):
         self._broadcaster = Notifier(dict())
         self.local = dict()
         self.archive = dict()
+        self.metadata = dict()
 
         self.ddb = ddb
         self._broadcaster.publish = ddb.update
 
-    def set(self, key, value, broadcast=False, persist=False, archive=True):
+    def set(self, key, value, metadata={}, broadcast=False, persist=False, archive=True):
```
dnadlinger, line 119: When adding a new argument to a parameter list with default values, it should be added to the end (or the arguments made keyword-only) to avoid backwards-compatibility issues.

I'd also make this default to `None` and explicitly implement the default value in an `if metadata is None` check, as I've bitten one too many times by the fact that default value instances are shared between all callers across the program lifetime in Python (e.g. here all `self.metadata[key]` entries would refer to the same dict, and somebody could later fetch and change one of them, meaning to only modify that one entry).

artiq/language/environment.py
```diff
@@ -405,6 +420,30 @@ def get_dataset(self, key, default=NoDefault, archive=True):
...
+        """Returns the metadata of a dataset.
+         
+        Returns dictionary with items describing the dataset, including the units, 
+        scale and precision. 
+
+        This function is used to get additional information for displaying the dataset.
+
+        Returns a dictionary which may contain zero or more of the items:
+        
+        'unit' - Physical unit associated with dataset.
```
dnadlinger, line 433: These should probably match the respective documentation in `set_dataset` (or just link to it).

SimonRenblad: Linking to it sounds good.

SimonRenblad:
Regarding the renaming of `short_format` it may be a good idea to push this to future PR for metadata support in client and browser because they both use `short_format`.

artiq/tools.py
```diff
@@ -54,20 +55,40 @@ def elide(s, maxlen):
...
+    elif t is np.ndarray:
+        v_t = np.divide(v, scale)
+        v_str = np.array2string(v_t,
+                           max_line_width=1000,
+                           precision=precision,
+                           suppress_small=True,
+                           separator=', ',
+                           threshold=4,
+                           edgeitems=2,
+                           floatmode='maxprec')
```
dnadlinger, line 87: Indentation is wonky.

dnadlinger:
As a side note, the NumPy `precision` behaviour is pretty questionable:
```py
>>> np.set_printoptions(precision=1)
>>> np.array([0.1 * np.pi])
array([0.3])
>>> np.array([0.01 * np.pi])
array([0.])
>>> np.array([0.001 * np.pi])
array([0.])
>>> np.array([0.0001 * np.pi])
array([0.])
>>> np.array([0.00001 * np.pi])
array([3.1e-05])
>>> np.array([0.000001 * np.pi])
array([3.1e-06])
```

Not much we can (or rather, should) do about that, though.

artiq/language/environment.py
```diff
@@ -337,21 +337,36 @@ def setattr_device(self, key):
...
+                    unit=None, scale=None, precision=None,
                     broadcast=False, persist=False, archive=True):
         """Sets the contents and handling modes of a dataset.
 
         Datasets must be scalars (``bool``, ``int``, ``float`` or NumPy scalar)
         or NumPy arrays.
 
+        :param unit: A string representing the unit of the value.
+        :param scale: A numerical scaling factor by which the displayed value is
+            multiplied when referenced in the experiment.
```
dnadlinger, line 350: This should unambiguously describe what happens here, i.e. that it is inferred for the given unit by default, and only needs to be specified if the unit is not known, or a custom scaling is desired. It might be worth linking to the `NumberValue` documentation here as well (or the other way round).

SimonRenblad: As far as I remember it is copy-pasted from the NumberValue docstring. Some updated docstrings on both may be worthwhile.

SimonRenblad:
For scalar values it should be consistent since we use `format_float_positional`. Anyways since these considerations are purely for display purposes, it would be relatively easy to make changes later on.

artiq/master/worker_db.py
dnadlinger: People might have ARTIQ forks/plugins/… – you are right that we don't care about staying backwards-compatible here, but it still good engineering practice to ensure any callsites that need to be updated fail loudly (e.g. by removing the default values as you suggested, or making the arguments keyword-only).

SimonRenblad:
> In terms of the release notes, for a breaking change like this, it would be very useful to document what the necessary changes in user code are (in this case, to clients that consume the dataset db using the sync_struct interface).

User code that adds additional datasets manually would be affected. AFAIK mutations on existing datasets should remain unchanged since the path is the same.

SimonRenblad:
> What do you mean by that?

Let me clarify. Unless I have missed something, the dictionary object made accessible through DatasetDB used to have this structure:

```
data = {
    key: (persist, value),
    ...
}
```

My proposal changes it in this way:

```
data = {
    key: (persist, value, metadata),
    ...
}
```
Therefore:
+ All modifications to the `value` would still have the same path `key` -> `1` 
+ If a user, instead of using the convenience function `set` uses the `update` function to add an item to `data` (adding a dataset) they would need to be careful to include metadata `(persist, value, metadata)` as opposed to the previous `(persist, value)`. 

I have not verified this yet, will do right now.

SimonRenblad:
I thought it was exposed through the applets? Let me double check. If it is not user exposed then its a non-issue and no need to document.

SimonRenblad:
Then no worries :+1:

artiq/master/worker_db.py
SimonRenblad: removed default values

SimonRenblad:
Suggest to leave any changes of `scale` docstring to a later PR as it also affects `NumberValue`

SimonRenblad:
@sbourdeauducq PR ready for merge, pending any additional feedback

dnadlinger:
> The applets wrap it. And yes the wrapper will need an update.

It's, at best, a leaky wrapper, as the `mods` will contain the change. For instance, `ndscan` will break. Some of our other tooling subscribes directly to the `sync_struct` interface too.

artiq/language/environment.py
```diff
@@ -337,21 +337,36 @@ def setattr_device(self, key):
...
+        :param precision: The maximum number of digits to print after the
+            decimal point. Set ``precision=None`` to print as many digits as 
+            necessary to uniquely specify the value. Uses IEEE unbiased rounding.
+            Uses the ``numpy.printoptions`` interpretation of precision.
```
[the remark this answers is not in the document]
SimonRenblad: I have removed reference to numpy for now. Found that the ways I phrased it generally made the description more confusing, and the rest of the description should sufficiently describe the behavior. I suggest that if a reference is added to numpy, it is done in a later PR.

artiq/master/worker_db.py
```diff
@@ -111,26 +111,31 @@ def __init__(self, ddb):
...
         elif key in self._broadcaster.raw_view:
             del self._broadcaster[key]
 
         if archive:
             self.local[key] = value
         elif key in self.local:
             del self.local[key]
+        
+        if metadata is None:
+            metadata = {}
```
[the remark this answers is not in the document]
SimonRenblad: Removed, this check is not necessary

SimonRenblad:
Updated the docstring for `scale` in `set_dataset`, but did not update it in `NumberValue`, as I believe it may be more relevant in a separate PR. Therefore there is a slight inconsistency there at the moment.

PR #2140, opened 2023-07-20
big_number: Make widget to consistently work with dataset metadata

fsagbuya:
# ARTIQ Pull Request

## Description of Changes
Created patch to make the `big_number` applet to work well with dataset metadata #2126 . 

Major Changes:
- `QLCDNumber` changed to `QLabel`
- Allow editing of value and unit

## Type of Changes

|   | Type |

### All Pull Requests

- [x] Check the copyright situation of your changes and sign off your patches (`git commit --signoff`, see [copyright](https://github.com/m-labs/artiq/blob/master/CONTRIBUTING.rst#copyright-and-sign-off)).

### Code Changes

- Tested both on widget's edit and display

### Git Logistics

  ```
  
  ```

### Licensing

### DIFF
--- a/artiq/applets/big_number.py
+++ b/artiq/applets/big_number.py
@@ -1,10 +1,16 @@
 #!/usr/bin/env python3
+import re
+import logging
+import numpy as np
 
 from PyQt5 import QtWidgets, QtCore, QtGui
+
+from artiq.language import units
+from artiq.tools import short_format, scale_from_metadata
 from artiq.applets.simple import SimpleApplet
 
 
-class QResponsiveLCDNumber(QtWidgets.QLCDNumber):
+class QResponsiveLabel(QtWidgets.QLabel):
     doubleClicked = QtCore.pyqtSignal()
 
     def mouseDoubleClickEvent(self, event):
@@ -26,54 +32,89 @@ class NumberWidget(QtWidgets.QStackedWidget):
         QtWidgets.QStackedWidget.__init__(self)
         self.dataset_name = args.dataset
         self.ctl = ctl
-
-        self.lcd_widget = QResponsiveLCDNumber()
-        self.lcd_widget.setDigitCount(args.digit_count)
-        self.lcd_widget.doubleClicked.connect(self.start_edit)
-        self.addWidget(self.lcd_widget)
-
-        self.edit_widget = QCancellableLineEdit()
-        self.edit_widget.setValidator(QtGui.QDoubleValidator())
-        self.edit_widget.setAlignment(QtCore.Qt.AlignRight)
-        self.edit_widget.editCancelled.connect(self.cancel_edit)
-        self.edit_widget.returnPressed.connect(self.confirm_edit)
-        self.addWidget(self.edit_widget)
+        self.value = None
+        self.metadata = {}
 
         font = QtGui.QFont()
         font.setPointSize(60)
-        self.edit_widget.setFont(font)
 
-        self.setCurrentWidget(self.lcd_widget)
+        self.display_widget = QResponsiveLabel()
+        self.display_widget.setFont(font)
+        self.display_widget.setAlignment(
+            QtCore.Qt.AlignRight | QtCore.Qt.AlignVCenter)
+        self.display_widget.doubleClicked.connect(self.start_edit)
+        self.addWidget(self.display_widget)
+
+        self.edit_widget = QCancellableLineEdit()
+        self.edit_widget.setFont(font)
+        self.edit_widget.setAlignment(
+            QtCore.Qt.AlignRight | QtCore.Qt.AlignVCenter)
+        self.edit_widget.editCancelled.connect(
+            lambda: self.setCurrentWidget(self.display_widget))
+        self.edit_widget.returnPressed.connect(self.confirm_edit)
+        self.addWidget(self.edit_widget)
+
+        input_regexp = r"(([+-]?\d+(\.\d*)?|\.\d+)([eE][+-]?\d+)?)"
+        self.regexp = "^" + input_regexp + r"( [a-zA-Z]*)?$"
+
+        self.setCurrentWidget(self.display_widget)
 
     def start_edit(self):
-        # QLCDNumber value property contains the value of zero
-        # if the displayed value is not a number.
-        self.edit_widget.setText(str(self.lcd_widget.value()))
+        if self.value is None:
+            self.edit_widget.setText("---")
+        else:
+            unit = self.metadata.get("unit", "")
+            scale = int(scale_from_metadata(self.metadata))
+            if scale == 1.0:
+                scale = int(scale)
+            if isinstance(self.value, int) and isinstance(scale, int):
+                value = self.value//scale
+            else:
+                value = self.value / scale
+            self.edit_widget.setText(f"{value} {unit}".strip())
         self.edit_widget.selectAll()
         self.edit_widget.setFocus()
         self.setCurrentWidget(self.edit_widget)
 
     def confirm_edit(self):
-        value = float(self.edit_widget.text())
-        self.ctl.set_dataset(self.dataset_name, value)
-        self.setCurrentWidget(self.lcd_widget)
-
-    def cancel_edit(self):
-        self.setCurrentWidget(self.lcd_widget)
+        entry = re.match(self.regexp, self.edit_widget.text())
+        if entry is not None:
+            value = entry.group(1)
+            if np.issubdtype(self.type, np.integer):
+                value = self.type(int(float(value))) 
+            elif np.issubdtype(self.type, np.floating):
+                value = self.type(value)
+            elif self.type in [int, float]:
+               value = self.type(float(value)) 
+            else:
+                logging.warning("Unsupported data type:", self.type)
+                return
+            unit = entry.group(5).strip() if entry.group(5) else ""
+            scale = scale_from_metadata(self.metadata)
+            if scale == 1.0:
+                scale = int(scale)
+            precision = self.metadata.get("precision")
+            self.ctl.set_dataset(self.dataset_name, value*scale, 
+                                unit, scale, precision)
+            self.setCurrentWidget(self.display_widget)
+        else:
+            self.setCurrentWidget(self.display_widget)
+            
 
     def data_changed(self, value, metadata, persist, mods):
+        self.metadata = metadata.get(self.dataset_name, {})
         try:
-            n = float(value[self.dataset_name])
+            self.value = value[self.dataset_name]
+            self.type = type(self.value)
+            n = short_format(self.value, self.metadata)
         except (KeyError, ValueError, TypeError):
             n = "---"
-        self.lcd_widget.display(n)
+        self.display_widget.setText(n)
 
 
 def main():
     applet = SimpleApplet(NumberWidget)
     applet.add_dataset("dataset", "dataset to show")
-    applet.argparser.add_argument("--digit-count", type=int, default=10,
-                                  help="total number of digits to show")
     applet.run()
 
 if __name__ == "__main__":


artiq/applets/big_number.py
```diff
@@ -55,7 +57,8 @@ def start_edit(self):
 
     def confirm_edit(self):
         value = float(self.edit_widget.text())
```
SimonRenblad, line 59: be careful here, the `short_format` function may scale the value depending on the units provided, and when editing this needs to be handled instead of just parsing as float.

SimonRenblad (commented):
+ Could include to allow editing of the units as well, might as well since you need to check the edited input before updating the dataset
+ With precision enabled, a clean update is not really possible (changing a digit while keeping everything else constant), but there is not much to do about that

artiq/applets/big_number.py
```diff
@@ -64,6 +67,8 @@ def cancel_edit(self):
     def data_changed(self, value, metadata, persist, mods):
         try:
             n = float(value[self.dataset_name])
+            self.metadata = metadata.get(self.dataset_name, {})
+            n = short_format(n, self.metadata).split(" ")[0]
```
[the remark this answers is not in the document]
fsagbuya: The widget right now uses `QLCDNumber` which only display numeric values. Even if we try to have the unit be edited in the edit mode of the widget, when it switch to the display(lcd), unit will be omitted.

SimonRenblad (changes requested):
Current implementation does not take into account precision (num of decimal places) and assumes that it should stay the same as before. In order to mirror the dataset value in the dashboard, it may be worth to update the precision post edit here. This is unfortunately also problematic because the precision could be higher than the number of decimals shown. Perhaps only change precision if it needs to be increased?

fsagbuya:
> Current implementation does not take into account precision (num of decimal places) and assumes that it should stay the same as before. In order to mirror the dataset value in the dashboard, it may be worth to update the precision post edit here. This is unfortunately also problematic because the precision could be higher than the number of decimals shown. Perhaps only change precision if it needs to be increased?

Just retained the `precision` from the underlying dataset since it can only be edited from the gui. Edit widget now reflects the same raw value from the gui. While the display widget shows the formatted value.

artiq/applets/big_number.py
```diff
@@ -26,47 +27,78 @@ def __init__(self, args, ctl):
...
+        self.display_widget.setFont(font)
         self.edit_widget.setFont(font)
 
-        self.setCurrentWidget(self.lcd_widget)
+        self.setCurrentWidget(self.display_widget)
+
+    def get_unscaled_format(self):
+        scale = self.metadata.get("scale", None)
+        unit = self.metadata.get("unit", "")
+        unit_scale = scale_from_metadata({"unit": unit})
```
SimonRenblad, line 56: can just pass self.metadata straight into scale_from_metadata and divide by it I believe.

artiq/applets/big_number.py
```diff
@@ -26,47 +27,78 @@ def __init__(self, args, ctl):
...
+            value = self.value/scale
+        else:
+            value = self.value/unit_scale
+        return f"{value} {unit}".strip()
 
     def start_edit(self):
-        # QLCDNumber value property contains the value of zero
-        # if the displayed value is not a number.
-        self.edit_widget.setText(str(self.lcd_widget.value()))
+        if self.value == 0:
```
SimonRenblad, line 64: use a different default value, otherwise we cannot display `0`

artiq/applets/big_number.py
```diff
@@ -26,47 +27,78 @@ def __init__(self, args, ctl):
...
 
         font = QtGui.QFont()
         font.setPointSize(60)
+        self.display_widget.setFont(font)
         self.edit_widget.setFont(font)
 
-        self.setCurrentWidget(self.lcd_widget)
+        self.setCurrentWidget(self.display_widget)
+
+    def get_unscaled_format(self):
```
SimonRenblad, line 53: this looks like a duplicate of short_format, why not use that? `unscaled` in my opinion is not the best term to use here

artiq/applets/big_number.py
```diff
@@ -26,47 +27,78 @@ def __init__(self, args, ctl):
...
+        try:
+            value = float(text[0])
+        except ValueError:
+            self.display_widget.setText("---")
+            self.setCurrentWidget(self.display_widget)
+            return
+        unit = text[1] if len(text) > 1 else ""
+        scale = scale_from_metadata({"unit": unit})
+        value = value*scale
+        precision = self.metadata.get("precision", None)
```
SimonRenblad, line 83: see if you can increase the precision if the user inputs more decimal points than previously. Say precision is 2 and the value is 5.44 originally, and then the user changes it to 5.654 in this widget. In this case I would recommend increasing the precision to 3 such that the display mirrors appropriately. Precision is only a display parameter anyways

artiq/applets/big_number.py
```diff
@@ -26,47 +27,78 @@ def __init__(self, args, ctl):
...
     def cancel_edit(self):
-        self.setCurrentWidget(self.lcd_widget)
+        self.setCurrentWidget(self.display_widget)
 
     def data_changed(self, value, metadata, persist, mods):
+        self.metadata = metadata.get(self.dataset_name, {})
         try:
-            n = float(value[self.dataset_name])
+            self.value = float(value[self.dataset_name])
+            n = short_format(self.value, self.metadata)
```
SimonRenblad, line 95: related to my previous comment, why not store `n` as a member variable and pass it straight to the edit widget? just need to make sure the submission converts everything properly

SimonRenblad (changes requested):
Some duplicate code, would recommend checking the behavior of `short_format` carefully because it looks like it has been duplicated here

artiq/applets/big_number.py
SimonRenblad: edit: not a duplicate, but name still needs change :+1:

artiq/applets/big_number.py
SimonRenblad: edit: need slightly different display, so cannot do this.

artiq/applets/big_number.py
fsagbuya: Since the `edit_widget` reflects the raw value of the input (just like the current value in the edit dataset dialog box), changing the precision that way will be a bit odd. For example, if the input value of the dataset is 123.456 and the precision is 2, the output shown will be 123.46 in both dataset and display if big_number. However if we enter the edit mode, we will see the same input value 123.456 stored, and editing this, say 123.4567 will change the precision to 4. Since this adaptive precision is not the current behavior of the dataset, I think it would be fitting to leave the precision setting to the dataset alone for more consistency.

SimonRenblad: agree :+1:

artiq/applets/big_number.py
```diff
@@ -26,47 +27,77 @@ def __init__(self, args, ctl):
...
+        try:
+            value = float(text[0])
+        except ValueError:
+            self.display_widget.setText("---")
+            self.setCurrentWidget(self.display_widget)
+            return
+        unit = text[1] if len(text) > 1 else ""
+        scale = scale_from_metadata({"unit": unit})
+        value = value*scale
+        precision = self.metadata.get("precision", None)
```
SimonRenblad, line 82: None is redundant here since it is already the default. That being said, this is just a nitpick

SimonRenblad (changes requested):
Some minor fixes requested

artiq/applets/big_number.py
```diff
@@ -26,47 +27,71 @@ def __init__(self, args, ctl):
...
+            return
+        unit = text[1] if len(text) > 1 else ""
+        scale = scale_from_metadata({"unit": unit})
+        value = value*scale
+        precision = self.metadata.get("precision")
+        self.ctl.set_dataset(self.dataset_name, value, unit=unit,
+                             precision=precision)
+        self.setCurrentWidget(self.display_widget)
 
     def cancel_edit(self):
```
SimonRenblad, line 81: Considering this is a one-liner, I propose either: A) rename it to something like `end_edit` (or something that does not imply cancel, can find a better name). Then use this function in place of all `self.setCurrentWidget(self.display_widget)` or B) remove this function and in the editCancelled pass a lambda instead.

SimonRenblad (changes requested):
minor fix requested

artiq/applets/big_number.py
fsagbuya: Updated with option B

artiq/applets/big_number.py
```diff
@@ -26,54 +27,74 @@ def __init__(self, args, ctl):
...
         try:
-            n = float(value[self.dataset_name])
+            self.value = float(value[self.dataset_name])
+            n = short_format(self.value, self.metadata)
         except (KeyError, ValueError, TypeError):
             n = "---"
-        self.lcd_widget.display(n)
+        self.display_widget.setText(n)
+
+        if self.currentWidget() is self.edit_widget:
```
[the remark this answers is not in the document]
fsagbuya: This happens when user is still inside edit_widget but tried to go somewhere else without cancelling the edit or pressing enter (e.g. tried to change the value of dataset in gui). Proposed solution updates the edit_widget to shift to display when data_changed happens while still inside edit_widget. Will just transfer this patch to a separate PR if we want to retain this.

artiq/applets/big_number.py
```diff
@@ -26,54 +27,74 @@ def __init__(self, args, ctl):
...
         self.setCurrentWidget(self.edit_widget)
 
     def confirm_edit(self):
-        value = float(self.edit_widget.text())
-        self.ctl.set_dataset(self.dataset_name, value)
-        self.setCurrentWidget(self.lcd_widget)
-
-    def cancel_edit(self):
-        self.setCurrentWidget(self.lcd_widget)
+        text = self.edit_widget.text().split() or "---"
```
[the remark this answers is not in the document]
fsagbuya: Replaced with an `if` statement

artiq/applets/big_number.py
```diff
@@ -26,54 +27,74 @@ def __init__(self, args, ctl):
...
+        try:
+            value = float(text[0])
+        except ValueError:
+            self.display_widget.setText("---")
+            self.setCurrentWidget(self.display_widget)
+            return
+        unit = text[1] if len(text) > 1 else ""
```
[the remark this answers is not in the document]
fsagbuya: Renamed `text` to `entry`
if len(entry) > 2, unit will just get the second element and ignore the rest.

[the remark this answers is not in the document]
SimonRenblad: > Renamed `text` to `entry` if len(entry) > 2, unit will just get the second element and ignore the rest.

probably easiest to cancel entry if invalid case like this occurs. If I put in '15 m V' clearly I do not want it to say '15 m' which is what happens atm.

artiq/applets/big_number.py
```diff
@@ -26,54 +27,76 @@ def __init__(self, args, ctl):
...
+            self.reset_display()
+            return
+        try:
+            value = float(entry[0])
+        except ValueError:
+            self.reset_display()
+            return
+        unit = entry[1] if len(entry) > 1 else ""
+        scale = scale_from_metadata({"unit": unit})
+        value = value*scale
```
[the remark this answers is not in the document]
SimonRenblad: We override the scaling here as an explicit 'scale' parameter has not been considered in the widget, so anytime it is edited it will use the default unit system for overriding. This is dodgy in my opinion, would be good to consider a solution to this @fsagbuya.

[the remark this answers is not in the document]
SimonRenblad: Line 77 can probably be better rewritten using the 'getattr(units, unit)' way used in many other places. The 'scale_from_metadata' helper is used for resolving conflicts between the default unit scaling and a predefined scale.

artiq/applets/big_number.py
```diff
@@ -26,54 +30,70 @@ def __init__(self, args, ctl):
...
 
     def start_edit(self):
-        # QLCDNumber value property contains the value of zero
-        # if the displayed value is not a number.
-        self.edit_widget.setText(str(self.lcd_widget.value()))
+        if self.value is None:
+            self.edit_widget.setText("---")
+        else:
+            unit = self.metadata.get("unit", "")
+            scale = getattr(units, unit, 1)
```
SimonRenblad, line 67: Ignorant of any scale param provided in metadata... will not work on any custom defined units. Let's say I defined an arbitrary unit 't' (name irrelevant) and set the scale to 1000 in browser, this will display 1000x the value in the applet TextEdit as a consequence, and I will lose the scale metadata when opening in dashboard. Use `scale_from_metadata` to handle this instead.

artiq/applets/big_number.py
```diff
@@ -26,54 +30,70 @@ def __init__(self, args, ctl):
...
-        value = float(self.edit_widget.text())
-        self.ctl.set_dataset(self.dataset_name, value)
-        self.setCurrentWidget(self.lcd_widget)
-
-    def cancel_edit(self):
-        self.setCurrentWidget(self.lcd_widget)
+        entry = re.match(self.regexp, self.edit_widget.text())
+        value = float(entry.group(1))
+        unit = entry.group(5).strip() if entry.group(5) else ""
+        scale = getattr(units, unit, 1)
```
SimonRenblad, line 78: Same problem here as mentioned above.

artiq/applets/big_number.py
```diff
@@ -26,54 +30,70 @@ def __init__(self, args, ctl):
...
+        precision = self.metadata.get("precision")
+        self.ctl.set_dataset(self.dataset_name, value*scale, unit=unit,
```
SimonRenblad, line 80: scale? Just pass the old scale straight through, it is not editable in this applet anyways.

artiq/applets/big_number.py
```diff
@@ -26,54 +30,70 @@ def __init__(self, args, ctl):
...
+        self.setCurrentWidget(self.display_widget)
 
     def data_changed(self, value, metadata, persist, mods):
+        self.metadata = metadata.get(self.dataset_name, {})
         try:
-            n = float(value[self.dataset_name])
+            self.value = float(value[self.dataset_name])
+            n = short_format(self.value, self.metadata)
         except (KeyError, ValueError, TypeError):
             n = "---"
```
SimonRenblad, line 90: "---" cannot be submitted, the widget will not exit editing mode. Same with an empty string. If the user cancels at an invalid state, the original dataset should be kept as is.

SimonRenblad (changes requested):
+ `scale` handling is not sufficient at this stage and needs to be ironed out. See comments.
+ More testing needs to be done with invalid dataset types etc to be handled gracefully. The widget locks up too easily in editing mode at the moment.
+ Considering the move away from the LCD display style, does this widget really need to have a Label "mode" and a TextEdit "mode"? Wouldn't a single TextEdit sufficiently replace this behavior? May be worth at least exploring.

artiq/applets/big_number.py
SimonRenblad: I know we have gone back and forth on this, but the problematic thing with using `scale_from_metadata` mentioned in previous comments was also not passing the full metadata obj (only unit). Let's try using the `scale` metadata in this widget, should it be defined, and only override in case of a change of units (in which case only defaults should be permissable IMO). Alternatively we may lock-in the units and make them display only in this widget.

artiq/applets/big_number.py
fsagbuya: What do you mean by this? From what I understand, this is referring to editing a value from the dashboard while also edit mode in widget is currently open, any entry will not be save including the valid ones. There is a patch for this that will be issued in a separate PR.

> This happens when user is still inside edit_widget but tried to go somewhere else without cancelling the edit or pressing enter (e.g. tried to change the value of dataset in gui). Proposed solution updates the edit_widget to shift to display when data_changed happens while still inside edit_widget. Will just transfer this patch to a separate PR if we want to retain this.

fsagbuya:
> * `scale` handling is not sufficient at this stage and needs to be ironed out. See comments.
> * More testing needs to be done with invalid dataset types etc to be handled gracefully. The widget locks up too easily in editing mode at the moment.
> * Considering the move away from the LCD display style, does this widget really need to have a Label "mode" and a TextEdit "mode"? Wouldn't a single TextEdit sufficiently replace this behavior? May be worth at least exploring.

- handling `scale` is now updated, thanks for mentioning that. Edit mode is now more consistent with the underlying dataset value. Please let me know if you have suggestions to make this cleaner.
- see comment on widget lock up
- I think the widget's original purpose is to be a display widget rather than an edit widget. Editing here serves only as an extension. While TextEdit can do both, edit there is more expose to the user. Editing of data should be primarily done in the dashboard or user code, IMO.

artiq/applets/big_number.py
SimonRenblad: No I am not talking about that specifically. Here are the precise steps to reproduce:

1. Set a dataset to a string value (not a number)
2. Call the big_number applet with this dataset as input
3. Click on the "---" to edit it
4. At this point, pressing enter did not leave edit mode, and when deleting all the '-' chars, pressing enter also did not work. I was forced to override the string value with a number in order to exit the applet.

fsagbuya: Updated. Pressing enter when input is invalid will revert to previous valid value or "---" on start. Added a warning log as well.

artiq/applets/big_number.py
```diff
@@ -26,54 +31,73 @@ def __init__(self, args, ctl):
...
     def confirm_edit(self):
-        value = float(self.edit_widget.text())
-        self.ctl.set_dataset(self.dataset_name, value)
-        self.setCurrentWidget(self.lcd_widget)
-
-    def cancel_edit(self):
-        self.setCurrentWidget(self.lcd_widget)
+        entry = re.match(self.regexp, self.edit_widget.text())
+        if entry is not None:
+            value = float(entry.group(1))
```
SimonRenblad, line 76: Datasets can take multiple float / integer types from numpy, this applet will collapse them all to native float. See #2176.

artiq/applets/big_number.py
```diff
@@ -26,54 +31,73 @@ def __init__(self, args, ctl):
...
+            unit = entry.group(5).strip() if entry.group(5) else ""
+            scale = scale_from_metadata(self.metadata)
+            precision = self.metadata.get("precision")
+            self.ctl.set_dataset(self.dataset_name, value*scale,
```
SimonRenblad, line 80: See comment about type casting. Careful with the scale multiplication here.

artiq/applets/big_number.py
```diff
@@ -26,54 +31,73 @@ def __init__(self, args, ctl):
...
+        else:
+            logging.warning("Invalid input")
+            self.setCurrentWidget(self.display_widget)
+            
 
     def data_changed(self, value, metadata, persist, mods):
+        self.metadata = metadata.get(self.dataset_name, {})
         try:
-            n = float(value[self.dataset_name])
+            self.value = float(value[self.dataset_name])
```
SimonRenblad, line 91: See data type casting comment.

artiq/applets/big_number.py
SimonRenblad: Unsure whether this is of concern in this applet, but figured I would comment on it in case.

artiq/applets/big_number.py
```diff
@@ -26,54 +31,73 @@ def __init__(self, args, ctl):
...
+        self.edit_widget.editCancelled.connect(
+            lambda: self.setCurrentWidget(self.display_widget))
         self.edit_widget.returnPressed.connect(self.confirm_edit)
         self.addWidget(self.edit_widget)
 
-        font = QtGui.QFont()
-        font.setPointSize(60)
-        self.edit_widget.setFont(font)
+        float_regexp = r"(([+-]?\d+(\.\d*)?|\.\d+)([eE][+-]?\d+)?)"
+        self.regexp = "^" + float_regexp + r"( [a-zA-Z]*)?$"
```
SimonRenblad, line 57: needs better name

SimonRenblad (changes requested):
+ Logging redundant
+ Potential type casting issue

artiq/applets/big_number.py
fsagbuya: Can you give examples where the type casting to float might fail on this?

SimonRenblad: A `np.float8` type will convert to a `float`. More problematic is an `int` will also be forced to `float`. My recommendation would be to (in the case of an existing dataset) store the type of the value and perform the appropriate type casting before running set_dataset again.

SimonRenblad (changes requested):
+ Issues with types still present, see how this is handled in the dashboard
+ Other types can be displayed, but break when editing. Entering a `dict` will display `dict(1)` but clicking on it creates an error. Suggest to simply disable editing in the case of non-scalar datasets.

fsagbuya:
Please see #2234

PR #2151, opened 2023-07-31
Fix last `except` handler running if no others match on Kasli-SoC

### LINKED ISSUE #2108: Kasli-SoC runs last `except` handler if no others match
# Bug Report

## One-Line Summary

When using a `try`-`except` block on the Kasli-SoC, if none of the type guards match, the last exception handler will be incorrectly run.

### Steps to Reproduce

Set up a Kasli-SoC and run the following experiment:

```python
class ThrowsException(aq.EnvExperiment):
    def build(self):
        self.core: Core = self.get_device("core")

    @kernel
    def run(self):
        x = "Okay"
        try:
            raise RTIOOverflow("Overflow")
        except RTIOUnderflow:
            x = "Underflow"
        print(x)
```

### Expected Behavior

The experiment fails with `Terminating with exception (RTIOOverflow: Overflow)`

The exception succeeds and prints out `"Underflow"`, indicating that the `except` block was hit.

* Operating System: Linux
* ARTIQ version: 75f6bdb6a10accb54b4b3ef1b167f822a8a54caa and [`artiq-zynq@f1ee3a758446a489f43a8fdfdcb1f47fe05b12f8`](https://git.m-labs.hk/M-Labs/artiq-zynq/commit/f1ee3a758446a489f43a8fdfdcb1f47fe05b12f8).
* Hardware involved: Kasli-Soc

### Additional notes
The fallthrough block in the `try`/`except` block [emits an LLVM `resume` instruction](https://github.com/m-labs/artiq/blob/7ad32d903a42ce4687eaf4c4f18c5e6c4a6747ea/artiq/compiler/transforms/llvm_ir_generator.py#L1912):

```
try.dispatch.endif:                               ; preds = %try.dispatch
  resume { i8*, i8* } %.12, !dbg !16
```

Changing this to emit a call to `__artiq_resume` (either by changing the code generator, or adding a `except: raise` block to the above experiment) fixes this bug. However, I'm not familiar enough with either LLVM or ARTIQ's unwinding code to know if this is a sensible fix.

```
try.dispatch.endif:                               ; preds = %try.dispatch
  call void @__artiq_resume() #3, !dbg !18
  unreachable
```

SquidDev:
# ARTIQ Pull Request

## Description of Changes
This changes the compiler to generate a call to `@__artiq_resume` in the `try`/`except` fallthrough block, rather than LLVM's `resume` instruction.

This fixes a block on Kasli-SoC, where the last `except` block is run, if none of the preceding blocks match - see #2108 for further details.

### Related Issue
Closes #2108.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

### All Pull Requests

- [X] Close/update issues.
- [X] Check the copyright situation of your changes and sign off your patches (`git commit --signoff`, see [copyright](../CONTRIBUTING.rst#copyright-and-sign-off)).

### Code Changes

- [x] Test your changes or have someone test them. Mention what was tested and how:
  - With this patch, the experiment in #2108 behaves as expected on both Kasli 1.1 and Kasli-SoC.
  - We've been running this for a little over a month now (with both Kasli-SoC and Kasli 1.1 masters), and not encountered any issues.

### Licensing

### DIFF
--- a/artiq/compiler/transforms/llvm_ir_generator.py
+++ b/artiq/compiler/transforms/llvm_ir_generator.py
@@ -1376,7 +1376,7 @@ class LLVMIRGenerator:
 
         stack_save_needed = False
         for i, arg in enumerate(insn.arguments()):
-            llarg = self.map(arg)            
+            llarg = self.map(arg)
             if isinstance(llarg.type, (ll.LiteralStructType, ll.IdentifiedStructType)):
                 stack_save_needed = True
                 break
@@ -1909,6 +1909,10 @@ class LLVMIRGenerator:
                 self.add_pred(landingpadbb, self.llbuilder.basic_block)
                 self.llbuilder.branch(target)
             else:
-                self.llbuilder.resume(lllandingpad)
+                llinsn = self.llbuilder.call(
+                    self.llbuiltin("__artiq_resume"), [], name=insn.name
+                )
+                llinsn.attributes.add('noreturn')
+                self.llbuilder.unreachable()
 
         return llexn


dnadlinger:
This doesn't seem like the correct fix. The unwinder should not even jump to the landing pad if there are no matching exception types, as there is no cleanup, so surely this just papers over the issue?

dnadlinger:
`_Unwind_Resume` should indeed handle the `ForcedUnwind` use case correctly; the issue here seems to be that the LLVM backend – correctly – assumes that the landing pad should only be executed if at least one of the clauses matches.

SquidDev:
> This doesn't seem like the correct fix. The unwinder should not even jump to the landing pad if there are no matching exception types, as there is no cleanup, so surely this just papers over the issue?

PR #2193, opened 2023-09-12
Shuttler: Add coredevice example code for Shuttler

linuswck:
# ARTIQ Pull Request
This PR adds coredevice example code for shuttler.

## Description of Changes
After the Link connection between kasli and efc with shuttler installed is established, running this example code will:
1. Blink the EFC Board LEDs (L0, L1)
2. Generate 10MHz Sinusoidal Wave in 1V amplitude with 1V voltage offset on all PDQ channels on Shuttler board
    - PDQ output signals are measured on Shuttler

Note: A small delay is added in between each volt and dds config to avoid `RTIO Underflow` error. Removing both `self.led()` and all `delay(10*us)` will reveal this error.

## Type of Changes

|   | Type |

### Licensing


artiq/examples/kasli_shuttler/repository/shuttler.py
```diff
@@ -0,0 +1,77 @@
...
+                self.leds[i].pulse(.1*s)
+                delay(.1*s)
+
+    @kernel
+    def pdq_volt_mu(self, volt):
+        return round(2 ** 13 / 5 * volt)
+
+    @kernel
+    def pdq_freq_mu(self, freq_mhz):
+        return round(float(2) ** 32 / 125 * freq_mhz)
```
Spaqin, line 32: These functions seem useful beyond the example code - should they be a part of the coredevice driver?

occheung: Speaking of this, are we truncating 2 MSBs from Shuttler core?

linuswck: > Speaking of this, are we truncating 2 MSBs from Shuttler core? The `2 ** 13` factor seems to imply this.

Yes. 2 MSBs are truncated from Shuttler core as AD9117 is a 14bit DAC.

dhslichter:
> > 2. Generate 10MHz Sinusoidal Wave in 1V amplitude with 1V voltage offset on all PDQ channels on Shuttler board
> 
> Can you shape it (amplitude/frequency) for a better demo?

Ideally a test waveform that covers all possibilities (amplitude envelope for the sine wave, on top of a time-varying offset, with a phase jump and frequency jump for the sine wave at some point, for example).

artiq/examples/kasli_shuttler/repository/shuttler.py
dhslichter: If you truncate the MSBs from Shuttler core, does it make it hard to detect when the output waveform is digitally clipping? Would it not be better to truncate the 2 LSBs?

dhslichter: In addition, if/when we get around to implementing sigma-delta modulation to increase bit depth, we will want to have more than 14 bits. This works much more naturally if we are truncating the 2 LSBs for 14 bits, and then keeping one or both if we go to 15 or 16 bits using sigma-delta.

dhslichter: @sbourdeauducq the PDQ core should run at a higher resolution than the output, otherwise the truncation errors during the waveform computation from e.g. summing offset with DDS tones will compound and degrade the quality of the output signal. Also if we plan for future sigma-delta we will need even more bits in the PDQ core.

PR #2203, opened 2023-09-15
gui: Transition to Qt6

dnadlinger:
Qt 6.0 was released amost three years ago, and the current release (6.5) is an LTS release. Dependencies such as `qasync` and `pyqtgraph` and the wider ecosystem (e.g. `matplotlib`) transparently support Qt6 as well, and it has been available in Nix and various distros for a while. Thus, transitioning to Qt 6 seems a feasible proposition at this point.

My impetus for this in particular was the fact that Qt 5 isn't actually supported well on recent macOS versions (and in particular, not officially on aarch64, breaking packaging through Nix/Poetry).

Some changes are due to deprecations in Qt 6 which were outright removed already in PyQt, for instance `QRegExp` or the `x()`/`y()` `QMouseEvent` properties. Most of the diff, however, is due to a PyQt change, where enumeration values now no longer being available directly in the parent namespace.

Not yet tested on Windows or Linux/Nix.


dnadlinger:
Ah, seems a few fell through the cracks even though I searched the entire code base, thanks – I'll incorporate these and try using it in anger on Windows this weekend.

As for Wayland/Qt 6, does that work with Qt 5, or are you running on X11 there?

Incidentally, I have applet support without embedding ~ready, as foreign windows are completely broken on macOS, but this of course breaks docking.

dnadlinger:
Integrated the above comments.

@sbourdeauducq Could you check whether your working Qt 5 install uses Wayland as well? Does `-platform xcb` fix it on Qt 6? While Wayland should in principle support composing multiple processes together, I couldn't see any code for that in Qt 5 at a glance either.

dnadlinger:
Fixed artiq_browser startup and file open dialogs – I don't use it, and the constants had escaped my impromptu regexp filters.

As for xcb vs Wayland, could we just `export QT_QPA_PLATF0RM="xcb"`? Or `"xcb;wayland"` and log a warning on startup/fall back on the incoming embed-less implementation? This could be done in `wrapQtApp` via `qtWrapperArgs`.

dnadlinger:
We have been using this in daily use on Windows for a couple days on one experiment now, and there don't seem to be any glaring issues left – except for the fact that the dashboard hangs on exiting. Using `-v`, the following message is logged about once every second:
```
DEBUG:dashboard:asyncio:<_IocpProactor overlapped#=41 result#=0> is running after closing for 1.0 seconds
```
I'm not quite sure which operation this is.

The dashboard closes just fine on macOS.

dnadlinger:
The dashboard closing issue actually does not appear to be specific to Qt 6 at all; it also occurs with current master, and seems to be triggered by using a newer Python version (3.8 fine, 3.10 broken).

dnadlinger:
> We do need docking to work on Linux and Windows.

Oh yes, certainly, but if somebody just _had_ to run on Wayland, this would be a better alternative than having no dashboard at all (same story as on macOS).

dnadlinger:
Applets do work on Windows, but on Python 3.10+, you really need https://github.com/CabbageDevelopment/qasync/commit/5f6580ee3b865dd6a69f52a081c988ea65f4ff4e for things not to crash and burn (unrelated to Qt 6, but this is what I mentioned above).

hartytp:
@sbourdeauducq @dnadlinger what's blocking this from being merged? The qasync issue @dnadlinger mentioned is now fixed upstream.

Spaqin:
> PySide6 or PyQt6?

The differences are cosmetic and seem to be only in the license; since PySide6 is more permissive we can just go with that.

Issue #1497, opened 2020-07-27
compiler: Function return values (?) not life-time-tracked


dnadlinger:
Escape the confines of the ARTIQ compiler escape analysis with this one neat trick - infinite memory corruption at your fingertips!

```py
@kernel
def leak(a):
    return a

class LeakTest(EnvExperiment):
    def build(self):
        self.setattr_device("core")

    @kernel
    def run(self):
        # Fails to compile as it should:
        #     return leak([1, 2, 3])
        # However…
        return leak([1, 2, 3])
```

(This kernel shouldn't compile, but does on e4b16428f58f7504dec658dbe53d40227ace67a3.)

dnadlinger:
@pathfinder49 One to add to the list…

dnadlinger:
Reopening, as fix was reverted in https://github.com/m-labs/artiq/commit/ae999db8f6814c63eae563aca69276feb59d305d (see commit message for details).

PR #2239, opened 2023-10-08
compiler: Catch escaping numpy.{array, full, transpose}() results

dnadlinger:
Function calls in general can still be used to hide escaping
allocations from the compiler (issue #1497), but these calls in
particular always allocate, so we can easily and accurately handle
them.


dnadlinger:
@sbourdeauducq We might want to backport this to older releases, at least after initial testing has confirmed my assesment that this is in fact precise (i.e. should not reject corruption-free code).

PR #2249, opened 2023-10-11
jsonschema: shuttler: fix wrong default drtio_dest

linuswck:
# ARTIQ Pull Request

## Description of Changes
Shuttler on Kasli-Soc should have a default drtio_destination of 5 instead of 4. This PR adds a if-then-else to set default value conditionally according to the target.
Note: if-the-else cannot be used to define default value directly.

Tests Done
1. define shuttler with no "drtio_destination" given on "kasli" target
2. define shuttler with no "drtio_destination" given on "kasli_soc" target
3. define shuttler with valid "drtio_destination" given (set to some number)
4. define shuttler with invalid "drtio_destination" given (set to some string of text)

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

### Licensing


Spaqin:
To make it even more complicated, Kasli 1/1.1 should have DRTIO destination at 3...

occheung:
I original put the default there just to make it compatible to earlier JSONs that I made. I decided to not think about the validity of the destination on the Shuttler entry level too much, as there weren't any mechanism that prevents assigning the same destination to different satellites anyway.

PR #2275, opened 2023-11-03
kasli: Correct the GTP TX Clock Path during Initialization

linuswck:
# ARTIQ Pull Request
This problem is discovered when we observed a non-deterministic phase of system clock at different reset and power up of the system. This PR apparently fixes this issue.

## Description of Problem
Below shows a brief drawing of the clocking structure of Kasli Master and EFC Shuttler Board are connected together.
Kasli Master and EFC Kasli Satellite Clocking Structure
Note1: There are 2 more GTP Lanes in Slave Mode not shown in this drawing.
Note2: The image is uploaded here for future reference if the clocking structure is needed to be changed.

Datasheet Rules Violated: During GTP TX Initialization, TXOUT must be fed back into TXUSRCLK during initialization.
If the clock path is changed for GTP TX after Initialization, GTP TX must be reset and goes through initialization again.
image

However, in current gateware implementation, the MMCM clock input is switched only after GTP TX Init is done. This means during GTP TX Initialization, the TXUSRCLK2 and TXUSRCLK are being driven by the external 125MHz oscillator clock source instead of TXOUTCLK. Thus, the GTP Delay Alignment and Phase Alignment will result in incorrect result. Then, System Clock is being shifted by the GTP TX Delay Aligner incorrectly and randomly.

## Description of Changes
This PR corrects the startup sequence and clock path configuration.
Startup Sequence
1. Firmware is booted up by the external 125MHz oscillator
2. Firmware configures the Si5324 Oscillator and wait for it to lock
3. (Change) Firmware instructs the clock switch FSM to switch MMCM clock input via `stable_clkin` csr register 
4. (Change) GTP TX Init FSM starts after clock is switched and MMCM is locked. During this time, reset in sys clk domain is asserted.
5. (Change) GTP TX Initialization is completed. Reset in sys clk domain is deasserted.

### Test
This PR is tested on Kasli v2.0 board and based on the latest commit(378dd0e5caf3a2afcb5ffe85f91fcdb68b9359f2). Phase difference is measured between System Clock being connected to an EEM Port's Pin through an ODDR Primitive and EEM CLOCK OUT MMCX connector on Kasli PCB.

Kasli Firmware and Gateware start up as usual. There is no change in phase difference observed after TX Init is completed in different startup conditions like cold start (when the temperature of FPGA is cold), multiple power cycles, multiple resets without power cycle.

### Related PR
Misoc: [PR](https://github.com/m-labs/misoc/pull/144)

### Actions Required after Merging
1. Flake Dependency updates for Misoc
2. Port this change to Kasli-soc platform, which should have similar problem.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

### Licensing


Spaqin:
Is KC705 also affected by this? I know it will not support Shuttler and it uses GTX rather than GTP, but it could be worthwhile cleaning that up as well while we're at it.

linuswck:
> Is KC705 also affected by this? I know it will not support Shuttler and it uses GTX rather than GTP, but it could be worthwhile cleaning that up as well while we're at it.

Yes. It is affected. Porting the changes to these platforms that uses GTX.

PR #2277, opened 2023-11-04
Minor documentation changes

JILAelec:
# ARTIQ Pull Request

## Description of Changes

I worked through the documentation and corrected things as I went along.

## Type of Changes

|   | Type |
| ✓  | :scroll: Docs |


doc/manual/getting_started_core.rst
```diff
@@ -151,16 +151,29 @@ It is often necessary that several pulses overlap one another. This can be expre
...
+    class Tutorial(EnvExperiment):
+        def build(self):
+            self.setattr_device("core")
+            self.setattr_device("ttl0")
+            self.setattr_device("ttl1")
+
+        @kernel
+        def run(self):
+            self.core.reset()
+            self.ttl0.output()
```
Spaqin, line 166: This ``output()`` call seems to be specific to your setup? As it's not present in the example above, and only ``ttl0`` is affected; and on Sinara TTL cards it will not do anything either.

JILAelec: I think it's just there to ensure that ttl0 is in output mode. For consistency, there should be the same statement for ttl1 or we should drop both of them. I don't have a particularly strong opinion either way.

architeuthidae:
Changes incorporated into #2431 and #2451 larger manual edits. Thanks

PR #2435, opened 2024-06-12
doc: Add "Using Drtio" and overhaul DRTIO page

architeuthidae:
# ARTIQ Pull Request

## Description of Changes

Created new page "Using DRTIO and subkernels", closer to the tutorial style of the two "Getting started" pages, to contain DRTIO setup instructions and subkernel tutorial (pulled out of "Getting started with the core language" page). Added some material to subkernel documentation, in particular some warnings about potential pitfalls for the programmer.  DRTIO page now mostly consists of internal description, which is already very detailed and largely left as is, except that some of the introductory paragraphs are given to "Using DRTIO" as an overview. 

## Type of Changes

|   | Type |
| ✓  | :scroll: Docs |

### All Pull Requests

### Documentation Changes

### Git Logistics

  ```
  
  ```

### Licensing


doc/manual/using_drtio_subkernels.rst
```diff
@@ -0,0 +1,206 @@
...
+
+Subkernels
+----------
+
+Rather than only offloading the RTIO channels to satellites and limiting all processing to the master core device, it is fully possible to run kernels directly on satellite devices. These are referred to as *subkernels*. Using subkernels to process and control remote RTIO channels can free up resources on the core device.  
+
+Subkernels behave for the most part like regular kernels; they accept arguments, can return values, and are marked by the decorator ``@subkernel(destination=i)``, where ``i`` is the satellite's destination number as used in the routing table. To call a subkernel, call it like any other function. There are however a few caveats: 
+
+   - subkernels do not support RPCs,
+   - subkernels do not support (recursive) DRTIO (but they can call other subkernels and send messages to each other, see below),
```
architeuthidae, line 95: I gather (though I didn't check) that DRTIO is technically possible specifically within DMA operations -- since it's fed back through the master, and it's explicitly mentioned in one of the implementation PRs that this will be distributed among satellites as necessary -- but I deliberately didn't mention this as it seems like something of more of a coincidence than a feature, and probably unnecessarily confusing in this context.

architeuthidae: "Does not support DRTIO" is taken directly from the original phrasing. #2235 describes it the same way. The addition of "recursive" is meant to clarify that they are in themselves DRTIO, but don't support _further_ DRTIO, so to speak. I understood it to mean subkernels are not allowed to access RTIO channels of other satellites / the master... except for the apparent loophole via DDMA, unless that's blocked off by some intermediate mechanism ....? @Spaqin?

Spaqin: DDMA is done differently - rather than just directly playing back all RTIO events, they get sorted and sent out to appropriate satellites to be played back later. Master is one exception to this, there's no support in firmware for playing back DDMA. 

Which means DDMA is using the aux protocol with routing that allows the aux packets to go anywhere, rather than the main DRTIO protocol. So yes, there would be a "loophole" in that sense.

I am quite sure that it wouldn't be possible to dispatch events upstream, but downstream (to further satellites) may still work, that should be verified.

architeuthidae: Right, so the question to me remains whether to describe this DDMA quote-unquote "loophole" in the docs or not. Or I can also just rephrase if we don't like calling this "recursive DRTIO", which admittedly is not very clear

Spaqin: > Why not?

DMA control is accessible in the [kernel CPU only](https://github.com/m-labs/artiq/blob/master/artiq/gateware/targets/kasli.py#L319) on standalone/master firmware. Asking to do it from comm core (where all DDMA data is sent and handled) in current architecture (where kernel CPU sends requests and comm only replies) wouldn't be possible, and frankly it's not a huge disadvantage - if you need DDMA involving master, just run it from the master.

Spaqin: > But then it's not distributed since all events have to come from the master's SDRAM and RTIO CRI bus.

No, it can be distributed still (distribution must be [enabled first](https://github.com/m-labs/artiq/blob/master/artiq/coredevice/dma.py#L78)).  Just the local part runs from kernel CPU.

 > I suppose it's the same on satellites, subkernels cannot start DMA?
 
 They can, but it's a workaround - DMA start syscall is different for satellites, and it [sends a request](https://github.com/m-labs/artiq/blob/master/artiq/firmware/ksupport/lib.rs#L451) to the comm CPU to start DMA. On master, similar approach could be used, but since kernel<->comm communication is slower, I don't think that's a good approach, since it would slow down every DMA call, regardless if DDMA is used or not.

doc/manual/drtio.rst
```diff
@@ -20,90 +14,41 @@ The lower layers of DRTIO are similar to White Rabbit, with the following main d
...
-Using DRTIO
------------
+Time transfer and clock syntonization is typically done over the serial link alone. The DRTIO code is written as much as possible to support porting to different types of transceivers (Xilinx MGTs, Altera MGTs, soft transceivers running off regular FPGA IOs, etc.) and different synchronization mechanisms.
 
 Terminology
-+++++++++++
+-----------
 
-In a system of interconnected DRTIO devices, each RTIO core (driving RTIO PHYs; for example a RTIO core would connect to a large bank of TTL signals) is assigned a number and is called a *destination*. One DRTIO device normally contains one RTIO core.
+In a system of interconnected DRTIO devices, each RTIO core is assigned a number and is called a *destination*. One DRTIO device normally contains one RTIO core.
```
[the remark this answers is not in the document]
architeuthidae: the term "PHY" isn't explained anywhere else in the manual and never otherwise referenced except for two of the argparse generated descriptions in core_reference -- IMO the parenthetical isn't really adding anything in this specific section, which is supposed to be about the DRTIO-specific terminology, and only confuses the sentence. if we want to concretely establish the correspondence RTIO core = PHYs somewhere for people who already know what a PHYs is and/or the general definition of RTIO core for those who don't, then this page doesn't seem like the place for it (could add a sentence to that effect in "RTIO Concepts", maybe.)

doc/manual/drtio.rst
```diff
@@ -1,17 +1,11 @@
-Distributed Real Time Input/Output (DRTIO)
-==========================================
+DRTIO system 
+============
 
-DRTIO is a time and data transfer system that allows ARTIQ RTIO channels to be distributed among several satellite devices synchronized and controlled by a central core device.
+DRTIO is the time and data transfer system that allows ARTIQ RTIO channels to be distributed among several satellite devices, synchronized and controlled by a central core device. The main source of DRTIO traffic is the remote control of RTIO output and input channels. The protocol is optimized to maximize throughput and minimize latency, and handles flow control and error conditions (underflows, overflows, etc.)
 
-The link is a high speed duplex serial line operating at 1Gbps or more, over copper or optical fiber.
```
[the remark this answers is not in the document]
architeuthidae: it still exists, it's now in the introduction of "using drtio" (line 8). i missed the opportunity to update it for DRTIO-over-EEM, though, i'll do that

doc/manual/using_drtio_subkernels.rst
```diff
@@ -0,0 +1,209 @@
...
+
+
+.. note:: 
+    As with other configuration changes (e.g. adding new hardware), if you are in possession of a non-distributed ARTIQ system and you'd like to expand it into a DRTIO setup, it's easily possible to do so, but you need to be sure that both master and satellite are (re)flashed with this in mind. As usual, if you obtained your hardware from M-Labs, you will normally be supplied with all the binaries you need, through ``awfs_client`` or otherwise.  
+
+.. note:: 
+    Do not confuse the DRTIO *master device* (used to mean  the central controlling core device of a distributed system) with the *ARTIQ master* (the central piece of software of ARTIQ's management system, which interacts with ``artiq_client`` and the dashboard.) ``artiq_run`` can be used to run experiments on DRTIO systems just as easily as non-distributed ones, and the ARTIQ master interacts with the central core device regardless of whether it's configured as a DRTIO master or standalone.
```
Spaqin, line 17: ``used to mean  the central`` has double space

doc/manual/using_drtio_subkernels.rst
```diff
@@ -0,0 +1,209 @@
...
+
+Subkernels behave for the most part like regular kernels; they accept arguments, can return values, and are marked by the decorator ``@subkernel(destination=i)``, where ``i`` is the satellite's destination number as used in the routing table. To call a subkernel, call it like any other function. There are however a few caveats: 
+
+   - subkernels do not support RPCs,
+   - subkernels do not support (recursive) DRTIO (but they can call other subkernels and send messages to each other, see below),
+   - they support DMA, for which DDMA is considered always enabled,  
+   - their return values must be fully annotated with an ARTIQ type,
+   - their arguments should be annotated, and only basic ARTIQ types are supported,
+   - they can raise exceptions, but the exceptions cannot be caught by the master (they can only be caught locally or propagated directly to the host), 
+   - while ``self`` is allowed as an argument, it is retrieved at compile time and a purely local object afterwards -- any changes made by other kernels will not be shown, and changes made locally will not be visible anywhere else.
```
Spaqin, line 103: I think changes to the ``self`` object wouldn't be "shown", but rather applied or visible.

doc/manual/using_drtio_subkernels.rst
```diff
@@ -0,0 +1,209 @@
...
+
+Subkernels in practice
+^^^^^^^^^^^^^^^^^^^^^^
+
+Subkernels begin execution as soon as possible when called. By default, they are not awaited, but awaiting is necessary to receive results or exceptions. The await function ``subkernel_await(function, [timeout])`` takes as argument the subkernel to be awaited and, optionally, a timeout value in milliseconds. If the timeout is reached without response from the subkernel, a :exc:`~artiq.coredevice.exceptions.SubkernelError` is raised. If no timeout value is supplied the function waits indefinitely. Negative timeout values are ignored.
```
Spaqin, line 108: Negative timeout values aren't ignored per se but rather indicate awaiting indefinitely, just like 0.

doc/manual/using_drtio_subkernels.rst
```diff
@@ -0,0 +1,209 @@
...
+
+        @kernel
+        def run(self):
+            subkernel_add(2, 2)
+            result = subkernel_await(subkernel_add)
+            assert result == 4
+
+Subkernels are compiled after the main kernel and immediately sent to the designated satellite. When they are called, the master simply instructs the subkernel to load and run the corresponding kernel. When ``self`` is used in subkernels, it is embedded into the compiled and uploaded data; this is the reason why changes made do not propagate between kernels.
+
+If a subkernel is called on a satellite where a kernel is already running, the newer kernel overrides silently, and the previous kernel will not be completed.
```
Spaqin, line 131: I would also note that a subkernel and DDMA (from another device, not counting DMA within subkernel itself) will not run simultaneously and trying to do so will cause an error.

doc/manual/using_drtio_subkernels.rst
```diff
@@ -0,0 +1,209 @@
...
+        @kernel
+        def run(self):
+            simple_self()
+            subkernel_send(1, "message", 150)
+            result = subkernel_await(simple_self)
+            assert result == 170
+
+The ``subkernel_send(destination, name, value)`` function requires three arguments: a destination, a name for the message (to be used for identification in the corresponding ``subkernel_recv()``), and the passed value.
+
+The ``subkernel_recv(name, type, [timeout])`` function requires two arguments: message name (matching exactly the name provided in ``subkernel_send``) and expected type. Optionally, it also accepts a third argument, a timeout for the operation in milliseconds. As with ``subkernel_await``, the default behavior is to wait as long as necessary, and a negative argument is ignored.
```
Spaqin, line 207: same remark with negative numbers

doc/manual/using_drtio_subkernels.rst
architeuthidae: = equivalent to giving no argument at all = ignored, no? I still kind of think it's the least convoluted way to convey the necessary information (the salient part really just being "Don't try to use negative timeouts, it won't do anything useful")

doc/manual/drtio.rst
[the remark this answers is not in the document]
architeuthidae: Makes sense, will do

doc/manual/using_drtio_subkernels.rst
architeuthidae: Added this, but at a slightly different location, at line 137 with the "satellite and downstream devices are inaccessible to master while subkernel running" paragraph

architeuthidae:
All discussed corrections should be made

Issue #1967, opened 2022-09-16
RPC string corruption


b-bondurant:
# Bug Report

## One-Line Summary

String RPC return values (and/or kernel return values?) seem to be getting corrupted, which can indirectly cause a kernel panic.

### Steps to Reproduce

Run the following experiment:
```python
from artiq.experiment import *


class BadString(EnvExperiment):
    def build(self):
        self.setattr_device('core')

    @rpc
    def get_string_rpc(self) -> TStr:
        return "a string"

    @kernel
    def get_string(self) -> TStr:
        self.core.reset()
        my_str = self.get_string_rpc()
        # seems to cause the kernel panic
        # print(my_str)
        return my_str

    def run(self):
        print(self.get_string())
```

### Expected Behavior

Prints `a string` to the console.

### Actual (undesired) Behavior

Prints `ring` to the console. And if you uncomment the `print` statement in the kernel it causes a kernel panic with the following message:
```
[ 64082.239740s]  INFO(kernel): panic at /nix/store/2w4h6z0k128aa52r1law0zysi68arp34-python3.9-artiq-7.8123.3038639/lib/python3.9/site-packages/artiq/firmware/libproto_artiq/rpc_proto.rs:209:69: called `Result::unwrap()` on an `Err` value: Utf8Error { valid_up_to: 0, error_len: Some(1) }
[ 64082.264133s] ERROR(runtime::session): session aborted: unexpected request RunAborted from kernel CPU
```
And on the host side:
```
a string
Traceback (most recent call last):
  File "/nix/store/2w4h6z0k128aa52r1law0zysi68arp34-python3.9-artiq-7.8123.3038639/bin/.artiq_run-wrapped", line 9, in <module>
    sys.exit(main())
  File "/nix/store/7qs5shpgswzik9vsr7rkvz2l7znzmjz0-python3-3.9.13-env/lib/python3.9/site-packages/artiq/frontend/artiq_run.py", line 224, in main
    return run(with_file=True)
  File "/nix/store/7qs5shpgswzik9vsr7rkvz2l7znzmjz0-python3-3.9.13-env/lib/python3.9/site-packages/artiq/frontend/artiq_run.py", line 210, in run
    raise exn
  File "/nix/store/7qs5shpgswzik9vsr7rkvz2l7znzmjz0-python3-3.9.13-env/lib/python3.9/site-packages/artiq/frontend/artiq_run.py", line 203, in run
    exp_inst.run()
[...]

PR #2470, opened 2024-07-03
doc: Rewrite and overhaul of 'Compiler' page

### LINKED ISSUE #656: Default types overridden
Do parts of the ARTIQ codebase override or clear default Python types? For the most part I have been able to get around these issues, but now it has become a hassle. For instance, I am unable to use default Python arrays, dictionaries, and even some casting methods such as `str`. For example, when casting an integer into a string: `fatal: name 'str' is not bound to anything`

Issues like this occur when even creating dictionaries: `fatal: name 'dict' is not bound to anything; did you mean 'int'?` Finally, it seems that even Python arrays don't usually work in my environment: when setting a variable `my_array = []`, I am told that this object does not have functions such as `append`, `pop`, or `push`. I was just wondering if these types are overridden by one of the classes or packages in `artiq.experiment`. When testing in the Python interpreter, these methods still work even after importing `artiq.experiment`, so does is this a side effect of using the `@kernel` decoration?

### LINKED ISSUE #717: manual: document object mutability and lifetimes
There are a few other differences between ARTIQ-Python and regular Python that should be documented in the manual:

* Data passing/mutability: explain the different allowed and forbidden ways of passing data from kernels back up to the host (attribute writeback or rpc, but not: modify list argument, return list).
* Lifetime: explain that (maybe not even why) you can't return lists allocated inside kernels. Not to kernels and not to host-Python.

These may also be FAQs.

### LINKED ISSUE #1162: Compiler doesn't parse @PREFIX.kernel (kernel decorator with path)
# Description
ARTIQ compiler doesn't recognize `@PREFIX.kernel` decorator.

## Replicating
Instead of using `from artiq.experiments import *`, use `import artiq.language.core`. Then decorate function with `@artiq.language.core.kernel`.

### Sample Code

I would like to use following, because per [PEP20](https://www.python.org/dev/peps/pep-0020/), **Explicit is better than implicit**. Also, because it avoids cluttering the namespace and speeds imports.

Sample based on LED tutorial:
```python
import artiq.language.environment as artiq_env
import artiq.language.core as artiq_lang_core

class LED(artiq_env.EnvExperiment):
    def build(self):
        self.setattr_device("core")
        self.setattr_device("led")

    @artiq_lang_core.kernel
    def run(self):
        self.core.reset()
        self.led.on()

```

### Expected Behavior
ARTIQ compiler should compile the function for the core device as it would any other function with `@kernel`

### Actual Behavior
ARTIQ throws CompileError (when run in separate `pytest` config), or standard error with above sample & `artiq_run`. Seems to be issue with https://github.com/m-labs/artiq/blob/0fef2afccbf7460dacac839caaab86ff376f3e03/artiq/compiler/transforms/inferencer.py#L1292-L1302

Tested also with `import artiq.language.core;  @artiq.language.core.kernel`, and produced same result.

## Log Message/Traceback

With artiq_run: `error: decorators are not supported @artiq_lang_core.kernel`.

### Full traceback
```
$ artiq_run ./led_kernel_explicit.py --vv
DEBUG:artiq.compiler.import_cache: hook installed
DEBUG:artiq.compiler.import_cache: added 'PATH/artiq/coredevice/ttl.py' to cache
./led_kernel_explicit.py:9:5-9:6: error: decorators are not supported
    @artiq_lang_core.kernel
    ^
```

### LINKED ISSUE #1298: Returning array from kernel
# Bug Report

## One-Line Summary

Trying to return lists or numpy arrays created on the coredevice as return value of kernel functions gives error.

### Steps to Reproduce
This throws an error:
```
    def run(self):
        a = self.krun()

    @kernel
    def krun(self):
        lst = [1.0]*100
        return lst
```

However, this works fine:
```
    def run(self):
        self.krun()
        print(self.arr)

    @kernel
    def krun(self):
        lst = [1.0]*100
        self.update_arr(lst)

    def update_arr(self, arr):
        self.arr = arr
```

```
root:While compiling /home/ion/testing/bugs_mwe/array_from_kernel.py
/home/ion/testing/bugs_mwe/array_from_kernel.py:17:16-17:19: error: cannot return an allocated value that does not live forever
        return lst
               ^^^
/home/ion/testing/bugs_mwe/array_from_kernel.py:13:5: note: the value of the expression is alive from this point...
    @kernel
    ^      
/home/ion/testing/bugs_mwe/array_from_kernel.py:17:19: note: ... to this point
        return lst
                  ^
/home/ion/testing/bugs_mwe/array_from_kernel.py:17:16-17:19: note: this expression has type list(elt=float)
        return lst
               ^^^
```

### Your System
Linux machine
ARTIQ v5.0.dev+452.gf8846a33

I guess this might be related to #224, except here it is a list that should be returned. Similar thing happens when lst is created via np.array or np.full.
I am not sure whether this is a bug, a limitation or a feature, but since the documentation is lacking with regards to this, I thought I should open a issue here.

### LINKED ISSUE #1462: Documentation: Kernel <-> Host Attribute Synchronisation/writeback
# Question

## Description

The kernel-host attribute synchronisation behaviour is not apparent to new users from reading the documentation.

The code below is intended to demonstrate the current behaviour:
```Python3

class AttributeTest(EnvExperiment):
    def build(self):
        self.setattr_device("core")

    def prepare(self):
        self.a = "A"

    @kernel
    def run(self):
        print("krun")
        print(self.a)
        self.rpc() 
        print("krun after rpc")
        print(self.a)
        self.a = "A_2"
        print(self.a)
        self.rpc()

    def rpc(self):
        print("rpc")
        print(self.a)
        self.a = "A_1"
        print("rpc modified")
        print(self.a)
```
This prints:
```
krun
A
rpc
A
rpc modified
A_1
krun after rpc
A
A_2
rpc
A_1
rpc modified
A_1
```

I've been aware of this behaviour for some time. However, I can't quite decide if it is intended behaviour or a bug. It also raises the question of why kernel_invariants need to be explicitly specified (given that attribute changes outside the compiler are not synchronised to the core device).

architeuthidae:
# ARTIQ Pull Request

## Description of Changes

Rewrite and overhaul of 'Compiler' manual page. Details on types and decorators. Attempts to clarify longstanding confusion points re: lists, arrays, etc.

### Related Issue
Closes #1462 
Closes #1162 
Closes #717 
Closes #656
Closes #1298 -- at least in that the behavior is now documented, and (pending NAC3?) still intended.

Does not close but does document #1967. 

## Type of Changes

|   | Type |
| ✓  | :scroll: Docs |

### All Pull Requests

- [ ] Close/update issues.

### Documentation Changes

### Git Logistics

  ```
  
  ```

### Licensing


doc/manual/compiler.rst
```diff
@@ -54,74 +72,80 @@ The Python types correspond to ARTIQ type annotations as follows:
...
+Kernels can freely modify attributes of objects shared with the host. However, by necessity, these modifications are actually applied to local copies of the objects, as the latency of immediate writeback would be unsupportable in a real-time environment. Instead, modifications are written back *when the kernel completes;* notably, this means RPCs called by a kernel itself will only have access to the unmodified host version of the object, as the kernel hasn't finished execution yet. In some cases, accessing data on the host is better handled by calling RPCs specifically to make the desired modifications. 
+
+.. warning::
 
-This results in memory corruption at runtime.
+    Kernels *cannot and should not* return lists, arrays, or strings they have created, or any objects containing them; in the absence of a heap, the way these values are allocated means they cannot outlive the kernels they are created in. Trying to do so will normally be discovered by lifetime tracking and result in compilation errors, but in certain cases lifetime tracking will fail to detect a problem and experiments will encounter memory corruption at runtime. For example: :: 
+
+        def func(a):
```
Spaqin, line 102: I know it was in the original, but should we still have negative code examples in the manual? Could be confusing if skimming through the manual, and if it's in a Github issue, linking to it would probably be the better option.

Does remind me that Rust kind of works around negative examples with a special icon indicating that the code is broken/will fail to compile.

Spaqin: Nevermind, disregard this - it's in a warning section, very visible as a potential problem.

doc/manual/compiler.rst
```diff
@@ -54,74 +76,125 @@ The Python types correspond to ARTIQ type annotations as follows:
...
+
+    will compile, **but corrupts at runtime.** On the other hand, lists, arrays, or strings can and should be used as inputs for RPCs, and this is the preferred method of returning data to the host. In this way the data is inherently read and sent before the kernel completes and there are no allocation issues. 
+
+Available built-in functions 
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 
+
+ARTIQ makes various useful built-in and mathematical functions from Python, NumPy, and (if present) SciPy available in kernel code. They are not guaranteed to be perfectly equivalent to their host equivalents (for example, ``numpy.rint`` normally rounds-to-even, but in kernel code rounds toward zero) but their behavior should basically predictable. 
+
+
+.. list-table::
```
architeuthidae, line 128: I don't really like this flat list, as it's disconnected from the code and potentially painful to maintain (not to mention format) but I do think it should be possible to find out which functions are available somewhere in the documentation, as opposed to "read the source code" or "just try it and see what breaks." That said, very open to suggestions on a better way to present this information.  (And is it correct to link to the NumPy/SciPy documentation? As someone not particularly used to those libraries, it wouldn't be intuitively clear to me what "j0" or "expm1" means without some kind of reference...)

doc/manual/compiler.rst
```diff
@@ -54,74 +76,125 @@ The Python types correspond to ARTIQ type annotations as follows:
...
+        *   - ``copysign``, ``nextafter()``
+    +   * `SciPy special functions <https://docs.scipy.org/doc/scipy/reference/special.html>`_ 
+        *   - ``erf()``, ``erfc()``
+            - ``gamma()``, ``gammaln()``
+            - ``j0()``, ``j1()``, ``y0()``, ``y1()``
+
+Basic NumPy array handling (``np.array()``, ``numpy.transpose()``, ``@``, element-wise operation etc.) is also available. NumPy functions are implicitly broadcast when applied to arrays. 
+
+.. warning:: 
+    A kernel ``print`` call normally specifically prints to the host machine, either the terminal of ``artiq_run`` or the dashboard log when using ``artiq_dashboard``. This makes it effectively an RPC, with some of the corresponding limitations. In subkernels and whenever compiled through ``artiq_compile``, where RPCs are not supported, it is instead considered a print to the local UART log.
```
Spaqin, line 160: >  whenever compiled through ``artiq_compile``

Compiled as an idle or startup kernel with ``artiq_compile``?

> local UART log.

UART only on satellites, master/standalone log is also accessible by network (coremgmt), just won't be displayed on the terminal (as there's no terminal to connect to)

doc/manual/compiler.rst
Spaqin: Disclaimer: as the software is in beta this list may be non-exhaustive or there may be discrepancies?

doc/manual/compiler.rst
architeuthidae: Oops, I corrected this and then somehow must have un-corrected it while fistfighting rST table formatting

doc/manual/compiler.rst
architeuthidae: As far as I can tell from commit histories etc. this has been pretty static since introduced four years ago, I'm not sure if that's true as such. But a disclaimer probably couldn't hurt

architeuthidae: Race condition: commented before I saw yours. Makes sense

doc/manual/compiler.rst
architeuthidae: Also, `artiq_compile` doesn't allow RPCs as a matter of principle, and it isn't technically limited to idle or startup kernels, nor does it know that it's working on an idle or startup kernel. So I'd stick with the broader statement, it seems more accurate

architeuthidae:
Force-push to edit upon the revelation that ARTIQ installs with SciPy in all normal cases anyway...

SimonRenblad (approved):

fsagbuya (approved):

PR #2521, opened 2024-07-30
Safe copy with attributes of private key

thomasfire:
# ARTIQ Pull Request

## Description of Changes

Safe copy with attributes of private key.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -470,8 +470,10 @@
             ''
             export HOME=`mktemp -d`
             mkdir $HOME/.ssh
-            cp --preserve=mode /opt/hydra_id_ed25519 $HOME/.ssh/id_ed25519
-            cp --preserve=mode /opt/hydra_id_ed25519.pub $HOME/.ssh/id_ed25519.pub
+            touch $HOME/.ssh/id_ed25519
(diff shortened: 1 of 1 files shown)


thomasfire:
while this solution looks hacky, the alternatives either will not work or will be even more hacky.
* both kind of links will not work, because ssh will dereference it anyway and check permissions
* the original key attrs cannot be lower than 640, because nix builder is not tied to the user and typically uses nixbld1-nixbld32 users for building
* install, cp and other coreutils are copying file first, and then setting the permission
* tar sets permissions before extraction, but piping the tar in and out is more hacky than just using touch and cat

Other solutions:
* fork openssh and remove that permissions check
* fork cp or install and set permissions before writing to the file
* implement our own (or find one) custom copy, that would set permissions before writing
* ~~fork nix and make it use only one specified user~~

thomasfire:
75ffbeba4d59dce5a594060751726b467b40663a needs revert

thomasfire:
> Doesn't hurt and the code is slightly shorter.

it hurts, because ssh cannot work with 644 permissions on the key

PR #2532, opened 2024-08-09
Special testing process for LVDS TTL

### LINKED ISSUE #2287: Proper LVDS TTL testing process needed
# ARTIQ Feature Request

For now artiq_sinara_tester doesn't have dedicated test process for the LVDS TTL card, which results in need of continuous reflashing the gateware in order to test all ports, and also needs to be run 4 times each to test different subset of ttls.

The least can be done is the ability to choose 4 channels for testing the inputs. Also outputs better be checked simultaneously with inputs, as checks by oscilloscope are hard to conduct properly.

thomasfire:
# ARTIQ Pull Request

## Description of Changes

Implements special testing process for LVDS TTL - needs it being set so that it has at least one group of inputs and outputs each. 

### Related Issue

Closes #2287 

## Type of Changes

|   | Type |

### All Pull Requests

### Code Changes

- [x] Tested with assembly team without giving additional instructions.

### DIFF
--- a/artiq/frontend/artiq_sinara_tester.py
+++ b/artiq/frontend/artiq_sinara_tester.py
@@ -75,11 +75,13 @@ class SinaraTester(EnvExperiment):
                     dev = self.get_device(name)
                     if "led" in name:  # guess
                         self.leds[name] = dev
+                    elif "board" in desc and desc["board"] == "dio_lvds":
+                        self.ttl_lvds_outs[name] = self.get_device(name)
                     else:
                         self.ttl_outs[name] = dev
                 elif (module, cls) == ("artiq.coredevice.ttl", "TTLInOut"):
                     if "board" in desc and desc["board"] == "dio_lvds":
-                        self.ttl_lvds[name] = self.get_device(name)
+                        self.ttl_lvds_ins[name] = self.get_device(name)
                     else:
                         self.ttl_ins[name] = self.get_device(name)
                 elif (module, cls) == ("artiq.coredevice.urukul", "CPLD"):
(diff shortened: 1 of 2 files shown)


artiq/frontend/artiq_sinara_tester.py
```diff
@@ -244,6 +253,63 @@ def test_ttl_ins(self):
...
+        if len(self.ttl_lvds_ins) == 0:
+            print(f"No available input channels for testing. Skipping...")
+            return
+
+        ttl_in_index, ttl_out_index = 0, -1
+        default_ttl_in_name, _ = self.ttl_lvds_ins[ttl_in_index]
+        default_ttl_out_name, _ = self.ttl_lvds[ttl_out_index]
+
+        print("LVDS TTL channels are tested in groups of 4.")
+        print("You can choose any channel within the group.")
```
Spaqin, line 287: Would it be better to specify the group (channel % 4), or have it "align" to a group?

thomasfire: It already does it (?) - if you specify any channel in group `ttl0...ttl3`, it will select that group.

Spaqin: As in, instead of asking for a channel, it would ask for a group (instead of ttl3, group 0). Especially here, where a unit of test is essentially a group anyway - you connect the RJ45 cable to one. Then you wouldn't have to calculate the lower boundary either.

It is a bit of a nitpick and functionally wouldn't change anything, but I would consider it being a bit confusing at a first glance.

thomasfire: I thought about it, but decided that it would confuse the tester more.

PR #2542, opened 2024-08-16
doc: Add 'Overview' diagram to manual

architeuthidae:
# ARTIQ Pull Request

## Description of Changes

As of this PR, looks like this: 

image

which is a midpoint between two earlier colour schemes considered respectively too muddy and too eye-searing: 

image

image

IMO 'network' and 'rtio' are the most important fields and need the brightest colours, but this is definitely also a bit heavy on the magenta. Feedback appreciated if anyone has any. Otherwise, obviously, please correct any factual errors. 

Note also that the matrix colours have meanings, though they are uncaptioned: brand colour ARTIQ dark blue / black is used for PCs, ARTIQ teal is used for peripherals, and the core device is highlighted in the same dark red as the master to emphasize they're the two really important poles in this diagram. 

Generates similarly, and equally readably, albeit with somewhat less aesthetically appealing spacing, in PDF form. 

## Type of Changes

|   | Type |
| ✓  | :scroll: Docs |

### All Pull Requests

### Documentation Changes

- [x] Check, test, and update the documentation in [doc/](../doc/). Build documentation (`nix build .#artiq-manual-html; nix build .#artiq-manual-pdf`) to ensure no errors.

### Licensing


Spaqin:
I'd replace ``Cloned repository`` in other machines with just ``repository`` to keep it consistent. Or ``Experiment repository`` if you want to be more specific.

``Other utilities`` feels kind of tacked on, unless defined in text; otherwise it feels like just extra clutter.

architeuthidae:
> I'd replace `Cloned repository` in other machines with just `repository` to keep it consistent. Or `Experiment repository` if you want to be more specific.

The point I'm trying to emphasize (for the non-Git-literate) is that the client copy of the repository is just that, a copy; that it can exchange data with the master, but the master possesses the, well, master copy, *even if* it's a bare repository. It seems to be something that can cause confusion. So I'd argue for leaving this one in. 

> `Other utilities` feels kind of tacked on, unless defined in text; otherwise it feels like just extra clutter.

To avoid having to edit the screenshot I'll wait for everyone's feedback before changing anything, but noted. Might remove it from the client machine and replace it with specifically namedropping `artiq_run` in the host.

SimonRenblad:
As for the "other utilities", I assume this also includes the analyzer / moninj proxies. It should maybe be noted that the dashboard may communicate with the proxies (it is the main use case after all). I am not certain that these connections are important enough to note here, but something to keep in mind.

On another note, I would argue that some of the repeated content in the other machine vs host machine could probably be cut. I don't see why `artiq_dashboard` would need to be repeated twice, for example. Maybe instead of two separate segments for host vs other, box out the components that need to run on the host machine vs which can run on host / other machine.

architeuthidae:
> As for the "other utilities", I assume this also includes the analyzer / moninj proxies. It should maybe be noted that the dashboard may communicate with the proxies (it is the main use case after all). I am not certain that these connections are important enough to note here, but something to keep in mind.

They are! These are the built-in controllers mentioned in host machine -> controllers. They're connected to the client box at large rather than the dashboard specifically, but that's easily changed.

architeuthidae:
> On another note, I would argue that some of the repeated content in the other machine vs host machine could probably be cut. I don't see why `artiq_dashboard` would need to be repeated twice, for example. Maybe instead of two separate segments for host vs other, box out the components that _need_ to run on the host machine vs which can run on host _or_ other machine.

Cautiously I'm going to argue to keep the duplication. There are a few details here which are inevitably simplified, which I've tried to cover with 'most of these components can be run independently' -- this is not trying to be a full functional mesh of how everything _can_ connect, just a simplified layout of _a_ believable/possible ARTIQ setup, so people can wrap their heads around it. IMO, it's pretty important to emphasize that clients/controller managers _can_ be run on the master's machine or remotely, _and_ that you can do both at once -- the duplication hopefully makes 'both at once' especially graphically obvious.

There's also other simplifications, like assuming `artiq_flash` is run from the same machine as the master. But that one seems necessary to me as a terminology problem: IIRC the documentation also refers to the 'host machine' when flashing / using artiq_run, and I think this diagram is the wrong place to try and specify there's no actual obligation for this to be the same PC.

architeuthidae:
Also probably worth mentioning that I have aspirations of a second, probably similarly convoluted diagram, which will live in the management system section and focus only on the specifics of how those various pieces interact, as opposed to here getting grouped vaguely as 'well it's doing ... something ... over network...'

SimonRenblad:
> it's pretty important to emphasize that clients/controller managers can be run on the master's machine or remotely, and that you can do both at once -- the duplication hopefully makes 'both at once' especially graphically obvious.

This emphasis seems to be contradictory to the treatment of `artiq_flash`, given that it can also be run from other devices.

I would accept duplicating the group heading "ARTIQ clients" and "ARTIQ controllers", but I don't see why the subheadings also need to be duplicated.

SimonRenblad:
Nitpicking a bit but connecting the striped line from satellite device to only "etc" seems to imply that it doesnt connect to TTLs. It would perhaps be more clear if both the core and satellite device boxes connected to the general "peripherals" box.

architeuthidae:
> This emphasis seems to be contradictory to the treatment of `artiq_flash`, given that it can also be run from other devices.

`artiq_flash` is not really a daily part of a normal workflow, whereas having multiple clients running on various machines absolutely can be. IMO it's covered by the footnote I'm about to push. 

> I would accept duplicating the group heading "ARTIQ clients" and "ARTIQ controllers", but I don't see why the subheadings also need to be duplicated. I believe the non-bolded entries "artiq_dashboard" etc could be removed in the "other machines" box.

(a) to be able to specify e.g. that the dashboard specifically communicates with the proxies; (b) for simple aesthetics. Repetition makes it look more solid. 
 
> Additionally, I would change "(Potentially) remote". The line seems to be implying contrasting it with real-time hardware (ie core device and crates). A better description would perhaps be "PCs and non-real time hardware" (or just "non real-time hardware") or something along those lines that contrasts better with "real-time hardware"

Thumbs up.

architeuthidae:
New draft looks like this: 

image

I have an alternative version which gracefully fails to mention that the remote dashboard talks to the proxy controllers, and looks like this: 

image

which is missing information, but feels a lot more readable -- I think I might vote note to draw the dashboard-proxy connection, and keep that for a later management system diagram.

dnadlinger:
From experience guiding people through the network setup over the years, a good understanding of which parts of the system connect directly to the core device is important (not least because sometimes the core device is put on a separate network with the host PC only). Maybe we could show `aqctl_corelog`/`aqctl_coreanalyzer_proxy` more explicitly? If they are deemed part of the "ARTIQ master", then maybe move the "Built-in controllers" entry to the master box and mention the two controllers in footnote 2?

architeuthidae:
> From experience guiding people through the network setup over the years, a good understanding of which parts of the system connect directly to the core device is important (not least because sometimes the core device is put on a separate network with the host PC only). 

> Maybe just remove the "Other machines" and "Host machines" boxes? They don't add useful information, they are incorrect (the dashboard can run on one of the "other" machines for example) and add clutter to the diagram.

Working with these two suggestions, new draft: 
image
Caption, paraphrased: Most of these components can be run independently of each other, and network-connected components can be spread out among multiple machines.  

Footnotes, paraphrased: 
1. A common workflow for 'client machines' is to install ARTIQ and clone the experiment repository. You can have as many of these as you want. The communication is the same. Other ARTIQ commands and utilities can also be used from these machines.
2. A common workflow is to have controllers for one or several 'slow' devices running on a different machine. In this case, it's necessary to have one controller manager per machine. You can have as many of these as you want. Note that running controllers and controller managers doesn't require a full ARTIQ install, and the minimal package `artiq-comtools` can be installed instead. Otherwise, if these controllers are run next to the built-in controllers, they can be handled by the same controller manager. Just adjust your device database accordingly.  
3. The various other ARTIQ commands may also connect to the master, read the device database, communicate with the core device, etc. See their individual references. 
4. These are the built-in ARTIQ controllers used by ARTIQ clients to provide certain functionalities, listed in Utilities and beginning with `aqctl`; see relevant section of tutorial. If you don't care about these functionalities, it is not necessary to run these, and if you aren't running any controllers at all, you don't need a controller manager either. 
5. In DRTIO configurations only. 

If I haven't lost track, I think this manages to include all the information mentioned thus far. It also has the advantage of clearly visually emphasizing what exactly communicates with the core device. The vague dashed 'comtools machine' / 'client machine' boxes hopefully establish an image for a common way these things can be set up without implying too much of a dichotomy between 'remote and 'host' operations. 

(P.S.: would have pushed this, but I'm home sick and put together this version in Overleaf, so it'll probably wait til tomorrow regardless of verdict).

Spaqin:
Much better now 👍 no need for having a separation between master and clients indeed.

Not sure if the footnote 5 is necessary as the link between master and satellite is already marked as DRTIO, but it won't hurt.

architeuthidae:
> The new diagram looks fine, except that the dataset database is missing.

I believe that the dataset database exclusively belongs to the master -- nothing else really interacts with it? In that case to mention it explicitly I'd add it to the master's node:  

image

architeuthidae:
Current state now (slight spacing changes ended up being necessary so as to render well in LaTeX PDF -- something about the spacing seems to work slightly differently than in the HTML...) 

image

doc/manual/overview.tex
```diff
@@ -0,0 +1,212 @@
...
+      \node[single, master](ddb) {Device database};
+      \&
+      \node[font=\tt, single](coremgmt) {artiq\_coremgmt\textsuperscript{3}};
+      \&
+      \node[single, master](builtin) {Built-in controllers\textsuperscript{4}};
+       \\
+};
+
+% BORDER LINE
+\node[linelabel] (label1) at (-0.25, 6.3) {pcs and remote devices};
```
thomasfire, line 119: is it possible to make it appear as `PCs` instead of `PCS`? Latter may bring some confusion

architeuthidae: Kind of: 
image
as opposed to 
image

Not using smallcaps makes the font noticeably unbalanced. Of course it can also be: 

image

but I must say I prefer the smallcaps.

PR #2559, opened 2024-08-27
Support CoreMgmt over DRTIO

### LINKED ISSUE #2498: Coremgmt support for satellites
There's no other interface with satellites than an USB cable for now. With ``artiq_coremgmt``, we should also be able to:

- set config keys, both string- (``sed_spread_enable``) and file-based (``boot`` on Zynq),
- reboot,
- would be also nice to access logs (setting log levels optional IMO).

On artiq_coremgmt side, it would just require the user to specify the satellite number, e.g.:

```
artiq_coremgmt -s 1 reboot
```

occheung:
# ARTIQ Pull Request

## Description of Changes
This PR implements the followings:
1. Flashing gateware/bootloader/firmware through `artiq_coremgmt` via `config write` similar to Zynq device.
2. Implement corresponding core-mgmt handling for satellites
3. Create new DRTIO messages for (2).

Item (1) should belong to its own PR, but I am not sure if we should toe Zynq's line on this.
Coremgmt now requires a DRTIO destination byte. It is currently implemented as mandatory, so it is a breaking change.

### Related Issue
Closes #2498 

### Known Issues
- ~~Satellite will be unable to respond to AUX traffic while flashing. Taking down DRTIO link and reboot is WIP. We may want to rethink the new DRTIO messages for this.~~ Resolved.
- ~~Routing the new reply messages are probably missing.~~ Only needs to be controlled by the root core device. It should only forward packets, routing is unnecessary.

### Tests
Remote flashing, rebooting, logging (via `artiq_coremgmt` and `aqctl_corelog`), and configuration via DRTIO between 2 core devices. Core devices ranges from Kasli 1.1, 2.0 and Kasli-SoC.

A corresponding PR will be opened to artiq-zynq soon.

## Type of Changes

|   | Type |

### All Pull Requests

- [x] Close/update issues.

### Code Changes

### Git Logistics

  ```
  
  ```

### Licensing

### DIFF
--- a/artiq/frontend/flash_tools.py
+++ b/artiq/frontend/flash_tools.py
@@ -1,6 +1,7 @@
 import atexit
 import os
 import tempfile
+import struct
 
 
 def artifact_path(this_binary_dir, *path_filename, srcbuild=False):
@@ -64,9 +65,6 @@ def fetch_bin(binary_dir, components, srcbuild=False):
 # Copyright 2014-2017 Robert Jordens <jordens@gmail.com>
 # after
 # https://github.com/mfischer/fpgadev-zynq/blob/master/top/python/bit_to_zynq_bin.py
-
-import struct
-
 
 def flip32(data):
     sl = struct.Struct("<I")


artiq/firmware/runtime/mgmt.rs
```diff
@@ -1,132 +1,547 @@
...
+                }
+            }
+        }
+    }
+
+    pub fn config_write(io: &Io, aux_mutex: &Mutex,
+        ddma_mutex: &Mutex, subkernel_mutex: &Mutex, 
+        routing_table: &drtio_routing::RoutingTable, linkno: u8,
+        destination: u8, stream: &mut TcpStream, key: &String, value: &Vec<u8>) -> Result<(), Error<SchedError>> {
+        let mut message = Cursor::new(Vec::with_capacity(key.len() + value.len() + 4 * 2));
```
Spaqin, line 345: + 4 * 2?

occheung: It will transmit both key and value through the DRTIO. Both key and value has a length parameter.

Spaqin: Oh yeah, forgot it's not C-strings we're dealing with.

occheung: Technically we can derive one of them using the size of total payload, but there is already string and bytes IO in the protocol, so I decided against it

artiq/firmware/satman/main.rs
```diff
@@ -489,6 +492,105 @@ fn process_aux_packet(dmamgr: &mut DmaManager, analyzer: &mut Analyzer, kernelmg
             Ok(())
         }
 
+        drtioaux::Packet::CoreMgmtGetLogRequest { destination: _destination, .. } |
+        drtioaux::Packet::CoreMgmtClearLogRequest { destination: _destination } |
+        drtioaux::Packet::CoreMgmtSetLogLevelRequest {destination: _destination, .. } => {
+            forward!(router, _routing_table, _destination, *rank, *self_destination, _repeaters, &packet);
+
+            error!("RISC-V satellite devices do not support buffered logging");
```
Spaqin, line 500: How difficult would it be to implement that? I think it would be useful to have remote log access on RISC-V too.

occheung: I think we just need to change the UARTLogger to BufferLogger, should be straight-forward.

artiq/firmware/satman/main.rs
```diff
@@ -489,6 +492,105 @@ fn process_aux_packet(dmamgr: &mut DmaManager, analyzer: &mut Analyzer, kernelmg
...
+
+            coremgr.add_data(&data, length as usize);
+            let mut succeeded = true;
+            if last {
+                succeeded = coremgr.write_config().is_ok();
+                debug!("Write succeeded: {}", succeeded);
+                coremgr.clear_data();
+            }
+
+            if succeeded {
```
Spaqin, line 566: Why not have a ``CoreMgmtReply { succeeded: bool }`` packet instead? Just for consistency with other aux packets.

occheung: I just found it less symmetric when we deal with requests that has a return payload (e.g. Config read, then it is whatever `ConfigReadReply{ .. }` and `CoreMgmtReply{ false }`, never would we see `CoreMgmtReply{ true }`).

I will change it for consistency.

occheung:
Satellite now disables transceiver TX before flashing, and reboot on its own after. The log from master device while the satellite is flashing looks like this:
```
[    19.307458s]  INFO(runtime::mgmt): new connection from <host ip:port>
[    22.075730s] ERROR(runtime::rtio_mgt::drtio): [DEST#1] communication failed (LinkDown)
[    22.282018s]  INFO(runtime::rtio_mgt::drtio): [LINK#0] link is down
[    22.286957s]  INFO(runtime::rtio_mgt::drtio): [DEST#1] destination is down
[    33.694013s]  INFO(runtime::rtio_mgt::drtio): [LINK#0] link RX became up, pinging
[    37.659062s]  INFO(runtime::rtio_mgt::drtio): [LINK#0] remote replied after 21 packets
[    37.877679s]  INFO(runtime::rtio_mgt::drtio): [LINK#0] link initialization completed
[    37.885080s]  INFO(runtime::rtio_mgt::drtio): [DEST#1] destination is up
[    37.890785s]  INFO(runtime::rtio_mgt::drtio): [DEST#1] buffer space is 128
```

artiq/firmware/satman/mgmt.rs
```diff
@@ -0,0 +1,82 @@
...
+        };
+
+        let value = self.current_payload.read_bytes().unwrap();
+
+        match key.as_str() {
+            "gateware" | "bootloader" | "firmware" => {
+                drtioaux::send(0, &drtioaux::Packet::CoreMgmtAck)?;
+                #[cfg(not(soc_platform = "efc"))]
+                unsafe {
+                    clock::spin_us(10000);
```
Spaqin, line 59: Is the wait there to make sure the message is sent out?

occheung: Yes. There are no specific reason for picking 10ms.

occheung: The delay also compensates the time the master devices need to receive the message.
With a delay of 10us, the master device has a large chance to detect link down before processing the message, so the `remote_coremgmt` module would report `aux packet error (link down)`.
100us seems to work reliably in a 2 RISC-V devices setup (master <-> satellite).

artiq/firmware/satman/mgmt.rs
```diff
@@ -0,0 +1,82 @@
...
+                self.clear_data();
+                error!("error on reading key: {:?}", err);
+                return drtioaux::send(0, &drtioaux::Packet::CoreMgmtNack);
+            }
+        };
+
+        let value = self.current_payload.read_bytes().unwrap();
+
+        match key.as_str() {
+            "gateware" | "bootloader" | "firmware" => {
```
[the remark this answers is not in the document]
occheung: Having some other instruction to prompt the flash looks hacky, especially when we implement this under `config write`. The config isn't really written in this case.

And there are 3 binaries we need to flash. Maybe we make a new `coremgmt` action to specifically deal with this instead?

dnadlinger:
Is there a sufficient amount of checksumming on the binary payloads before they are written to flash? We have been (very occasionally, but still) seeing what looks like data corruption on the DRTIO links (aux packet error log messages), and potentially writing out corrupted gateware/firmware sounds like a debugging nightmare.

hartytp:
> seeing what looks like data corruption on the DRTIO links (aux packet error log messages), and potentially writing out corrupted gateware/firmware sounds like a debugging nightmare.

Side note: I suspect this may be a widespread issue which people just aren't noticing. I do wonder whether having a checksum on DRTIO would be a good idea to make this kind of thing more obvious and easier to debug?

artiq/frontend/artiq_coremgmt.py
```diff
@@ -85,6 +89,24 @@ def get_argparser():
     t_boot = tools.add_parser("reboot",
                               help="reboot the running system")
 
+    # flashing
+    t_flash = tools.add_parser("flash",
+                               help="flash the running system")
+
+    p_zynq = t_flash.add_argument("-z", "--zynq", default=False,
```
Spaqin, line 96: any chance for auto detection?

RELEASE_NOTES.rst
```diff
@@ -19,6 +19,8 @@ ARTIQ-9 (Unreleased)
 * Python 3.12 support.
 * Compiler can now give automatic suggestions for ``kernel_invariants``. 
 * Idle kernels now restart when written with ``artiq_coremgmt`` and stop when erased/removed from config.
+* Support for coredevice reflashing through new the ``flash`` tool in ``artiq_coremgmt``.
```
architeuthidae, line 22: typo: 'through new the'

artiq/frontend/artiq_coremgmt.py
occheung: I guess we can simply try both riscv and zynq binaries, then go for the one that succeed.

artiq/firmware/satman/main.rs
```diff
@@ -495,6 +500,130 @@ fn process_aux_packet(dmamgr: &mut DmaManager, analyzer: &mut Analyzer, kernelmg
...
+            if last {
+                drtioaux::send(0, &drtioaux::Packet::CoreMgmtDropLink)
+            } else {
+                drtioaux::send(0, &drtioaux::Packet::CoreMgmtReply { succeeded: true })
+            }
+        }
+        drtioaux::Packet::CoreMgmtDropLinkAck { destination: _destination } => {
+            forward!(router, _routing_table, _destination, *rank, *self_destination, _repeaters, &packet);
+
+            #[cfg(not(soc_platform = "efc"))]
```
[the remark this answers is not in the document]
occheung: https://github.com/m-labs/artiq/blob/333623e24bdec00783bc89c1e8b6b49a74bc9e1c/artiq/firmware/satman/main.rs#L755-L758
This checks not EFC, though I do agree it should check for DRTIO EEM as well.
Probably not an intended change in #2410, as pointed out in #2134. I will make a PR to revert this.

artiq/frontend/artiq_coremgmt.py
Spaqin: Just like runtime/satman is currently detected.

occheung:
The runtime firmware currently does not support the termination of EFC's DRTIO connection, we may need to implement reconnect for DRTIO-over-EEM as well.

artiq/frontend/artiq_coremgmt.py
```diff
@@ -138,6 +162,39 @@ def main():
...
+                bins = []
+                for bin_name in bin_list:
+                    bins.append(fetch_bin(
+                        args.directory, bin_name, args.srcbuild))
+                retrieved_bins.append(bins)
+            except FileNotFoundError:
+                pass
+
+        if retrieved_bins is None:
+            raise FileNotFoundError("both risc-v and zynq binaries not found")
```
architeuthidae, line 189: nitpick: probably 'neither risc-v nor zynq'

Issue #2652, opened 2025-01-14
Artiq compiler reads wrong return types from `@rpc def return_many(): return ([4], [True],)`


dimitsev:
# Bug Report

## One-Line Summary

In the function below, the Artiq compiler thinks that both return types are `TList[TInt32]`:

```
def return_many() -> TTuple([TList(TInt32), TList(TBool)]):
    return ([4], [True],)
```

## Issue Details

[https://m-labs.hk/artiq/manual-beta/compiler.html#basic-artiq-python](https://m-labs.hk/artiq/manual-beta/compiler.html#basic-artiq-python) explicitly shows this tip, which we have tested with success:

```
def return_many() -> TTuple([TInt32, TFloat, TStr]):
    return (4, 12.34, "hello",)
```

Surprisingly, this also works for us:

```
def return_many() -> TTuple([TList(TInt32), TList(TFloat), TList(TStr)]):
    return ([4], [12.34], ["hello"],)
```

However, our function at the top with signature `-> TTuple([TList(TInt32), TList(TBool)])` does **not** work for us.

Maybe there is a problem that only affects `TBool` or `TList(TBool)`?

### Minimal self-contained examples

```
from artiq.experiment import EnvExperiment, kernel, rpc, host_only
from artiq.language.types import TNone, TBool, TInt32, TFloat, TStr, TList, TTuple

def return_many() -> TTuple([TList(TInt32), TList(TBool)]):
    return ([4], [True],)

def get_type(obj) -> TStr:
    return str(type(obj))

class TestSatellites(EnvExperiment):
    
    def build(self) -> TNone:
        self.setattr_device("core")
        (self.a, self.b) = return_many()
        print(" ----    @rpc-environment ----")
        print("type(a) =", get_type(self.a[0]))
        print("type(b) =", get_type(self.b[0]))
    
    @kernel
    def run(self) -> TNone:
        self.core.reset()
        print(" ---- @kernel-environment ----")
        print("type(a) =", get_type(self.a[0]))
        print("type(b) =", get_type(self.b[0]))
```

outputs

```
 ----    @rpc-environment ----
type(a) = <class 'int'>
type(b) = <class 'bool'>
 ---- @kernel-environment ----
type(a) = <class 'numpy.int32'>
type(b) = <class 'numpy.int32'>
```

whereas

```
from artiq.experiment import EnvExperiment, kernel, rpc, host_only
[...]

PR #2654, opened 2025-01-17
compiler: fix TList[TBool] inferred as TList[TInt32] in embedding

SimonRenblad:
# ARTIQ Pull Request

## Description of Changes

See discussion in #2652. 

### Related Issue

#2652 

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

### All Pull Requests

- [ ] Close/update issues.

### Code Changes

The following minimal example can be used to verify the behavior.
```python

def get_type(obj) -> TStr:
    return str(type(obj))

class TestBoolList(EnvExperiment):
    def build(self):
        self.setattr_device("core")
        self.x = [True]

    @kernel
    def run(self):
        print(get_type(self.x[0]))

```
Output before: `<class='numpy.int32'>`
Output after: `<class='bool'>`. 

Unit tests:
`test_embedding` and `test_performance` both pass on a Kasli-SoC.

### Git Logistics

  ```
  
  ```

### Licensing

### DIFF
--- a/artiq/compiler/embedding.py
+++ b/artiq/compiler/embedding.py
@@ -310,7 +310,9 @@ class ASTSynthesizer:
         if len(value) > 0:
             v = value[0]
             is_T = True
-            if isinstance(v, int):
+            if isinstance(v, bool):
+                is_T = False
+            elif isinstance(v, int):
                 T = int
             elif isinstance(v, float):
                 T = float


dnadlinger:
Wouldn't a regression test be pretty straightforward to add to test_embedding?

artiq/compiler/embedding.py
```diff
@@ -310,7 +310,9 @@ def fast_quote_list(self, value):
         if len(value) > 0:
             v = value[0]
             is_T = True
-            if isinstance(v, int):
+            if isinstance(v, bool):
+                is_T = False
```
dnadlinger, line 314: Does this just work with `T = bool` instead (keeping the fast path)?

SimonRenblad: It's not a one-line change. Additional logic would need to be added a few lines down to handle the `bool` case. Prior to the bug fix, the `elts` were simply coerced to `int` here:

```python
for i, (v, t) in enumerate(zip(value, text)):
    l = len(t)
    elts[i] = asttyped.NumT(
        n=int(v), ctx=None, type=typ,
        loc=source.Range(
            self.source_buffer, start, start + l,
            expanded_from=self.expanded_from))
    start += l + 2
```

In terms of implementation, adding another case for boolean lists here is simple. However, from the code above and the link below it seemed to me that the intention was for boolean lists **not** to be on the fast path.  

https://github.com/m-labs/artiq/blob/ebb4058d09226a0900ca8ede8e0b87299fb884ea/artiq/compiler/embedding.py#L757

That being said, if you think its warranted to change this behavior, I'll see about creating a PR for it.

PR #2657, opened 2025-01-22
Urukul: Update artiq.coredevice.urukul and artiq.coredevice.ad9910 for new Urukul capabilities

newell:
# ARTIQ Pull Request

## Description of Changes

This PR adds coredevice driver capabilities and tests for new Urukul CPLD functionaly (see related issue).

such as:

1. Decoupling the individual DDS channels.
2. Individual DDS Channel:
   * PROFILE
   * IO_UPDATE
   * OSK
   * DRCTRL
   * DRHOLD
   * ATT_EN
3. Digital Ramp Generator (DRG)
4. Backward compatible Urukul coredevice driver that can work between protocol revision 8 and new editions.

### Related Issue

https://github.com/quartiq/urukul/issues/7

## Type of Changes

|   | Type |

### All Pull Requests

### Code Changes

### Licensing

### DIFF
--- a/artiq/coredevice/ad9910.py
+++ b/artiq/coredevice/ad9910.py
@@ -13,18 +13,10 @@ urukul_sta_smp_err = urukul.urukul_sta_smp_err
 
 __all__ = [
     "AD9910",
-    "PHASE_MODE_CONTINUOUS",
-    "PHASE_MODE_ABSOLUTE",
-    "PHASE_MODE_TRACKING",
-    "RAM_DEST_FTW",
-    "RAM_DEST_POW",
-    "RAM_DEST_ASF",
-    "RAM_DEST_POWASF",
-    "RAM_MODE_DIRECTSWITCH",
-    "RAM_MODE_RAMPUP",
-    "RAM_MODE_BIDIR_RAMP",
-    "RAM_MODE_CONT_BIDIR_RAMP",
-    "RAM_MODE_CONT_RAMPUP",
+    "PHASE_MODE_CONTINUOUS", "PHASE_MODE_ABSOLUTE", "PHASE_MODE_TRACKING",
+    "RAM_DEST_FTW", "RAM_DEST_POW", "RAM_DEST_ASF", "RAM_DEST_POWASF",
+    "RAM_MODE_DIRECTSWITCH", "RAM_MODE_RAMPUP", "RAM_MODE_BIDIR_RAMP",
+    "RAM_MODE_CONT_BIDIR_RAMP", "RAM_MODE_CONT_RAMPUP",    
 ]
 
 _PHASE_MODE_DEFAULT = -1
@@ -144,33 +136,13 @@ class AD9910:
         to the same string value.
     """
 
-    def __init__(
-        self,
-        dmgr,
-        chip_select,
-        cpld_device,
-        sw_device=None,
-        pll_n=40,
-        pll_cp=7,
-        pll_vco=5,
-        sync_delay_seed=-1,
-        io_update_delay=0,
-        pll_en=1,
-    ):
-        self.kernel_invariants = {
-            "cpld",
-            "core",
-            "bus",
-            "chip_select",
-            "pll_en",
-            "pll_n",
-            "pll_vco",
-            "pll_cp",
-            "ftw_per_hz",
-            "sysclk_per_mu",
-            "sysclk",
-            "sync_data",
-        }
+    def __init__(self, dmgr, chip_select, cpld_device, sw_device=None,
+                 pll_n=40, pll_cp=7, pll_vco=5, sync_delay_seed=-1,
+                 io_update_delay=0, pll_en=1):
+        self.kernel_invariants = {"cpld", "core", "bus", "chip_select",
+                                  "pll_en", "pll_n", "pll_vco", "pll_cp",
+                                  "ftw_per_hz", "sysclk_per_mu", "sysclk",
+                                  "sync_data"}
         self.cpld = dmgr.get(cpld_device)
         self.core = self.cpld.core
         self.bus = self.cpld.bus
@@ -189,14 +161,8 @@ class AD9910:
             assert clk <= 60e6
             assert 12 <= pll_n <= 127
             assert 0 <= pll_vco <= 5
-            vco_min, vco_max = [
-                (370, 510),
-                (420, 590),
-                (500, 700),
-                (600, 880),
-                (700, 950),
-                (820, 1150),
-            ][pll_vco]
+            vco_min, vco_max = [(370, 510), (420, 590), (500, 700),
+                                (600, 880), (700, 950), (820, 1150)][pll_vco]            
             assert vco_min <= sysclk / 1e6 <= vco_max
             assert 0 <= pll_cp <= 7
         else:
@@ -211,13 +177,12 @@ class AD9910:
 
         if isinstance(sync_delay_seed, str) or isinstance(io_update_delay, str):
             if sync_delay_seed != io_update_delay:
-                raise ValueError(
-                    "When using EEPROM, sync_delay_seed must be "
-                    "equal to io_update_delay"
-                )
+                raise ValueError("When using EEPROM, sync_delay_seed must be "
+                                 "equal to io_update_delay")                
             self.sync_data = SyncDataEeprom(dmgr, self.core, sync_delay_seed)
         else:
-            self.sync_data = SyncDataUser(self.core, sync_delay_seed, io_update_delay)
+            self.sync_data = SyncDataUser(self.core, sync_delay_seed,
+                                          io_update_delay)
 
         self.phase_mode = PHASE_MODE_CONTINUOUS
 
@@ -271,10 +236,9 @@ class AD9910:
         :param addr: Register address
         :param data: Data to be written
         """
-        self.bus.set_config_mu(
-            urukul.SPI_CONFIG | spi.SPI_END, 24, urukul.SPIT_DDS_WR, self.chip_select
-        )
-        self.bus.write((addr << 24) | ((data & 0xFFFF) << 8))
+        self.bus.set_config_mu(urukul.SPI_CONFIG | spi.SPI_END, 24,
+                               urukul.SPIT_DDS_WR, self.chip_select)
+        self.bus.write((addr << 24) | ((data & 0xFFFF) << 8))        
 
     @kernel
     def write32(self, addr: TInt32, data: TInt32):
@@ -283,13 +247,11 @@ class AD9910:
         :param addr: Register address
         :param data: Data to be written
         """
-        self.bus.set_config_mu(
-            urukul.SPI_CONFIG, 8, urukul.SPIT_DDS_WR, self.chip_select
-        )
+        self.bus.set_config_mu(urukul.SPI_CONFIG, 8,
+                               urukul.SPIT_DDS_WR, self.chip_select)        
         self.bus.write(addr << 24)
-        self.bus.set_config_mu(
-            urukul.SPI_CONFIG | spi.SPI_END, 32, urukul.SPIT_DDS_WR, self.chip_select
-        )
+        self.bus.set_config_mu(urukul.SPI_CONFIG | spi.SPI_END, 32,
+                               urukul.SPIT_DDS_WR, self.chip_select)        
         self.bus.write(data)
 
     @kernel
@@ -298,16 +260,12 @@ class AD9910:
 
         :param addr: Register address
         """
-        self.bus.set_config_mu(
-            urukul.SPI_CONFIG, 8, urukul.SPIT_DDS_WR, self.chip_select
-        )
+        self.bus.set_config_mu(urukul.SPI_CONFIG, 8,
+                               urukul.SPIT_DDS_WR, self.chip_select)        
         self.bus.write((addr | 0x80) << 24)
         self.bus.set_config_mu(
             urukul.SPI_CONFIG | spi.SPI_END | spi.SPI_INPUT,
-            16,
-            urukul.SPIT_DDS_RD,
-            self.chip_select,
-        )
+            16, urukul.SPIT_DDS_RD, self.chip_select)            
         self.bus.write(0)
         return self.bus.read()
 
@@ -317,16 +275,12 @@ class AD9910:
 
         :param addr: Register address
         """
-        self.bus.set_config_mu(
-            urukul.SPI_CONFIG, 8, urukul.SPIT_DDS_WR, self.chip_select
-        )
+        self.bus.set_config_mu(urukul.SPI_CONFIG, 8,
+                               urukul.SPIT_DDS_WR, self.chip_select)        
         self.bus.write((addr | 0x80) << 24)
         self.bus.set_config_mu(
             urukul.SPI_CONFIG | spi.SPI_END | spi.SPI_INPUT,
-            32,
-            urukul.SPIT_DDS_RD,
-            self.chip_select,
-        )
+            32, urukul.SPIT_DDS_RD, self.chip_select)
         self.bus.write(0)
         return self.bus.read()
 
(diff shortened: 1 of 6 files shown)


Spaqin:
Has this been tested with real hardware?

newell:
> Has this been tested with real hardware?

Yes.  If you would like to run some tests yourself you can flash the CPLD and then run `test_ad9910_waveform.py` to see some of the new functionality in action on your oscilloscope.

I’ve uploaded a video to Google Drive demonstrating the new capabilities (no guarantee for future readers that this link will remain active):

https://drive.google.com/file/d/18Y5HMqlYAk4xTw7fwv_1mqRqKBfi1A85/view?usp=sharing

The tests executed during the video were:

```
$ python3 -m unittest test_ad9910_waveform.py
Running test: test_att_single_tone
.sRunning test: test_drg_nodwell_amplitude
.sRunning test: test_drg_nodwell_frequency
.sRunning test: test_drg_nodwell_phase
.sRunning test: test_drg_normal_amplitude
.sRunning test: test_drg_normal_frequency
.sRunning test: test_drg_normal_phase
.sRunning test: test_drg_normal_with_hold_amplitude
.sRunning test: test_drg_normal_with_hold_frequency
.sRunning test: test_drg_normal_with_hold_phase
.s..Running test: test_osk_auto
.sRunning test: test_osk_manual
.sRunning test: test_single_tone
.sRunning test: test_toggle_profiles
.s
Ran 30 tests in 189.314s

OK (skipped=14)
```

artiq/coredevice/urukul.py
```diff
@@ -156,19 +677,21 @@ class CPLD:
...
     """
+
     kernel_invariants = {"refclk", "bus", "core", "io_update", "clk_div"}
 
     def __init__(self, dmgr, spi_device, io_update_device=None,
                  dds_reset_device=None, sync_device=None,
                  sync_sel=0, clk_sel=0, clk_div=0, rf_sw=0,
                  refclk=125e6, att=0x00000000, sync_div=None,
-                 core_device="core"):
+                 proto_rev=0x08, core_device="core"):
```
Spaqin, line 694: Preferably proto_rev would be pulled from the device itself during init.

artiq/coredevice/urukul.py
```diff
@@ -124,6 +135,516 @@ def set_mu(self, ftw: TInt32):
...
+        cpld.bus.set_config_mu(
+            SPI_CONFIG | spi.SPI_END | spi.SPI_INPUT, 28, SPIT_CFG_RD, CS_CFG
+        )
+        cpld.bus.write((cpld.cfg_reg & 0xFFFFFFF) << 4)
+        hi = cpld.bus.read()
+        lo = cpld.bus.read()
+        return (int64(hi) << 28) | lo
+
+    @kernel
+    def init(self, cpld):
```
Spaqin, line 485: Inits for both 8 and 9 are missing proto_rev checks from STA. It's especially important to verify especially during the transitional period, and the simple side effect that a proto_rev check would also verify the presence of the device (e.g. would give a mismatch if EEM cables were mixed up)

Spaqin:
Would this be a part of ARTIQ-8? How about the ICE version later, considering CPLD is EOL and ARTIQ-9 is a long way ahead still? There's some changes breaking backwards compatibility in the driver (``set_profile`` now requires a channel parameter).

artiq/coredevice/urukul.py
```diff
@@ -191,77 +714,112 @@ def __init__(self, dmgr, spi_device, io_update_device=None,
             assert sync_div is None
             sync_div = 0
 
-        self.cfg_reg = urukul_cfg(rf_sw=rf_sw, led=0, profile=DEFAULT_PROFILE,
-                                  io_update=0, mask_nu=0, clk_sel=clk_sel,
-                                  sync_sel=sync_sel,
-                                  rst=0, io_rst=0, clk_div=clk_div)
+        self.proto_rev = proto_rev
+        self.version_manager = CPLDVersionManagerFactory.get_version(proto_rev)
```
Spaqin, line 718: generally agreed that ``version_manager`` is not the best name (it's just the CPLD itself after all, with an abstraction layer for different protocol versions), and the Factory could be skipped altogether - just create the objects directly in the ``if`` below.

artiq/coredevice/urukul.py
newell: This was the route I tried originally, but not knowing which version doesn't work well with the compiler. That is, if I probe in `init()` for the `proto_rev`, once I have this information I then need to set which `self.cfg_reg` will be but the compiler wants to know beforehand what type this will be.  I don't see a way around this using the legacy compiler.  Additionally, `UrukulMonitor` relies on this information (I had forgotten to add this portion to the PR) so this will also need to be supplied beforehand as well.

Spaqin: I understand that's a pain and there seems to be no way around it (could ``self.cfg_reg`` be a part of ``version``)? And even if you can't auto-detect and apply the correct driver, it should at very least verify that the user-provided version matches up with what Urukul is reporting.

artiq/coredevice/urukul.py
```diff
@@ -124,6 +135,490 @@ def set_mu(self, ftw: TInt32):
...
+        """
+        cpld.bus.set_config_mu(SPI_CONFIG | spi.SPI_INPUT, 24, SPIT_CFG_RD, CS_CFG)
+        cpld.bus.write(((cpld.cfg_reg >> 24) & 0xFFFFFF) << 8)
+        cpld.bus.set_config_mu(
+            SPI_CONFIG | spi.SPI_END | spi.SPI_INPUT, 28, SPIT_CFG_RD, CS_CFG
+        )
+        cpld.bus.write((cpld.cfg_reg & 0xFFFFFFF) << 4)
+        hi = cpld.bus.read()
+        lo = cpld.bus.read()
+        return (int64(hi) << 28) | lo
```
occheung, line 474: The SR has 27 bits only. What is the significance of reporting `hi` as well instead of just discarding after read?

artiq/coredevice/urukul.py
newell: In regards to `self.cfg_reg` being part of the `self.version` instance, I would still need to pass in all the needed parameters from `__init__` to create the configuration register, so I am not really sure how this would help as I would need to create placeholders in `__init__` so that I could then use them within `init` depending on what probing would find, so currently I don't see a way around this.  Hopefully I understood what you were implying.

In regards to verifying that the user-provided versions match, yes I agree that would be a good addition and will add that.

artiq/coredevice/urukul.py
occheung: Actually, I think you can set the config of the first SPI transaction to not generate an RTIO input in the first place.

newell: > Actually, I think you can set the config of the first SPI transaction to not generate an RTIO input in the first place.

I am not sure what you mean here.  Can you elaborate?

occheung: Is it possible to not set SPI_INPUT flag for the first 24-bits SPI transaction, and leave the second 28-bits transaction as-is (with SPI_INPUT)? If you do this, the RTIO core should only generate an input for the second 28-bits transaction only.

newell:
I believe I have addressed all the review comments so far.  Is there anything else needed to get this merged?

artiq/coredevice/urukul.py
```diff
@@ -124,6 +135,492 @@ def set_mu(self, ftw: TInt32):
...
+        return cpld.bus.read()
+
+    @kernel
+    def init(self, cpld):
+        """Initialize Urukul with ProtoRev8.
+
+        Resets the DDS I/O interface.
+        Does not pulse the DDS ``MASTER_RESET`` as that confuses the AD9910.
+        """
+        if cpld.proto_rev != STA_PROTO_REV_8:
```
Spaqin, line 307: I meant, can you read STA from the Urukul itself to compare the proto rev? This test here will always pass as you set ``proto_rev`` in __init__ yourself anyway.

newell: Well that is slightly embarrassing.  When I had read your original message I knew what you were talking about and in my mind I was planning on doing that but when I got around to implementing it I went blank on that.  Fix pushed.

Spaqin (approved):

artiq/coredevice/ad9910.py
```diff
@@ -933,6 +953,51 @@ def cfg_sw(self, state: TBool):
         """
         self.cpld.cfg_sw(self.chip_select - 4, state)
 
+    @kernel
+    def cfg_osk(self, state: TBool):
+        """Set CPLD CFG OSK state. The OSK bit is controlled by the
+        logical or of the CPLD configuration shift register OSK bit.
```
architeuthidae, line 959: suggest "logical OR" here and below just to make the sentence less of a [garden path](https://en.wikipedia.org/wiki/Garden-path_sentence). also, OR of the shift register bit with what?

newell: Thanks for catching this, I had blindingly followed `cfg_sw` doc-string.  This, nor the others, have a logical OR.

newell:
I have tested this a bunch on my end, but it would be nice if I could get someone else to test this on their end, if possible, for some extra stress testing as this is a large branch.

artiq/coredevice/ad9910.py
```diff
@@ -895,7 +911,9 @@ def get(self, profile: TInt32 = DEFAULT_PROFILE
     def set_att_mu(self, att: TInt32):
         """Set digital step attenuator in machine units.
 
-        This method will write the attenuator settings of all four channels. See also
+        This method will write the attenuator settings of the channel
```
architeuthidae, line 914: Not very clear. Will write the attenuator settings of _this_ channel, I suppose? Accordingly `(With Urukul proto_rev 0x08...` or `(For Urukul proto_rev 0x08...` would also make more sense.

Spaqin:
SUServo was tested successfully with a Kasli 2.0 after few minor compiler/driver setbacks:
- cfg_reg was forced to be int64 due to compiler incompatibilities (seems like the original cfg for proto_rev 9 was interpreted as int32 even though it's 54 bits)
- blind init was re-added - there is no readback in SUServo mode, so we have to trust that the protocol version was chosen correctly.

occheung:
The new way of handling `dds.io_update` looks odd.

If the DDS doesn't instantiate with an `io_update` device, we "pulse" `IO_UPDATE` by writing the CFG. But `MASK_NU` is off by default, so it does nothing by default.

If we do instantiate with an `io_update` device, we actually pulse `IO_UPDATE` TTL. But it updates all `IO_UPDATE` of all DDSes (again without `MASK_NU`). You wouldn't expect this to happen when toggling other auxiliary TTLs like  `dds.sw`.

Should we really expose these quirks to the API? To use the new I/O update correctly, user would have read the urukul gateware to know MASK_NU, and they would have known the individual I/O update. I think it should be considered advanced features, and let users figure it out when needed.

newell:
> The new way of handling `dds.io_update` looks odd.
> 
> If the DDS doesn't instantiate with an `io_update` device, we "pulse" `IO_UPDATE` by writing the CFG. But `MASK_NU` is off by default, so it does nothing by default.
> 
> If we do instantiate with an `io_update` device, we actually pulse `IO_UPDATE` TTL. But it updates all `IO_UPDATE` of all DDSes (again without `MASK_NU`). You wouldn't expect this to happen when toggling other auxiliary TTLs like `dds.sw`.
> 
> Should we really expose these quirks to the API? To use the new I/O update correctly, user would have read the urukul gateware to know MASK_NU, and they would have known the individual I/O update. I think it should be considered advanced features, and let users figure it out when needed.

My aim was to expose maximum control. I agree the current default behavior for `dds.io_update` regarding `MASK_NU` and the effect of the non-TTL `io_update` device isn't intuitive and could cause confusion (as you mentioned, a user would need to know that they need to set `MASK_NU` for this to work).

If more people would prefer this, I'm open to simplifying the behavior of `dds.io_update` by handling `MASK_NU` for the non-TTL `io_update` device.  This would at least abstract out the `MASK_NU` portion from the user and would force what is done with `MASK_NU`.  However, and although a use case doesn't come to my mind at the moment, this might not be what some users want and hence why I left the control up to the user.

jordens:
No-no. That's patently untrue. There were unresolved issues with that PR and I did not indicate that I wanted to test it.

dnadlinger:
> So what? It's not a race to the bottom about who can get away with the buggiest code.

Sure, but in the context of coming up with a strategy for collaboration in an open-source project, it is somewhat relevant that the issue already existed in the same form previously. Yes, considering purely the totality of development work in the abstract, i.e. divorced from the fact that it is divided between different people and organisations (each with their own constraints), it would be preferrable to first clear out any technical debt before building on a given piece of code. But from Peter's/my perspective, higher standards were applied to our contribution than to what existed before. The original author of the questionable piece of code couldn't (or wasn't inclined to) provide an explanation either; not ideal, sure, but since that was already deployed, just putting a pin in it and concentrating on the new functionality doesn't seem unreasonable.

Needless to say, it took a meaningful amount of work to develop phase-coherent SUServo (this is the second design iteration after hitting a badly documented aspect of the AD9910 phase autoclear behaviour), and we chose to spend the extra work on trying to upstream it for all the usual reasons of open-source collaboration. I hope you can see how then being asked to resolve issues already present in the code first does not seem proportionate. If this had been something obvious, sure, then it would have made sense to tackle it there and then, but as Peter said, it wasn't clear to either me or him how exactly this was supposed to work in the first place, or what possible mitigations would be given the constrained environment of the CPLD on Urukul <=1.3(?).

> Uh, that's not the reason at all. The reasons were the technical issues with the PR and lack of testing and forward development - the PR was more or less just another OSS code dump.

Could you elaborate on this? Looking back through the PR history, I concede that I could have stepped in making sure this progresses along and trying to find other workable ways to get this merged after Peter's enthusiasm to spend any more evenings on this with vague prospects was dimished (given that this all was seemingly held up by quartiq/urukul#11 being deadlocked) and the PR threads had fizzled out (with Peter being the last one active, I might add). Still, at this point
 - the code is probably more well-tested than most regular commits made to non-unit-tested parts of ARTIQ, given that it has been running for 5+ years across more than half a dozen deployments (yes, it would be great to have HITL tests, but we don't have them for the existing code either), and
 - we have been maintaining it internally anyway (assuming that this is what you meant with forward development) to keep our experiments up to date with upstream ARTIQ.

I'm not trying to assign blame to you here; I am all too aware that making demands of other people's time, i.e. acting from a position of entitlement, is not a fruitful approach in open source collaborations. Still, I hope you can look at this in retrospect and recognise a situation where you did have a technical volunteer show up with a broadly solid, non-trivial contribution (even if perhaps less well-versed in the folkways of open source development than others), only for them to ultimately give up in frustration. You have previously expressed dismay at the relative dearth of contributions from users (apart from monentary ones through hardware purchasing) – a frustration I share. At the same time, it is hard for me not to see instances such as this one as an example of failure in community building.

My main interest at this point, however, is not to rehash a situation from five years ago (though I suppose doing so might ultimately also lead to improved outcomes in similar cases in the future), but to figure out how we can best coordinate on SUServo going forward. As long as the patches aren't actually integrated into upstream proper, we'll have no choice but to maintain a fork implementing this for the `proto_rev` 8 variant internally anyway (and will be happy to provide updated diffs if necessary), but e.g. there is little incentive for us to also keep track of the protocol changes/refactorings upstream unless there is a clear plan where that leads to a beneficial outcome.

> though not sure if it would be on time for the ARTIQ-9 release.

From the end user point of view – i.e. excluding those of us who are happy to build and Urukul gateware themselves and have a internal system for hardware/gateware revision management in place –, I think it would be good to integrate any changes into the first ARTIQ release and hardware supplied with `proto_rev` 9, just to minimise the complexity in terms of different protocol and gateware variants to keep track of.

dnadlinger:
@newell Just made a tracking issue here to focus the conversation: https://git.m-labs.hk/M-Labs/urukul-pld/issues/6.

PR #2663, opened 2025-01-29
Fix throwing non-embedded builtin exceptions

SquidDev:
# ARTIQ Pull Request

## Description of Changes
`comm_kernel` currently encodes all (Python) builtin exceptions with an id of 0. However, lookup for these exceptions happens from the `artiq.coredevice.exceptions` module. This means if you try to throw a builtin exception not in the embedding map, then it crashes when receiving it on the host. This also prevents you catching these exceptions in kernel code.

We now encode only use an exn id of 0 if the exception is in the exceptions module (and thus already in the embedding map).

I believe this is a relatively recent regression — I noticed this while in the process of updating to latest ARTIQ. This was possibly introduced in #2526, but afraid I have not bisected this.

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

### All Pull Requests

### Code Changes

### DIFF
--- a/artiq/coredevice/comm_kernel.py
+++ b/artiq/coredevice/comm_kernel.py
@@ -617,10 +617,10 @@ class CommKernel:
                 self._write_int32(embedding_map.store_str(function))
             else:
                 exn_type = type(exn)
-                if exn_type in builtins.__dict__.values():
-                    name = "0:{}".format(exn_type.__qualname__)
-                elif hasattr(exn, "artiq_builtin"):
+                if hasattr(exn, "artiq_builtin"):
                     name = "0:{}.{}".format(exn_type.__module__, exn_type.__qualname__)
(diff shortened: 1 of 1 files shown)


AR-PyT:
I was under the impression that we predefined the set of builtin exceptions we would use in `artiq.coredevice.exceptions` and that these were the only exceptions we would see in a normal use-case (the PR fixes this assumption). 
On the kernel end however, if we had to raise a builtin exception for instance the `IsADirectoryError`, we would still need to add it to the `EmbeddingMap` otherwise the lookup in `ksupport::eh_artiq::get_exception_id` will fail.
Either we can either store other builtin exceptions as objects (done in this PR) if we do not intend to raise them from kernel or add the definitions of other builtin exceptions to the `EmbeddingMap` (currently 13 added out of 52 in python builtins module included).

Issue #2031, opened 2022-12-14
[kasli] i2c gpio expander and i2c eeprom don't check ACK/NAK


jordens:
# Bug Report

## One-Line Summary

NAK on I2C write needs to be handled

## Issue Details

https://github.com/m-labs/artiq/blob/874d298ceb68ffa776592f855365ca5e2e885575/artiq/firmware/libboard_misoc/io_expander.rs#L53-L55
https://github.com/m-labs/artiq/blob/a85b4d5f5ed80e50f1005530b16570057672d770/artiq/firmware/libboard_misoc/i2c_eeprom.rs#L43-L47

### Your System (omit irrelevant parts)

* ARTIQ version: current release-7, master
* Hardware involved: Kasli

PR #2694, opened 2025-03-05
i2c: make write nack an error, force to handle nacks

Spaqin:
# ARTIQ Pull Request

## Description of Changes

Looking at recent I2C changes, I stumbled on #2031 and it seemed like a small change.

This PR changes the behavior of the I2C driver in Kasli. Rewrote the errors, they're now enums rather than static str. Additionally, NACKs are now an Error and thus cannot be ignored.

Aware that NACKs can be used if the I2C receiver controls the information flow (``During the transfer, the receiver cannot receive any more data bytes.`` from I2C spec). This situation can still be caught for coredevice drivers, even though in practice it's not used.

Tested with kc705 HITL, SPI/I2C still OK.

### Related Issue

#2031 

## Type of Changes

|   | Type |
| ✓  | :bug: Bug fix  |

### All Pull Requests

### Code Changes

### Git Logistics

  ```
  
  ```

### Licensing


artiq/firmware/runtime/kern_hwreq.rs
```diff
@@ -90,7 +91,7 @@ mod remote_i2c {
     pub fn write(io: &Io, aux_mutex: &Mutex, ddma_mutex: &Mutex, subkernel_mutex: &Mutex,
         routing_table: &drtio_routing::RoutingTable,
         linkno: u8, destination: u8, busno: u8, data: u8
-    ) -> Result<bool, &'static str> {
+    ) -> Result<(), &'static str> {
```
SimonRenblad, line 94: Why do you not use the same error enum for both remote and local? Since it seems you use it here anyways and then match the str down in `process_kern_kwreq`.

Spaqin: Right. I got tripped up since the same macro is used for SPI as well, but actually it doesn't matter, I can generate an i2c/spi::Error anyway.

artiq/firmware/libboard_misoc/i2c.rs
```diff
@@ -1,8 +1,34 @@
...
+    NoI2C,
+    InvalidBus,
+    Nack,
+    SCLLow,
+    SDALow,
+    ArbitrationLost,
+    IOExpanderError,
+    OtherError,
+}
+
```
SimonRenblad, line 12: implement `From<Error> for &str` and you won't need to do all the `map_err` stuff afaik.

SimonRenblad: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=32fce25b23b6fc72e7f06a7d2c7ccdcc

Spaqin: the error value generated in kern_hwreq is immediately discarded anyway so there it's unnecessary; for the rest (e.g. si5324) it's error mapped only to pinpoint the cause (useful for debugging if si5324 or another i2c chip fails)

SimonRenblad: "One should always prefer implementing From over [Into](https://doc.rust-lang.org/std/convert/trait.Into.html) because implementing From automatically provides one with an implementation of [Into](https://doc.rust-lang.org/std/convert/trait.Into.html) thanks to the blanket implementation in the standard library." from the docs

Issue #2697, opened 2025-03-12
Support running DMA from user provided buffer


yuyichao:
# ARTIQ Feature Request

## Problem this request addresses

The host contains all the information needed to generate the content of a DMA trace. This allows the host to do all the heavy lifting while achieving maximum output throughput on the kernel with zero (re)compilation time. However, there's no builtin tools to run such a trace in a kernel.

## Describe the solution you'd like

I think it should be acceptable to keep this as an advanced feature and requires the user to use internal syscall's and keep up to sync with the internal dma buffer format. However, there are a few issues that make this currently difficult to do. Below are the few items that can make this significantly easier to support.

1. Provide a function to flush the proper cache.

    Currently I did this with a LLVM function doing volatile read in a loop, mimicking the behavior of `flush_l2_cache`. This should be as simple as exposing an api from the firmware.

2. Provide a way to get a pointer from a list/ndarray/bytes/bytearray.

    I don't know if there's a way to add a generic version but a type specific version is really easy to codegen in llvm. This should be doable as a syscall as well.

3. Provide a way for the user to allocate a properly aligned buffer.

    FWIW, I don't really understand why DMA buffer starting address needs to be 64 bytes aligned. The sizes of the entries in the buffer are not multiple of 64 (in fact it is always a odd number) so the DMA engine is clearly able to split out an instruction from an unaligned address (though I guess the read may be done with 64byte block size?). In any case, if the alignment requirement cannot be removed, this can probably be fixed with a data type with larger alignment. (Currently dealt with in my code by flipping max_target_alignment when emitting the rpc call I used to get the data from the host). Fixing this would probably need the biggest change in the compiler. OTOH, manual realignment of the data is also possible with a syscall function.

## Additional context

I have implemented all the features above in our experiment using very hacky monkey patches of the artiq compiler. It runs the dma buffer from the kernel stack memory just fine and improves the output throughput by ~10-20% on kasli compared to a custom serialization format.

Supporting this for subkernel would be nice too. However, I'm still on version 7 so I don't have much idea about the available ways to transfer data to the subkernels.

occheung:
> FWIW, I don't really understand why DMA buffer starting address needs to be 64 bytes aligned.

I believe it is to align with memory bus width. IIRC the width is 64 bytes on KC705 and 16 bytes on Kasli.
You might be able to reduce the alignment suggested by the firmware to these numbers.

I think the DMA engine will read garbage if the addresses does not align with memory bus width.

[...]

PR #2704, opened 2025-03-17
Support unaligned DMA buffer address

yuyichao:
# ARTIQ Pull Request

## Description of Changes

Split the base address into an aligned part and an offset. Add one cycle to the start of the DMA playback to skip ahead to the real start of the playback.
This makes it significantly easier to run DMA from a user provided buffer. (All the rest are easy compiler stuff).

Currently this is not tested at all, hence the draft status (I'm still figuring out how to setup the proper environment...) and I'm not even 100% sure if I'm using the correct syntax. However, I do want to post this to get some feedback. The main reasons I went for this rather than adding support to the compiler are that

1. I think this can be supported with minimum to no overhead.
2. This can actually make recording faster since the firmware doesn't have to realign the buffer anymore (not implemented in this PR yet).

### Related Issue

Ref https://github.com/m-labs/artiq/issues/2697.

## Type of Changes

|   | Type |

### All Pull Requests

- [ ] Close/update issues.

### Code Changes

### Git Logistics

  ```
  
  ```

### Licensing


yuyichao:
> Why not just allocate an aligned buffer in the first place?

Well, that was my question when I looked at the firmware code. I just figured that either no one care about that performance or that there's no easy way to do so in rust.

> How so?

The direct answer to this question is that this solves one of the three requirements I've listed in https://github.com/m-labs/artiq/issues/2697 . Since it reduce the number of missing pieces from 3 to 2 this change obviously made it easier.

But I assume you are asking about comparing to other approaches. And here's a comparison for all the approaches/aspects I've considered.

First of all, I'm assuming the alignment requirement can be removed without significant performance hit so I'm only considering the software/firmware side. With that out of the way, the usecases I've considered are,

1. Normal user of rpc
2. Normal user of array allocated on kernel
3. Normal use of DMA
4. Use rpc result for DMA
5. Use array allocated on kernel for DMA

The first three are mainly to make sure there's no regression and the latter two are usecases I'd like to support (though I personally only need 4 for now). Now comparing the approaches with or without the alignment requirements.

* If the DMA alignment requirement is to be kept, the main question is how to make 4 and 5 works.

    * The solution that does not require any compiler change (in terms of alignment) is to over allocate the buffer and realign the pointer. This is basically how DMA buffers is allocated now normally. However, I feel like this is incredibly stupid since the LLVM stack allocation code is clearly capable to allocate a property aligned buffer and the buffer doesn't even need to be resized here.
    * Now if we require the allocation to be properly aligned from the get go, AFAICT, the only solution that does not involve adding completely new metadata to the compiler is to create a type with 64bytes alignment (or bump max alignment to 64 for a lesser version that only support rpc). However, this would also mean that all dynamically sized rpc return allocation would be 64bytes aligned (again, without deeper change to the compiler) which could be a significant regression for people that returns a list of short arrays from RPC.
    * Alternatively, if call/allocation site specific metadata can be added to make sure only specific rpc/array allocation gets 64 bytes alignment that should make sure there's no regression in normal code.

* Now if the DMA alignment requirement is removed, then everything becomes significantly easier on the software side.

    * There's no need for additional alignment, so no compiler changes necessary. No need for realignment copy. No need for site-specific metadata and no need for unnecessary over-allocation.
    * In addition, DMA buffers (either the normal ones or user created ones from RPC or normal array allocation) can be stored more compactly, saving some memory for short traces.

yuyichao:
> Obviously it can be done: allocate a buffer with extra size (alignment-1), then discard the beginning.
> For a dynamic buffer like Vec then of course the reallocation has to be implemented manually perhaps with the principle above.

Sure. I did not write the code to do the extra realignment copy. I'm only commenting based on what you currently do and I'll take your word for what could be fixed.

> I don't see any valid reason to deal with misalignment in gateware. It wastes FPGA resources and potentially makes timing closure more difficult.

So how much more resource would adding one more state to the FSM cost?

> When receiving a host-compiled DMA buffer, the device could simply allocate an aligned buffer to store the data coming from the network.

I'm certainly fine with supporting this. However, this must be done without bumping the alignment for all rpc dynamic allocation. Ideally it should also support kernel side allocation as well though I don't need it so I don't care about it very much. If you have something in mind that's easier than touching gateware and can be supported in both the current compiler version as well as all future version I'll probably be fine to implement that.

PR #2724, opened 2025-04-16
Fold experiments qt6

mmalczak:
# ARTIQ Pull Request

## Description of Changes

Refactor widgets creation - move to smaller methods.
Split layout into always visible layout and foldable layout with scheduler settings.
Adjust sizes and resizing behavior.
Save/restore state of the layout.

## Type of Changes

|   | Type |

### All Pull Requests

### Code Changes

### Git Logistics

  ```
  
  ```

### Licensing

### DIFF
--- a/artiq/dashboard/experiments.py
+++ b/artiq/dashboard/experiments.py
@@ -126,9 +126,9 @@ class _ExperimentDock(QtWidgets.QMdiSubWindow):
         self.options = manager.get_submission_options(expurl)
         self.hdf5_load_directory = os.path.expanduser("~")
 
-        master_layout = QtWidgets.QVBoxLayout()
+        self.master_layout = QtWidgets.QVBoxLayout()
         self._create_argument_editor()
-        master_layout.addWidget(self.argeditor)
+        self.master_layout.addWidget(self.argeditor)
 
         # Create a toggle button that will collapse/expand the options.
         self.fold_toggle = QtWidgets.QToolButton(text="Collapse scheduler settings",
@@ -139,7 +139,7 @@ class _ExperimentDock(QtWidgets.QMdiSubWindow):
         self.fold_toggle.setToolTip("Collapse/Expand scheduler settings")
         self.fold_toggle.setArrowType(QtCore.Qt.ArrowType.DownArrow)
         self.fold_toggle.clicked.connect(self.on_fold_toggle)
-        master_layout.addWidget(self.fold_toggle)
+        self.master_layout.addWidget(self.fold_toggle)
 
         # Create a container widget (with a grid layout) for all the foldable
         # options.
@@ -148,7 +148,7 @@ class _ExperimentDock(QtWidgets.QMdiSubWindow):
         self.foldable_layout.setSpacing(5)
         self.foldable_layout.setContentsMargins(5, 5, 5, 5)
         self.foldable_container.setLayout(self.foldable_layout)
-        master_layout.addWidget(self.foldable_container)
+        self.master_layout.addWidget(self.foldable_container)
 
         # Create a container widget (with a horizontal layout) for
         # always-visible buttons.
@@ -157,11 +157,11 @@ class _ExperimentDock(QtWidgets.QMdiSubWindow):
         self.always_visible_layout.setSpacing(5)
         self.always_visible_layout.setContentsMargins(5, 5, 5, 5)
         self.always_visible_container.setLayout(self.always_visible_layout)
-        master_layout.addWidget(self.always_visible_container)
+        self.master_layout.addWidget(self.always_visible_container)
 
         # Set the master layout on the top widget.
         top_widget = QtWidgets.QWidget()
-        top_widget.setLayout(master_layout)
+        top_widget.setLayout(self.master_layout)
         self.setWidget(top_widget)
 
         # --- Create the various widget groups ---
@@ -388,7 +388,7 @@ class _ExperimentDock(QtWidgets.QMdiSubWindow):
 
         editor_class = self.manager.get_argument_editor_class(self.expurl)
         self.argeditor = editor_class(self.manager, self, self.expurl)
-        self.layout.addWidget(self.argeditor, 0, 0, 1, 5)
+        self.master_layout.insertWidget(0, self.argeditor)
         self.argeditor.restore_state(argeditor_state)
         self.apply_window_color()
 


SimonRenblad:
Thank you for the PR! Can you split it up into smaller ones / push the refactor separately?

artiq/dashboard/experiments.py
```diff
@@ -115,47 +111,113 @@ def drawComplexControl(self, control, option, painter, widget=None):
 class _ExperimentDock(QtWidgets.QMdiSubWindow):
     sigClosed = QtCore.pyqtSignal()
 
+
     def __init__(self, manager, expurl):
-        QtWidgets.QMdiSubWindow.__init__(self)
```
SimonRenblad, line 119: Why? We do not use `super` in most other qt subclasses.

artiq/dashboard/experiments.py
```diff
@@ -115,50 +115,94 @@ def drawComplexControl(self, control, option, painter, widget=None):
...
         self._create_repo_rev_widgets()
+        # Place the submit and termination buttons into the always-visible
+        # container.
         self._create_submit_widgets()
         self._create_reqterm_widgets()
 
+        self.on_fold_toggle()
+
+    def on_fold_toggle(self):
+        """Toggle the visibility of the options."""
```
SimonRenblad, line 184: remove docstring

artiq/dashboard/experiments.py
```diff
@@ -188,10 +184,17 @@ def on_fold_toggle(self):
             self.foldable_container.show()
             self.fold_toggle.setText("Collapse scheduler settings")
             self.fold_toggle.setArrowType(QtCore.Qt.ArrowType.DownArrow)
+            if self.width() < 550:
+                ideal_width = self.sizeHint().width()
+                self.resize(ideal_width, self.height())
+            if self.height() < 285:
+                self.resize(self.width(), 285)
```
SimonRenblad, line 193: Are these resizing changes necessary to get the folding behavior? I am skeptical of manual resizing with magic values unless really necessary.

artiq/dashboard/experiments.py
```diff
@@ -47,20 +47,16 @@ def __init__(self, manager, dock, expurl):
...
-        buttons = LayoutWidget()
-        buttons.addWidget(recompute_arguments, 1, 1)
-        buttons.addWidget(load_hdf5, 1, 2)
-        buttons.layout.setColumnStretch(0, 1)
-        buttons.layout.setColumnStretch(1, 0)
-        buttons.layout.setColumnStretch(2, 0)
-        buttons.layout.setColumnStretch(3, 1)
-        self.setItemWidget(self.bottom_item, 1, buttons)
+        self.setItemWidget(self.bottom_item, 0, load_hdf5)
+        self.setItemWidget(self.bottom_item, 1, recompute_arguments)
```
SimonRenblad, line 59: nit: I personally think this change looks worse than before. Also is it necessary?

SimonRenblad (changes requested):
Folding experiment settings:
I agree with the premise that we can potentially hide the scheduler settings, although I am not sure about this approach. Even when collapsed, the "Expand scheduler settings" button is still taking up real estate. I feel like this could be resolved through other means such as adding a checkbox menu item to the *existing* context menu on the experiment dock. This would also be more in line with how docks in the rest of the application can be closed. 

Screenshot From 2025-04-17 15-46-18

Screenshot From 2025-04-17 15-49-03

Resizing changes made to experiment dock/scheduler settings:
Personally, I don't think the resizing changes made here look good enough to justify the change, but I suppose that's up to debate. Can it be a separate PR?

artiq/gui/entries.py
```diff
@@ -23,7 +23,7 @@ def __init__(self):
             set_resize_mode = self.header().setSectionResizeMode
         else:
             set_resize_mode = self.header().setResizeMode
-        set_resize_mode(0, QtWidgets.QHeaderView.ResizeMode.ResizeToContents)
+        set_resize_mode(0, QtWidgets.QHeaderView.ResizeMode.Stretch)
```
SimonRenblad, line 26: This was not previously the case because it gave less width to the argument title (typically a short label) and more width to the argument widget (typically multiple components, sliders etc). Why the change?

marmeladapk:
The driving idea behind this PR was to make smaller windows possible so that more experiments fit in a window. If you have an experiment that takes little to no arguments, then scheduler settings take majority of space and limit how small the window can be. That's why layout and sizing changes are in a single PR.

Corresponding PR in which you can trace the origin of these changes in relation to problems that I found:
https://github.com/elhep/artiq/pull/8

> Even when collapsed, the "Expand scheduler settings" button is still taking up real estate. I feel like this could be resolved through other means. For example, by adding a check box menu item to the existing context menu on the experiment dock. See also how docks in the rest of the application can be closed through a context menu.

Yes, it's still taking up space, however I wanted this button to be still visible because:

- I feel that this state should be per experiment window not globally, so that you can still expand these settings selectively on experiment that use e.g. due date.
- In general I feel that context menus are not easily discoverable. 
- In a future PR we want to add a highlight to indicate that there are non-default values in scheduler settings.

This way if the scheduler doesn't behave the way you were expecting it to then you at least have a visible indicator that there are some additional settings. This may help also in a shared environment, where someone else might have made the changes.

artiq/dashboard/experiments.py
marmeladapk: IIRC this was to prevent issues when expanding settings on small window.
E.g. as shown here (last picture):
https://github.com/elhep/artiq/pull/8#pullrequestreview-2608665555

I agree that magic values are not the best way to do it, however otherwise we may need to calculate the height based on all visible elements.

artiq/dashboard/experiments.py
marmeladapk: Yes, it looks a bit worse, however it was a simple way to retain the visibility of these buttons when the window is narrow. Screenshots from the original PR:

> Minimum width is too low when scheduler settings are hidden -
> image
> image
> maybe make a single large row for recompute and load buttons
> maybe base minimum width on parameters - this might be tricky if there are groups

artiq/dashboard/experiments.py
SimonRenblad: If it can be easily resolved by the user resizing the window, I personally think it's acceptable that it looks a bit weird when the windows are small. This is common for many GUI applications as a trade-off for being flexible. It also would not block getting the folding behavior (so it could go in a later PR).

artiq/dashboard/experiments.py
SimonRenblad: Was this always the case, or introduced by your layout changes? If it was always the case, I still maintain that it can go in a separate PR as it addresses a separate concern (not related to folding behavior).

SimonRenblad:
> The driving idea behind this PR was to make smaller windows possible so that more experiments fit in a window. If you have an >experiment that takes little to no arguments, then scheduler settings take majority of space and limit how small the window can >be. That's why layout and sizing changes are in a single PR.

At least the widget refactor should be separate. I understand that it's the same idea, but sizing changes can be implemented without folding and vice versa. Minimal and atomic changes are preferred.

> Yes, it's still taking up space, however I wanted this button to be still visible because:
> 
>     * I feel that this state should be per experiment window not globally, so that you can still expand these settings selectively on experiment that use e.g. due date.

A context menu can be per experiment window, and it is what I am suggesting.

> 
>     * In general I feel that context menus are not easily discoverable.

A context menu is:
+ easier to implement
+ minimally invasive: does not alter the existing layout
+ makes copying this behavior to other areas of the application significantly easier
+ can be easily combined with more discoverable approaches (such as adding a toolbar for each window)
+ is IMO sufficiently discoverable considering the niche / optional use case presented here

> 
>     * In a future PR we want to add a highlight to indicate that there are non-default values in scheduler settings.

This can be done without tying it to a button.

I would also accept:
+ A toolbar per experiment dock with a "View" menu, commonplace in many GUI apps and could be extended to hiding the arguments area (if the user has none and doesn't care about it).
+ Using a tree widget or other Qt widget that has built in collapsible children.

My point is that a PR concerned with looking better / getting more functionality when height is compressed would want to consider a way to achieve this without adding another fixed height component into the layout.

SimonRenblad:
> If there is no fixed height, would the change be more acceptable for you?

That is one factor to consider, and a point in favor of the context menu. I only mentioned it specifically since the motivation of the PR is to get more functionality out of a small window. I don't think that having some kind of "variable height" button is a good substitute.

> It resizes properly when you eg. move the window. It must be missing some kind of repaint somewhere.

Can it be solved with something like an `update` call?

It is preferable if the solution can be easily replicated to other areas of the application, by encapsulating the collapsible logic in a separate class. 

If you are insistent on going the route of the button, here are a few other suggestions:
+ Drop the "Expand/Collapse". The title of the section with the arrow is sufficient to communicate the functionality
+ Styling: Hide the border of the push button and add a horizontal spacer to the right of it, as is common for this kind of component.

artiq/dashboard/experiments.py
marmeladapk: It was introduced - before it was impossible due to minimum width of the window being much bigger.

dnadlinger:
For a QSplitter/hidden section approach, it would also be possible to relcaim further space by not showing the Submit/Request termination buttons, as many expert users use the keyboard shortcuts for this (Ctrl+Enter, Ctrl+Backspace).

PR #2742, opened 2025-04-29
targets/kasli: make the number of Grabber ROI engines configurable

airwoodix:
# ARTIQ Pull Request

## Description of Changes

Add a property to the Kasli coredevice schema to configure the number of ROI engines for Grabber peripherals:

```json
        {
            "type": "grabber",
            "ports": [ 1 ],
            "roi_engine_count": 32
        }
```

The number of ROI engines is limited to 32 such that calls to [`Grabber.gate_roi`](https://github.com/m-labs/artiq/blob/8156151589019dd13770272e2f575234336725ea/artiq/coredevice/grabber.py#L62) can address every ROI engine (the `data` argument in `rtio_output` is a 32-bit integer).

Builds with 16 and 32 ROI engines are known to work. It is not clear whether the number of ROI engines has other restrictions, e.g. a power of two. The latter could not be enforced in the JSON schema (similarly to the number of SED lanes).

The change is fully backwards-compatible, with the default value set to 16 ROI engines, as before this patch.

## Type of Changes

|   | Type |
|   | :bug: Bug fix  |
|   | :hammer: Refactoring  |
|   | :scroll: Docs |

### All Pull Requests

- [ ] Close/update issues.

### Code Changes

### Git Logistics

  ```
  
  ```

### Licensing


artiq/gateware/eem.py
```diff
@@ -502,12 +502,12 @@ def io(eem, eem_aux, iostandard):
 
     @classmethod
     def add_std(cls, target, eem, eem_aux=None, eem_aux2=None, ttl_out_cls=None,
-            iostandard=default_iostandard):
+            iostandard=default_iostandard, roi_engine_count=16):
```
MorganTL, line 505: Since the json schema already include 16 in the default, no need to include 16 here again.

airwoodix: Would it be ok to break backwards compatibility for that? In principle, there could be other users of `artiq.gateware.eem.Grabber.add_std` than through the Kasli target builder.

Side question: would you be happy with having `roi_engine_count` as keyword-only argument?

```suggestion
            iostandard=default_iostandard, *, roi_engine_count):
```

MorganTL: The `add_std` is not a user facing api so it's okay.
And no, as other peripherals' `add_std` don't use keyword-only argument.

airwoodix: If the change is backwards compatible, would you consider backporting the change to `release-8`?

(The PHY gateware component also has 16 as default value for its `roi_engine_count` argument. If the default should be deduplicated, that one should be removed too.)

airwoodix: If I remove the default value and don't use a keyword-only argument, the `roi_engine_count` argument needs to be latest after `eem`, which will be a much bigger and less intuitive change. Which of the two variants would you prefer?

```python
def add_std(self, target, roi_engine_count, eem, eem_aux=None, ...):
    ...
```

or

```python
def add_std(self, target, eem, roi_engine_count, eem_aux=None, ...):
    ...
```

> The general guideline is to make the code as good as possible without caring about backward compatibility on the beta branch

What about backports? This patch would be really nice to have on `release-8`. It isn't new functionality, since it only exposes a setting that was already there.

The keyword-only argument would be a better option than a compulsory positional-or-keyword argument, since it reduces errors at the call site, especially with that many arguments.

RELEASE_NOTES.rst
```diff
@@ -17,6 +17,8 @@ ARTIQ-9 (Unreleased)
    - Zotino monitoring now displays the values in volts.
    - Support for the ultra-low-cost EBAZ4205 Zynq-7000 control card, with core device driver
      for the AD9834 DDS, tested with the ZonRi Technology Co., Ltd. AD9834-Module.
+   - Configurable number of Grabber ROI engines through the ``roi_engine_count``
+     property in the hardware description JSON file.
```
architeuthidae, line 21: Prefer "system description", hardware description isn't inaccurate but consistency with the docs is nice.

airwoodix: Changed, thanks!

PR #2750, opened 2025-05-09
artiq_flash: Add action to erase storage flash sectors

linuswck:
# ARTIQ Pull Request

## Description of Changes
Currently, if you want to re-calibrate the DRTIO-over-EEM interface of a (non-kasli and non-kasli-soc) board, you can only do so by erasing all the flash sectors and then re-flashing all gateware, bootloader, storage and firmware. If you have changed the board's gateware or any clock related cables, you need to do so to avoid data corruption. This PR adds an action to erase only the storage flash sectors for DRTIO-over-EEM interface re-calibration.

This new feature will be useful for both development and crate assembly, which are prone to frequent gateware changes or cable changes.

This new feature is tested on the phaser. 
1. `artiq_flash erase_storage -t phaser` to erase the storage flash sectors.
2. `artiq_flash start` to reboot phaser.
3. The log shows that DRTIO-over-EEM interface is calibrated at bootup.

## Type of Changes

|   | Type |

### Licensing

### DIFF
--- a/artiq/frontend/artiq_flash.py
+++ b/artiq/frontend/artiq_flash.py
@@ -30,6 +30,7 @@ Valid actions:
     * firmware: write firmware to flash
     * load: load main gateware bitstream into device (volatile but fast)
     * erase: erase flash memory
+    * erase_storage: erase the storage flash sectors
     * start: trigger the target to (re)load its gateware bitstream from flash.
       If your core device is reachable by network, prefer 'artiq_coremgmt reboot'. 
 
@@ -146,6 +147,16 @@ class Programmer:
                      "flash probe {bankname}",
                      "flash erase_sector {bankname} 0 last",
                      bankname=bankname)
+
+    def erase_storage(self, bankname, start_address, end_address):
+        self.load_proxy()
+        size = end_address - start_address
+        add_commands(self._script,
+            "flash probe {bankname}",
+            "flash erase_sector {bankname} {firstsector} {lastsector}",
+            bankname=bankname,
+            firstsector=start_address // self._sector_size,
+            lastsector=(start_address + size - 1) // self._sector_size)
 
     def load(self, bitfile, pld):
         os.stat(bitfile) # check for existence
@@ -329,6 +340,8 @@ def main():
             programmer.start()
         elif action == "erase":
             programmer.erase_flash("spi0")
+        elif action == "erase_storage":
+            programmer.erase_storage(*config["storage"], config["firmware"][1])
         else:
             raise ValueError("invalid action", action)
 
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -16,6 +16,7 @@ ARTIQ-9 (Unreleased)
    - Fastino monitoring with Moninj.
    - Zotino monitoring now displays the values in volts.
    - artiq_flash can now flash Phaser through a Digilent HS2 Programming cable.
+   - artiq_flash can now erase storage flash sectors specifically.
    - Support for the ultra-low-cost EBAZ4205 Zynq-7000 control card, with core device driver
      for the AD9834 DDS, tested with the ZonRi Technology Co., Ltd. AD9834-Module.
    - Configurable number of Grabber ROI engines through the ``roi_engine_count``


artiq/frontend/artiq_flash.py
```diff
@@ -292,13 +318,26 @@ def main():
...
-    if not args.action:
-        args.action = "gateware bootloader firmware start".split()
-    needs_artifacts = any(
-        action in args.action
-        for action in ["gateware", "bootloader", "firmware", "load"])
-    if needs_artifacts and args.dir is None:
-        raise ValueError("the directory containing the binaries need to be specified using -d.")
+    if not args.region:
+        if args.action =="write":
+            args.region = "gateware bootloader firmware".split()
```
MorganTL, line 323: refactor to `["gateware", "bootloader", "firmware"]` instead of `split()`

artiq/frontend/artiq_flash.py
```diff
@@ -292,13 +318,26 @@ def main():
...
+        elif args.action == "erase":
+            args.region = "gateware bootloader storage firmware".split()
```
MorganTL, line 325: see above for `.split()` comment

artiq/frontend/artiq_flash.py
```diff
@@ -292,13 +318,26 @@ def main():
...
+
+    needs_storage_artifacts, needs_dir_artifacts = False, False
+    if args.action == "write":
+        needs_dir_artifacts = any(
+            region in args.region
+            for region in ["gateware", "bootloader", "firmware"])
+        needs_storage_artifacts = any(
+            region in args.region
+            for region in ["storage"])
```
MorganTL, line 334: `"storage" in args.region`?

artiq/frontend/artiq_flash.py
```diff
@@ -309,28 +348,27 @@ def main():
...
-            raise ValueError("invalid action", action)
+    if args.action == "write":
+        for region in args.region:
+            if region == "firmware":
+                bin = fetch_bin(binary_dir, ["satman", "runtime"], args.srcbuild)
+            elif region == "storage":
+                bin = args.storage
+            else:
+                bin = fetch_bin(binary_dir, [region], args.srcbuild)
+            programmer.write_binary(*config[region], bin)
```
MorganTL, line 359: Is it intended to accept duplicate of region's name? If I use `artiq_flash write firmware firmware firmware`, it will flash firmware 3 times.

artiq/frontend/artiq_flash.py
```diff
@@ -309,28 +348,27 @@ def main():
...
+        programmer.start()
+    elif args.action == "erase":
+        flash_regions = {k: v for k, v in config.items() if k not in ["programmer"]}
```
MorganTL, line 362: I don't think it's necessary to remove the `"programmer"` key in the config dictionary but the variable name need to be change to config or something similar. If you really want to keep the flash_region, use `.pop()` instead

MorganTL (commented):
imo the `gateware` `firmware`, `bootloader` and `storage` should be options i.e. `--gateware`, `--firmware`, `--bootloader`, `--storage` with shortcuts like `-G`, `-F`, `-B`, `-S` So it's more ergonomic for user to just type `artiq_flash write -GFB` and the list of `args.region`  don't need to be sanitized to prevent duplicate flashing.

artiq/frontend/artiq_flash.py
```diff
@@ -140,12 +147,31 @@ def add_flash_bank(self, name, tap, index):
...
+                "flash erase_sector {bankname} {firstsector} {lastsector}",
+                bankname=start_bankname,
+                firstsector=start_address // self._sector_size,
+                lastsector=(start_address + size - 1) // self._sector_size)
+        else:
+            add_commands(self._script,
+                "flash probe {bankname}",
+                "flash erase_sector {bankname} {firstsector} last",
+                bankname=start_bankname,
+                firstsector=start_address // self._sector_size)
```
SimonRenblad, line 174: You could probably combine this into one command and reduce the size of the `if` statement here by allowing `lastsector` to also take the value "last".

artiq/frontend/artiq_flash.py
```diff
@@ -309,28 +348,27 @@ def main():
...
+            elif region == "storage":
+                bin = args.storage
+            else:
+                bin = fetch_bin(binary_dir, [region], args.srcbuild)
+            programmer.write_binary(*config[region], bin)
+        programmer.start()
+    elif args.action == "erase":
+        flash_regions = {k: v for k, v in config.items() if k not in ["programmer"]}
+        for region in args.region:
+            programmer.erase(region, flash_regions)
```
SimonRenblad, line 364: IMO might as well pass in `regions`  in plural here since you already modify `erase`/`erase_flash` heavily in this PR. Could allow you to reduce the number of commands added since continuous regions could be erased together. Also probably don't need to `load_proxy` `len(args.region)` times (unnecessary `os.stat` calls).

SimonRenblad (changes requested):
See comments.

SimonRenblad:
> imo the gateware firmware, bootloader and storage should be options i.e. --gateware, --firmware, --bootloader

The current interface allows for multiple actions, where the order matters. Using options for each action (a positional argument) does not seem feasible. I suggested to @linuswck to take a page from the [artiq_client handling of positional arguments as `KEY=VALUE`](https://github.com/m-labs/artiq/blob/b4d88f009b384d910397de1b0e9b2c4b0a01f12c/artiq/frontend/artiq_client.py#L77-L78):
```
artiq_flash erase=gateware,firmware,bootloader
```
whether it uses `=` or another delim I don't have a strong opinion. Shorthand `erase=gfb` and the like could also be supported.

artiq/frontend/artiq_flash.py
```diff
@@ -292,13 +317,47 @@ def main():
...
+                    elif c == "s":
+                        regions.append("storage")
+                    elif c == "f":
+                        regions.append("firmware")
+                    else:
+                        raise ValueError(f"unrecognized flash region: {regions[0]}")
+                regions = regions[1:]
+        else:
+            for region in regions:
+                if region not in ["gateware", "bootloader", "storage", "firmware"]:
```
MorganTL, line 344: `artiq_flash write=g,b,s,f` or `artiq_flash write=g,b` won't work. It's better to implement this as regex.

artiq/frontend/artiq_flash.py
```diff
@@ -292,13 +317,47 @@ def main():
...
+                    raise ValueError(f"unrecognized flash region: {region}")
+        actions.append([action, set(regions)])
```
MorganTL, line 346: The `actions` need to be renamed, it's not exactly a list of `action` afterall.

MorganTL (commented):
see comment

artiq/frontend/artiq_flash.py
```diff
@@ -140,12 +148,29 @@ def add_flash_bank(self, name, tap, index):
...
+
+            if next_region is not None:
+                _, end_address = config[next_region]
+                size = end_address - start_address
+                lastsector = (start_address + size - 1) // self._sector_size
+            else:
+                lastsector = "last"
+
+            add_commands(self._script,
+                "flash probe {bankname}",
```
SimonRenblad, line 170: Move outside the loop?

artiq/frontend/artiq_flash.py
Spaqin: I'd even say that 1 and ``n`` cases should be handled together.

artiq/frontend/artiq_flash.py
```diff
@@ -140,12 +148,29 @@ def add_flash_bank(self, name, tap, index):
...
             tap=tap, name=name, ir=0x02 + index)
 
-    def erase_flash(self, bankname):
+    def erase(self, regions, config):
         self.load_proxy()
-        add_commands(self._script,
-                     "flash probe {bankname}",
-                     "flash erase_sector {bankname} 0 last",
-                     bankname=bankname)
+        for region in regions:
```
SimonRenblad, line 153: Might be cleaner to loop through each region in the `config` and check if it should be erased or not. That way you could minimize the number of `flash erase_sector` calls (as mentioned in a previous comment) with relative ease. I'd guess any extra time spent doing this would be dwarfed by the JTAG comms anyways?

SimonRenblad (approved):

PR #2782, opened 2025-06-18
ad9910: Add set_profile convienience method

atseml:
Adds a convenience method for coredevice's `AD9910`, since in PROTO_REV 9 the profile pins of the individual channels are configurable now.

## Testing

Basic test code:
```python
        self.dds0.set(frequency=5*MHz, amplitude=1.0, profile=7)
        self.dds0.set(frequency=5*MHz, amplitude=0.5, profile=0)

        while True:
            delay(1.0)
            self.dds0.set_profile(0)
            delay(1.0)
            self.dds0.set_profile(7)
````

### DIFF
--- a/artiq/coredevice/ad9910.py
+++ b/artiq/coredevice/ad9910.py
@@ -1002,6 +1002,20 @@ class AD9910:
         self.cpld.cfg_att_en(self.chip_select - 4, state)
 
     @kernel
+    def set_profile(self, profile: TInt32):
+        """Set the PROFILE pins.
+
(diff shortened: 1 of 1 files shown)


atseml:
Whups, just found out this silently sets profiles to all DDS channels on ProtoRev 8.

atseml:
> Whups, just found out this silently sets profiles to all DDS channels on ProtoRev 8.

Force-pushed, warning added in the method docstring to inform users of the potential pitfall.

PR #2785, opened 2025-06-20
urukul: add slack for SPI startup

occheung:
## Background

Currently, there is only a 8 ns delay between enabling the SPI interface and transfer data via SPI.
The iCE40 FPGA may miss the first few clock edges.

This PR makes sure that the SPI transfer meet the differential and common mode voltage specifications on iCE40. ~~This change is added to PROTO_REV 9 specifically since iCE40 only supports this revision.~~ This patch is added to both rev 8 and 9 as both CPLD and iCE40 variants suffer from this phenomenon in various degree.
Experimentally, an additional delay as little as 80 ns is sufficient to read back the `proto_rev` field correctly.

See [urukul-pld PR](https://git.m-labs.hk/M-Labs/urukul-pld/pulls/1#issuecomment-15163).

Note: This issue was not observed on the Xilinx CPLD variant.

## Tests

iCE40 Urukul can read back `proto_rev` = 9 in the `sta_read()` call during `init()`.

### DIFF
--- a/artiq/coredevice/urukul.py
+++ b/artiq/coredevice/urukul.py
@@ -306,6 +306,11 @@ class ProtoRev8(CPLDVersion):
         cfg = cpld.cfg_reg
         # Don't pulse MASTER_RESET (m-labs/artiq#940)
         cpld.cfg_reg = cfg | (0 << ProtoRev8.CFG_RST) | (1 << ProtoRev8.CFG_IO_RST)
+        # Preemptively enable the SPI. Voltages of both common mode and
+        # differential are too small initially.
+        # This dummy config value is similar to the coming SPI config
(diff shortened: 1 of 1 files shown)


occheung:
> But did you see incorrect signals on the scope i.e. out of LVDS specs?

The same SCLK trace look like this in Xilinx CPLD variant. R7B was probed against GND.
SDS2504X Plus_JPG_1

LVDS transceiver specifies >135 mV differential voltage for HIGH input, <65 mV for LOW. There are no common mode related spec. This waveform meets the spec.

16 ns periodic square wave is also observed from the transceiver output.

> Does this behavior match what is expected from the Xilinx datasheets and user guides?

Xilinx datasheets only specifies the delay from T pin to IOB when it is in high-Z mode. However, it does not specify the time needed to satisfy the LVDS DC specs, after transitioning from 3-state.

PR #2789, opened 2025-06-27
Add support for passing flags to extern decorators

Oli-Ar:
## Feature
Allow flags to be passed to an extern decorator. Additionally add the `allow-external-alloc` flag to `cache_get` function, to allow for it to return a list type with [change 636](https://git.m-labs.hk/M-Labs/nac3/pulls/636) to nac3.

✓ | ✨ New feature

## Testing
`nix build` unit tests here and in nac3 repo. Failing test in nac3, due to `cache_get` list return value, now passes.

## Licensing

### DIFF
--- a/artiq/language/core.py
+++ b/artiq/language/core.py
@@ -68,10 +68,14 @@ def _register_class(cls):
     _registered_classes[cls] = module
 
 
-def extern(function):
+def extern(arg=None, flags={}):
     """Decorates a function declaration defined by the core device runtime."""
-    _register_function(function)
-    return function
+    if arg is None:
+        def inner_decorator(function):
+            _register_function(function)
+            return extern(function, flags)
+        return inner_decorator
+    return arg 
 
 
 def kernel(function_or_method):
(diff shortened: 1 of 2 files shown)


artiq/language/core.py
```diff
@@ -68,10 +68,14 @@ def _register_class(cls):
...
 
 
-def extern(function):
+def extern(arg=None, flags={}):
     """Decorates a function declaration defined by the core device runtime."""
-    _register_function(function)
-    return function
+    if arg is None:
+        def inner_decorator(function):
+            _register_function(function)
```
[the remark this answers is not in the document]
Oli-Ar: Will do

artiq/language/core.py
```diff
@@ -68,10 +68,14 @@ def _register_class(cls):
     _registered_classes[cls] = module
 
 
-def extern(function):
+def extern(arg=None, flags={}):
```
ivan-shrimp, line 71: The default arg for `flags` is a single `dict` (not `set`) object shared across all invocations of `extern`, which is an anti-pattern; any changes to this object would appear in other irrelevant calls to `extern`.

It's not an issue here since we just discard the flags (and re-parse it only in nac3), but then we should probably just set `flags=None` for clarity? (same for `@rpc`)

ivan-shrimp: This is indeed harmless; just that it catches the eye immediately when reading the code.

PR #2790, opened 2025-06-30
README/docs: update m-labs.hk links

architeuthidae:
Following recent restructuring of website.


Spaqin:
I assume this will need to be done with ARTIQ 8 as well, quite straightforward, but how about release 7?

Issue #2762, opened 2025-05-28
dashboard/log: add timestamp column


jordens:
# ARTIQ Feature Request

## Problem this request addresses

Log message timestamps are useful information but are currently not shown in the log dock.

## Describe the solution you'd like

Show a timestamps as a column.
Potentially make visible columns configurable.
Potentially make that column show time deltas instead of timestamps where they are small (e.g. less than 10 seconds), like `dmesg` does.

PR #2795, opened 2025-07-02
gui: add timestamps column to log

HarryPoon000:
### Feature
Timestamps of log messages in log dock of dashboard
###Related issue
#2762

### DIFF
--- a/artiq/gui/log.py
+++ b/artiq/gui/log.py
@@ -181,13 +181,17 @@ class _Model(QtCore.QAbstractItemModel):
             if item.parent is self:
                 if column == 0:
                     return v[1]
-                else:
+                elif column == 1:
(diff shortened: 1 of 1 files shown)


artiq/gui/log.py
```diff
@@ -46,7 +46,7 @@ class _Model(QtCore.QAbstractItemModel):
     def __init__(self, palette):
         QtCore.QAbstractTableModel.__init__(self)
 
-        self.headers = ["Source", "Message"]
+        self.headers = ["Source", "Message", "Timestamp"]
```
HarryPoon000, line 49: Should the timestamp be put before the source?

HarryPoon000:
Yup

PR #2810, opened 2025-07-15
artiq_coremgmt: support config read to file

SimonRenblad:
Adds support for reading from the core device config to file with `artiq_coremgmt config read <key> -f <filename>`.

Currently `artiq_coremgmt` errors when reading files from the config (routing_table, idle_kernel etc.). This PR allows for  reading files directly into a file on the host, and could be used for recovering a routing table, idle kernel or startup kernel with `artiq_coremgmt`.  

`artiq_coremgmt config read <key>` works as before. Unfortunately this leads to some inconsistency with `artiq_coremgmt config write -f <key> <filename>` (note the position of `-f` in both). 

### Testing
+ Tested with Kasli_SoC HW, reading and writing binary files and strings alike works as expected. 
+ Untested with Kasli v2.

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -53,6 +53,7 @@ Breaking changes:
      require PYON v2: ensure sipyco v2 is used.
    - When PYON v2 decoding fails in the following contexts, decoding as PYON v1 is attempted.
      The support for PYON v1 decoding will be removed in a future ARTIQ release.
+
       + Dataset DB file
       + HDF5 results ``expid``. Note that existing files are never altered.
       + artiq_client set-dataset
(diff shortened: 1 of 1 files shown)


SimonRenblad:
Changes:
+ Support multiple reads in the same command (mirroring `write`):
```
> artiq_coremgmt config read -s key1 -s key2 -f key3 filename.bin
contents_of_key1
contents_of_key2
```
+ Removed empty string check, this is only relevant for writing an empty string to Kasli-SoC. It is otherwise a firmware error if the key cannot be found.
+ Fixed leftover indent error from RELEASE_NOTES, moved to breaking changes list

PR #2824, opened 2025-07-29
feat: add `nac3_{free, malloc}`

HTGAzureX1212:
Same as adding dynamic memory allocation support to ARTIQ/Zynq.


HTGAzureX1212:
Would `fence` instructions help?

PR #2826, opened 2025-08-01
GTP: Prohibit RX initialization before clock switch

occheung:
## Description
This PR disallows RX from initializing before switching sys clock. This is achieved by forcing the FSM to wait for TX to finish initialization.

## Motivating/Issue
Since the `GTPRXInit` is in `sys` clock domain, the clock switch interrupts the RX FSM through clock domain reset.
There is a chance that `rtio_rx` clock to be stuck at around 156.25 MHz after clock switch.

Normally, `rtio_rx` clock is expected to be switching in between one of the following:
- no signal, when under reset;
- 156.25 MHz, when DRP is being configured after RX reset (the divider in PMA is /4).
- 125 MHz, when initialized (such that the divider in PMA is /5);

When stuck at 156.25 MHz, RX link can never be acquired until reconfiguration.

This PR proposes putting RX after TX for these reasons:
1. TX init cannot start until the RX clock is established. Finishing TX init implies the clock is switched and `sys` clock is ready.
2. We can assume the PLL is locked (checked by TX) and not having to check it in RX as suggested by UG482.

### DIFF
--- a/artiq/gateware/drtio/transceiver/gtp_7series.py
+++ b/artiq/gateware/drtio/transceiver/gtp_7series.py
@@ -2,6 +2,7 @@ from functools import reduce
 from operator import or_
 
 from migen import *
+from migen.genlib.cdc import MultiReg
 from migen.genlib.resetsync import AsyncResetSynchronizer
 
 from misoc.cores.code_8b10b import Encoder, Decoder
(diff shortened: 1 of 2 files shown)


Spaqin:
Since it's a bugfix, I assume this should also be applied to release-8?

PR #2828, opened 2025-08-06
firmware: flush uart before reboot

SimonRenblad:
### Flush UART before rebooting with `artiq_coremgmt`

Needs https://github.com/m-labs/misoc/pull/173.

Delays until the UART tx fifo is empty. 

#### Testing
Runtime firmware builds, UART log flushes correctly on reboot. Passes HITL tests.
Satman firmware builds. DRTIO setups not tested.


jordens:
flush functionality in misoc uart would be generically useful.

PR #2841, opened 2025-08-19
log set_level: modify buffer log level only

dly04:
Fix the global log level at TRACE, modifying only buffer log level when calling `artiq_coremgmt log set_level`.
By replacing `set_max_level `being called after `artiq_coremgmt log set_level` by `set_buffer_log_level` (copied from artiq-zync firmware).
So that `artiq_coremgmt log set_level` won't affect UART log level.
Tested on Kasli v2.0.2.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -11,11 +11,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1756033593,
-        "narHash": "sha256-/Tb+dC4rAwcB2NVC7I2iiUSucl+LML/Fmqrkal/bhRU=",
+        "lastModified": 1756174750,
+        "narHash": "sha256-GSYJKbI5uLhB5X9b+N23vZtSFkStvlKlj/QoKL5+YLE=",
         "owner": "m-labs",
         "repo": "artiq-comtools",
-        "rev": "4fa7cda7894d5c4d59d75ccda788bf3e35ff3d40",
+        "rev": "3635c4b9b61cde5d8c5b068263bed46752e6779e",
         "type": "github"
       },
       "original": {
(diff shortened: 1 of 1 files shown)


Spaqin:
How about making a config log setting that would be persistent across reboots? I don't think most users need ``TRACE`` level from the get go (especially as extra logs pushed would slow things down), but it could be helpful for debugging at times.

dly04:
Here the `update_global_log_level` function is implemented and called in `set_uart_log_level` and `set_buffer_log_level`.

artiq/examples/kasli_shuttler/kasli_shuttler.json
```diff
@@ -1,19 +0,0 @@
-{
```
SimonRenblad, line 1: Did you mean to commit this? Looks unrelated to the PR.

artiq/firmware/liblogger_artiq/lib.rs
```diff
@@ -83,7 +88,28 @@ impl BufferLogger {
...
+
+    pub fn set_buffer_log_level(&self, max_level: LevelFilter) {
+        self.buffer_filter.set(max_level);
+        self.update_global_log_level()
+    }
+
+    pub fn update_global_log_level(&self){
+        let uart_level = self.uart_filter.get();
+        let buffer_level = self.buffer_filter.get();
+        let global_level =
```
SimonRenblad, line 107: Just use `core::cmp::max`?

SimonRenblad (commented):
Please clean up the commit history.

sbourdeauducq:
Superseded by https://github.com/m-labs/artiq/pull/2846

PR #2844, opened 2025-08-22
gui.dndwidgets: fix pyqt DropEvent minor version change

SimonRenblad:
Despite what the docs say it seems that `QDropEvent.position()` may return `QPoint` instead of `QPointF`, leading to an error when `toPoint()` is used. To get around this, use `x()` and `y()` which is present in both and cast to `int`. 

I have been unable to reproduce the original issue, even with `PyQt6.8.0`.


SimonRenblad:
> Yes that's it - the issue is reproduced on Linux if you select X11 session in gdm.

Good catch, I hadn't considered that.

PR #2846, opened 2025-08-26
Coremgmt: use config log_level

dly04:
(Based on PR #2841 log set_level: modify buffer log level only)
Try to use `artiq_coremgmt config` command and `log_level `/ `uart_log_level` keys to configure filter levels, both at runtime and store in flash.

Previous situation:
`artiq_coremgmt log set_level` / `artiq_coremgmt log set_uart_level` changes filter levels only at runtime, with error handling of invalid sub-command and invalid `LevelFilter`.
`artiq_coremgmt config write log_level` / `artiq_coremgmt config write uart_log_level` store keys in flash and only take effects on boot via `setup_log_levels`, with no error handling.

This commit:
`artiq_coremgmt config write log_level` / `artiq_coremgmt config write uart_log_level`  store keys in flash as well as change filter levels at run time, by adding `set_log_filter_level` and call it in `config_write`.

Concerns:
1. What to do with `artiq_coremgmt log set_level`, should do the same as `artiq_coremgmt config write log_level`, or to delete them?
2. Error handling: previously `artiq_coremgmt log` have error checking, while `artiq_coremgmt config write` do not have (like other normal config keys). Shall the merged `artiq_coremgmt config write log_level` have error check on sub-command, UTF8 and `LevelFilter`?

### DIFF
--- a/artiq/firmware/libproto_artiq/drtioaux_proto.rs
+++ b/artiq/firmware/libproto_artiq/drtioaux_proto.rs
@@ -136,8 +136,6 @@ pub enum Packet {
 
     CoreMgmtGetLogRequest { destination: u8, clear: bool },
     CoreMgmtClearLogRequest { destination: u8 },
-    CoreMgmtSetLogLevelRequest { destination: u8, log_level: u8 },
-    CoreMgmtSetUartLogLevelRequest { destination: u8, log_level: u8 },
     CoreMgmtConfigReadRequest { destination: u8, length: u16, key: [u8; MASTER_PAYLOAD_MAX_SIZE] },
     CoreMgmtConfigReadContinue { destination: u8 },
     CoreMgmtConfigWriteRequest { destination: u8, last: bool, length: u16, data: [u8; MASTER_PAYLOAD_MAX_SIZE] },
@@ -448,15 +446,7 @@ impl Packet {
             0xd1 => Packet::CoreMgmtClearLogRequest {
                 destination: reader.read_u8()?,
             },
-            0xd2 => Packet::CoreMgmtSetLogLevelRequest {
-                destination: reader.read_u8()?,
-                log_level: reader.read_u8()?,
-            },
-            0xd3 => Packet::CoreMgmtSetUartLogLevelRequest {
-                destination: reader.read_u8()?,
-                log_level: reader.read_u8()?,
-            },
-            0xd4 => {
+            0xd2 => {
                 let destination = reader.read_u8()?;
                 let length = reader.read_u16()?;
                 let mut key: [u8; MASTER_PAYLOAD_MAX_SIZE] = [0; MASTER_PAYLOAD_MAX_SIZE];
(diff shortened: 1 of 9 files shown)


dly04:
I wrapped UTF8 and invalid filter level into `mgmt_proto::Error<SchedError>::Utf8` and `mgmt_proto::Error<SchedError>::UnknownLogLevel` respectively, is it ok to do so?

artiq/firmware/runtime/mgmt.rs
```diff
@@ -100,16 +85,32 @@ mod local_coremgmt {
     }
 
     pub fn config_write(io: &Io, stream: &mut TcpStream, key: &String, value: &Vec<u8>, restart_idle: &Urc<Cell<bool>>) -> Result<(), Error<SchedError>> {
+        if key == "log_level" || key == "uart_log_level" {
```
MorganTL, line 88: shouldn't be put under `match config::write(key, value) {..` similar to `if key == "idle_kernel" {`?

artiq/firmware/libproto_artiq/drtioaux_proto.rs
```diff
@@ -448,14 +446,6 @@ impl Packet {
             0xd1 => Packet::CoreMgmtClearLogRequest {
                 destination: reader.read_u8()?,
             },
-            0xd2 => Packet::CoreMgmtSetLogLevelRequest {
```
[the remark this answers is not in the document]
dly04: Got it

artiq/firmware/runtime/mgmt.rs
SimonRenblad: Seems overly cautious, in my opinion. In the case of a `log_level` value that cannot be parsed, the user will want to override it anyways by calling the command again. After writing the value to config it would still error to let the user know something went wrong.

Spaqin: Invalid values have to be guarded against anyway (with e.g. ``incorrect log level, reverting to INFO`` warning, to let the user know something is wrong) because coremgmt isn't the only way to write a config. You can create one with ``artiq_mkfs`` and flash it anyway. In this case, client-side check in ``artiq_coremgmt`` should be enough.

dly04: Understand. for invalid level, there is a "connection close" error message in terminal and a "unknown log level" error message in log.

artiq/firmware/runtime/mgmt.rs
```diff
@@ -106,10 +91,26 @@ mod local_coremgmt {
                     io.until(|| !restart_idle.get())?;
                     restart_idle.set(true);
                 }
-                Reply::Success.write_to(stream)
+                if key == "log_level" || key == "uart_log_level" {
```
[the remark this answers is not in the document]
MorganTL: Nisi fallor you are talking about choosing between 3 match arms or 2 match arms. Just choose the one that will result in a cleaner / higher quality code. 
Also next time please use the rust match arm syntax if instead of tuple.

PR #1782, opened 2021-11-17
Extend SUServo to variable number of Urukul cards

pmldrmota:
# ARTIQ Pull Request

## Description of Changes

- Extend IIR filter to any number of ADC input channels and DDS output channels with minimal ressource usage. (The number of channels was previously restricted to a power of 2, and to be the same for inputs and outputs.)
- EEM support for one Sampler and a variable number of Urukuls.
- More efficient mapping of RTIO addresses used by SUServo.

### Related Issue
Closes #1339

## Type of Changes

|   | Type |
| ------------- | ------------- |
| ✓  | :sparkles: New feature |

### All Pull Requests

- [x] Use correct spelling and grammar.
- [x] Update [RELEASE_NOTES.rst](../RELEASE_NOTES.rst) if there are noteworthy changes, especially if there are changes to existing APIs.
- [ ] Close/update issues.
- [x] Check the copyright situation of your changes and sign off your patches (`git commit --signoff`, see [copyright](../CONTRIBUTING.rst#copyright-and-sign-off)).

### Code Changes

- [x] Run `flake8` to check code style (follow PEP-8 style). `flake8` has issues with parsing Migen/gateware code, ignore as necessary.
- [ ] Test your changes or have someone test them. Mention what was tested and how.
- [x] Add and check docstrings and comments
- [x] Check, test, and update the [unittests in /artiq/test/](../artiq/test/) or [gateware simulations in /artiq/gateware/test](../artiq/gateware/test)

### Documentation Changes

- [ ] Check, test, and update the documentation in [doc/](../doc/). Build documentation (`cd doc/manual/; make html`) to ensure no errors.

### Git Logistics

- [x] Split your contribution into logically separate changes (`git rebase --interactive`). Merge/squash/fixup commits that just fix or amend previous commits. Remove unintended changes & cleanup. See [tutorial](https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase).
- [x] Write short & meaningful commit messages. Review each commit for messages (`git show`).

### Licensing

See [copyright & licensing for more info](https://github.com/m-labs/artiq/blob/master/CONTRIBUTING.rst#copyright-and-sign-off).
ARTIQ files that do not contain a license header are copyrighted by M-Labs Limited and are licensed under LGPLv3+.

PR #2850, opened 2025-08-29
Extend SUServo to variable number of Urukul cards

occheung:
Superseding #1782. Other than rebasing to master, it has the following key changes:

- RTIO config address spaces changes are removed/postponed to coherent SU-Servo PR (WIP).
- Unify `urukulx_ports` key in the JSON schema into a `urukul_ports` entry. It is currently represented as a nested list.

All existing JSONs with specific `urukulx_ports` should be replaced.

### Tests
On a standalone Kasli v2.0:
- Targets with an SU-Servo that contains 1-4 urukuls can be compiled.
- All IIR channels can select an ADC input, which scales the DDS amplitude.
Passed all SU-Servo gateware unit test, after making appropriate fixes.

Docstring needs an update to distinguish ADC channels from DDS channels (would be `i_channels` and `o_channels` in #1782).

### DIFF
--- a/artiq/gateware/suservo/iir.py
+++ b/artiq/gateware/suservo/iir.py
@@ -7,18 +7,20 @@ logger = logging.getLogger(__name__)
 
 # all these are number of bits!
 IIRWidths = namedtuple("IIRWidths", [
-    "state",    # the signed x and y states of the IIR filter
-                # DSP A input, x state is one bit smaller
-                # due to AD pre-adder, y has full width (25)
-    "coeff",    # signed IIR filter coefficients a1, b0, b1 (18)
-    "accu",     # IIR accumulator width (48)
-    "adc",      # signed ADC data (16)
-    "word",     # "word" size to break up DDS profile data (16)
-    "asf",      # unsigned amplitude scale factor for DDS (14)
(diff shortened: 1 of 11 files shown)


artiq/gateware/suservo/iir.py
```diff
@@ -7,18 +7,20 @@
...
+                    # due to AD pre-adder, y has full width (25)
+    "coeff",        # signed IIR filter coefficients a1, b0, b1 (18)
+    "accu",         # IIR accumulator width (48)
+    "adc",          # signed ADC data (16)
+    "word",         # "word" size to break up DDS profile data (16)
+    "asf",          # unsigned amplitude scale factor for DDS (14)
+    "shift",        # fixed point scaling coefficient for a1, b0, b1 (log2!) (11)
+    "adc_channels", # number of input ADC channels (8)
+    "dds_channels", # number of output DDS channels (8)
+                    # the number of IIR channels is matched to this
```
dnadlinger, line 21: Does `IIRWidths` now mix both bit widths as well as the number of channels? Not only are they not relaly related; mixing numbers and log2/bit-width seems error prone. This is why we used `ADCParams` (which already has a `channels` field) and `DDSParams` in #1782.

Issue #2112, opened 2023-06-09
SU Servo initialisation problems together with stand-alone Urukul


MagicCreed:
# Bug Report

## One-Line Summary

Problems initialising stand-alone Urukul AD9910 with SU-Servo in the same kernel call.

### Steps to Reproduce
Ran the following block of code
```py
class DDSTest(EnvExperiment):
    def build(self):
        self.setattr_device("core")
        self.setattr_device("urukul2_ch0")
        self.setattr_device("urukul2_cpld")

        self.setattr_device("suservo0")
        self.setattr_device("urukul0_cpld")
        self.setattr_device("suservo0_ch0")

    @kernel
    def run(self):
        self.core.reset()

        ################### Standalone Urukul ###################
        self.urukul2_cpld.init()
        self.urukul2_ch0.init()
        self.urukul2_ch0.sw.on()
        self.urukul2_ch0.set_att(0.0)
        self.urukul2_ch0.set(frequency=230 * MHz, phase=0.0, amplitude=1.0)
        #####################################################

        ################### SUServo Urukul ###################
        self.suservo0.init()
        self.suservo0.set_config(enable=1)

        self.suservo0.set_pgia_mu(0, 1)
        self.urukul0_cpld.set_att(0, 0.0)

        self.suservo0_ch0.set_iir(profile=0, adc=0, kp=0.05, ki=-10000.0, g=0.0)
        self.suservo0_ch0.set_dds(profile=0, frequency=70 * MHz, offset=-0.15)
        self.suservo0_ch0.set(en_out=1, en_iir=1, profile=0)
        #####################################################
```
### Expected Behavior

Both Urukuls should initialise and output their respective set frequencies.

### Actual (undesired) Behavior
```
ERROR:worker(4178,DDS_test.py):root:While compiling <repository>\scratch\DDS_test.py
<synthesized>:1:1-1:62: error: host object has an attribute 'io_update' of type <instance artiq.coredevice.urukul._RegIOUpdate {
                __objectid__: numpy.int32
        }>, which is different from previously inferred type <instance artiq.coredevice.ttl.TTLOut {
                __objectid__: numpy.int32
        }> for the same attribute
`<artiq.coredevice.urukul.CPLD object at 0x000001814E69A260>`
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
<repository>\scratch\DDS_test.py:15:1: note: expanded from here
[...]

PR #2862, opened 2025-09-26
add suservo CPLD

DonovanWebb:
With introduction of support for `ProtoRev9` in `coredevice.urukul`, workaround for using both SUServo and normal Urukul devices in the same experiment no longer working due to polymorphism issues. 
Introducing a suservo specific `CPLD` and providing class attributes for both `ProtoRev8` and `ProtoRev9`  in `coredevice.suservo` allows for using both SUServo and normal Urukul devices again.

### DIFF
--- a/artiq/coredevice/suservo.py
+++ b/artiq/coredevice/suservo.py
@@ -564,3 +564,66 @@ class Channel:
             raise ValueError("Invalid SUServo y-value!")
         self.set_y_mu(profile, y_mu)
         return y_mu
+
+
+class ProtoRev8(urukul.ProtoRev8):
+    """
+    For use with suservo CPLD to get around polymorphic issues.
+    """
+
+
+class ProtoRev9(urukul.ProtoRev9):
+    """
+    For use with suservo CPLD to get around polymorphic issues.
+    """
+
+
+class CPLD(urukul.CPLD):
+    """
+    Needed to avoid polymorphic issues when using SUServo and regular Urukuls in the same experiment.
+    """
+    rb_len = 8
+
+    ProtoRev8 = ProtoRev8 # needed to avoid polymorphic issues 
+    ProtoRev9 = ProtoRev9 # needed to avoid polymorphic issues
+
+    def __init__(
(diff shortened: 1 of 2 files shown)


artiq/coredevice/suservo.py
```diff
@@ -564,3 +564,66 @@ def set_y(self, profile, y):
...
+    """
+    For use with suservo CPLD to get around polymorphic issues.
+    """
+
+
+class CPLD(urukul.CPLD):
+    """
+    Needed to avoid polymorphic issues when using SUServo and regular Urukuls in the same experiment.
+    """
+    rb_len = 8
```
dnadlinger, line 585: This seems related to phase-coherent suservo support and probably shouldn't be in this PR.

dnadlinger:
For context, see https://github.com/m-labs/artiq/issues/2112 and several similar issues/forum threrads. Our usual out-of-tree workaround doesn't easily translate, because the constructor now creates `ProtoRev` instances of a fixed type.

artiq/coredevice/suservo.py
dnadlinger: `nac3` isn't anything close to a drop-in replacement, and in any case this driver is for the legacy compiler only. This PR is just cleaning up after the recent `ProtoRev9` changes, which _were_ merged onto the non-nac3 branch, but broke a common use case (mixing SUServo and non-SUServo Urukuls; cf. various forum discussions).

dnadlinger: It's not a bug per-se, it's a long-standing limitation. It would be possible to change to "monomorphising" types per instance with some sort of merging of identical shapes, but every Python class corresponding to an ARTIQ Python type is a pretty central part of the current design. The SUServo driver does not take into account this facet of the design (at least not while keeping compatibility with the non-SUServo Urukul driver in the same kernel). This is a real-world issue, as evidenced not just by the Oxford group's, but several other labs' reports here or on the forum, so it obvioussly needs to be fixed.

PR #2866, opened 2025-10-01
entries: fix argument state persistence

### LINKED ISSUE #2865: investigate/fix dashboard state on msys2
https://forum.m-labs.hk/d/1010-remember-dashboard-argument-states

fsagbuya:
Remove redundant argument state initialization in `EntryTreeWidget` since argument states are already properly initialized in `ExperimentManager`. This will now load the argument state from the dashboard state file.

Closes #2865


dreens (approved):
I'm too late. Was just trying to create a pull request but I think you solved it better than I did anyway. Thank you!

dreens (commented):
I'm too late. Was just trying to create a pull request but I think you solved it better than I did anyway. Thank you!

SimonRenblad:
> since argument states are already properly initialized in ExperimentManager

`EntryArea` in gui/applets is also subclassed from `EntryTreeWidget` and does not have access to the `ExperimentManager`.

PR #2869, opened 2025-10-03
Add minimum support for Kasli v2.1

atseml:
Minimally support [Kasli v2.1](https://github.com/sinara-hw/Kasli/pull/112). Turns on EEM power regardless of fault for the time being.

Marked as WIP as this depends on the unmerged https://github.com/m-labs/migen/pull/315 and https://github.com/m-labs/misoc/pull/174.


atseml:
> The firmware behavior could be discussed later, but at least that part would be done.

I've kept it minimal as this needs to be quick, we are shipping out v2.1's now

PR #2875, opened 2025-10-10
Implement SU-Servo Synchronization

occheung:
Introduces the `synchronization` mode for SU-Servo, which performs `SYNC_IN` and `IO_UPDATE` delay calibration, similar to existing `AD9910` calibration methods. Only PROTO_REV 9 Urukuls are supported.

Calibration relies on reading back `SMP_ERR` and `AD9910` registers on SU-Servo mode. Readback is implemented via an additional MISO pin, implemented on the Urukul firmware patch for PROTO_REV 9. ([PR](https://git.m-labs.hk/M-Labs/urukul-pld/pulls/7))

Calibration values are stored in the EEPROM, and reprogrammed on initialization. `SYNC_IN` delays are written via the shared DDS interface via SPI; `IO_UPDATE` values are stored in the gateware by writing the SU-Servo config register.
|     Bits     |           Functionality           |
|:------------:|:---------------------------------:|
|       0      |               Enable              |
|       1      |    IO_UPDATE delay write latch    |
| 3*i+2: 3*i+5 | IO_UPDATE delay of the i-th Urukul |

Supersedes the synchronization part of #1467. An additional PR will be raised to address the remaining phase tracking mode.

### Tests
Compiled Kasli v2.0 gateware with a SU-Servo, which contains 1 or 2 Urukul in synchronization mode. Completed the calibration in `artiq_sinara_tester` respectively.

PR #2882, opened 2025-10-21
Implement phase tracking mode for SU-Servo

occheung:
This PR allows SU-Servo to adjust the given POW according to the AD9910 phase tracking mode.

### Enable the phase tracking mode
Phase tracking mode is configurable per channel. Invoke Channel.set(en_pt=en_pt) to enable/disable phase tracking mode.

### Calculation
Each DDS channel now has a copy of its phase accumulator tracked internally in the servo gateware, which is updated every IIR processing iteration.
The global time stamp is identical among all channels, since DDS updates all channels at once.
See AD9910 docs for the equation. See comments for the phase tracking DSP in iir.py for the pipeline implementation.

### Memory
The phase accumulator, previous FTW, and the fiducial timestamp of each DDS channel are stored in an additional memory block.

### Synthesis
With 5 Urukuls, vivado reports a RAMB36 usage for the aforementioned memory.
```
+-----------------------------------------+----------------------------------------+------------------------+---+---+------------------------+---+---+------------------+--------+--------+
|Module Name                              | RTL Object                             | PORT A (Depth x Width) | W | R | PORT B (Depth x Width) | W | R | Ports driving FF | RAMB18 | RAMB36 | 
+-----------------------------------------+----------------------------------------+------------------------+---+---+------------------------+---+---+------------------+--------+--------+
|top__GCB10                               | m_phase_reg                            | 680 x 32(READ_FIRST)   | W | R | 680 x 32(READ_FIRST)   | W | R | Port A and B     | 0      | 1      | 

```
1 more DSP is inferred by vivado.

### Reset behavior
After setting `enable=0` to the servo, the user should wait until the pipelines are drained before re-enabling the servo.
The internal time stamp accumulator resets once disabled. The tracked phase accumulator and FTW should be cleared manually by the user, since both values are stored in memories.

### Breaking Changes
To facilitate phase tracking DSP pipelining, the values in config memory are rearranged.

### Logistics
I suggest merging #2875 first.
Together with #2875, supersedes #1467.

### DIFF
--- a/artiq/gateware/rtio/phy/servo.py
+++ b/artiq/gateware/rtio/phy/servo.py
@@ -13,12 +13,12 @@ class RTServoCtrl(Module):
 
         self.comb += [
                 ctrl.stb.eq(self.rtlink.o.stb),
-                self.rtlink.o.busy.eq(0),
+                self.rtlink.o.busy.eq(0)
         ]
         self.sync.rio_phy += [
                 If(self.rtlink.o.stb,
                     Cat(ctrl.en_out, ctrl.en_iir, ctrl.en_pt, ctrl.profile).eq(
-                            self.rtlink.o.data),
+                            self.rtlink.o.data)
                 )
         ]
 
--- a/artiq/gateware/suservo/iir.py
+++ b/artiq/gateware/suservo/iir.py
@@ -96,7 +96,7 @@ class DSP(Module):
                     # but that won't infer P reg, so we just clear
                     # and round down
                     p.eq(0),
-                ),
+                )
         ]
 
         self.comb += self.output.eq(p[w.shift:])
@@ -755,7 +755,7 @@ class IIR(Module):
         or low part of the memory location.
         """
         w = self.widths
-        addr = "ftw1 b1 pow cfg offset a1 ftw0 b0".split().index(coeff)
+        addr = "pow b1 ftw0 cfg offset a1 ftw1 b0".split().index(coeff)
         coeff_addr = ((channel << w.profile + 2) | (profile << 2) |
                 (addr >> 1))
         mask = (1 << w.coeff) - 1
@@ -824,6 +824,37 @@ class IIR(Module):
             raise ValueError("no such state", coeff)
         return signed(val, w.state)
 
+    def set_fiducial_timestamp(self, channel, profile, val):
+        w = self.widths
+        yield self.m_phase[profile | (channel << w.profile)].eq(val)
+
+    def get_fiducial_timestamp(self, channel, profile):
+        w = self.widths
+        val = yield self.m_phase[profile | (channel << w.profile)]
+        return val
+
+    def set_prev_ftw(self, channel, val):
+        w = self.widths
+        yield self.m_phase[(channel << 1) +
+                    ((1 << w.profile) * self.o_channels)].eq(val)
+
+    def get_prev_ftw(self, channel):
+        w = self.widths
+        val = yield self.m_phase[(channel << 1) +
+                    ((1 << w.profile) * self.o_channels)]
+        return val
+
+    def set_phase_accumulator(self, channel, val):
+        w = self.widths
+        yield self.m_phase[1 | (channel << 1) +
+                    ((1 << w.profile) * self.o_channels)].eq(val)
+
+    def get_phase_accumulator(self, channel):
+        w = self.widths
+        val = yield self.m_phase[1 | (channel << 1) +
+                    ((1 << w.profile) * self.o_channels)]
+        return val
+
     def fast_iter(self):
         """Perform a single processing iteration."""
         assert (yield self.done)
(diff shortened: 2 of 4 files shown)


occheung:
Repeated the previous test with fiducial time stamp = 8*ns for all profiles on channel 0 (C1), 0 otherwise. The skew before and after the change of frequency are both -8.00 ns (`Skew(C1, C2)`).

It is set before enabling the servo.
```py
        # Set fiducial timestamp of channel[0]
        self.suschannels[channels[0]].set_reference_time(8, 0)    # profile 0
        self.suschannels[channels[0]].set_reference_time(8, 1)    # profile 1
```
Time stamp is defined as 0 when the servo is enabled. I may interchange the parameters later.

SDS2504X Plus_JPG_1

Behavior of the internal time counter and fiducial time stamps after disabling the servo is still WIP.

artiq/coredevice/suservo.py
```diff
@@ -281,10 +283,92 @@ def set(self, en_out, en_iir=0, profile=0):
...
+        Fiducial time stamp refers to the variable `T` in phase tracking mode
+        equation. See :meth:`artiq.coredevice.ad9910.AD9910.set_phase_mode`.
+        Fiducial time stamp is defined as 0 when the DDSes are updated for
+        the first time since enabling the servo.
+
+        With en_pt=1 (see :meth:`set`), the DDS output phase of this channel
+        will refer to this fiducial time stamp.
+
+        :param profile: Profile number (0-31)
+        :param fiducial_ts: Fiducial time stamp
```
dnadlinger: This should probably be `fiducial_mu` (or in any case, end in `_mu`) following the usual core device driver naming convention to disambiguate time in seconds and time in machine units.

artiq/coredevice/suservo.py
```diff
@@ -281,10 +283,92 @@ def set(self, en_out, en_iir=0, profile=0):
...
+        servo.
+        """
+        addr = self.servo.phase_sel | ((4 * len(self.servo.cplds) * 32 + (self.servo_channel << 1) | 1) << 1)
+        self.servo.write(addr, 0)
+        self.servo.write(addr + 1, 0)
+
+    @kernel
+    def get_tracked_phase_accumulator(self):
+        """Read the tracked phase accumulator in the servo.
+
```
dnadlinger: Could mention that this is only sensible to use when the frequency has been set to zero, as otherwise there will be word tearing across the two-part read.

dnadlinger:
> What about non-SUServo?

For the phase tracking modes there, the reference time naturally takes on a meaning as relative to the start of the DMA sequence. This is because the absolute/tracking modes are implemented via resetting the DDS accumulator, so the result is naturally independent of absolute time/DDS history. The challenge on non-SUServo channels in DMA would in fact be the opposite, i.e. setting an absolute phase reference time (though then in practice, setting things up before the DMA playback and not clearing the accumulators in the DMA sequence often works).

artiq/coredevice/suservo.py
```diff
@@ -281,10 +284,134 @@ def set(self, en_out, en_iir=0, profile=0):
...
+        self.core.break_realtime()
+        hi = self.servo.read(addr + 1)
+        return (hi << 16) | (lo & 0xffff)
+
+    @kernel
+    def clear_tracked_phase_accumulator(self):
+        """Clears the tracked phase accumulator of this channel.
+        The tracked phase accumulator is an internal register that tracks the
+        corresponding register of the corresponding DDS. To properly track the
+        accumulator of the DDS, both accumulator should be cleared before
```
architeuthidae: typo: accumulators

artiq/coredevice/suservo.py
```diff
@@ -281,10 +284,134 @@ def set(self, en_out, en_iir=0, profile=0):
...
+
+        Fiducial time stamp refers to the variable `T` in phase tracking mode
+        equation. See :meth:`artiq.coredevice.ad9910.AD9910.set_phase_mode`.
+        Fiducial time stamp is defined as 0 immediately after the servo has
+        been enabled.
+
+        With en_pt=1 (see :meth:`set`), the DDS output phase of this channel
+        will refer to this fiducial time stamp.
+
+        This method advances the timeline by two coarse RTIO cycles.
```
architeuthidae: typo: cycle

artiq/coredevice/suservo.py
```diff
@@ -281,10 +284,134 @@ def set(self, en_out, en_iir=0, profile=0):
...
+
+        With en_pt=1 (see :meth:`set`), the DDS output phase of this channel
+        will refer to this copied fiducial time stamp.
+
+        The coarse part of the time stamp is sourced from the internal time
+        stamp accumulator of the servo. The accumulator only updates every
+        coarse RTIO cycle. The fine part of the time stamp is provided by the
+        corresponding part of the timeline cursor.
+
+        This method advances the timeline by one coarse RTIO cycles.
```
architeuthidae: typo: cycle

PR #2885, opened 2025-10-23
artiq_coremgmt: support flash boot.bin directly

SimonRenblad:
If `artiq_coremgmt flash PATH` is a file, try to flash it directly (for Kasli-SoC). Otherwise search the directory like before.

The original method is kept, as I think being able to `artiq_coremgmt flash result` is still quite useful for Kasli-SoC (aside from the `result` only containing `boot.bin`).


sbourdeauducq:
I would put the binary discovery code into a separate file and share it between artiq_flash and artiq_coremgmt.

PR #2894, opened 2025-10-24
Songbird: make DDS count configurable

Spaqin:
This PR is not just making the tone count configurable on Songbird; it goes a little beyond to fully facilitate this change.

Of course, first and foremost there's a new parameter for ``efc`` - ``--dds-count``. Default is still 4. JSON files (``artiq_ddb_template``) and ``artiq_sinara_tester`` also support arbitrary numbers.

Experimentally, EFC v1.0 will compile with 9 tones, v1.1 with a bigger FPGA can go up to 14. Over that, the LUT memory becomes the bottleneck.

To achieve that, a summer cascade (that adds up pairs of values in each stage) has been implemented, so latency increases with higher tone count. Overall I don't think the latency matters in this case, but it is a drawback.

Other things that have been done:
- minor optimizations to maybe make more space
- LUT shared between ``cossingen`` should be done in pairs (Vivado took care of it regardless)

### DIFF
--- a/artiq/gateware/songbird.py
+++ b/artiq/gateware/songbird.py
@@ -215,9 +215,7 @@ class SongbirdDDSModule(Module, AutoCSR):
 
             # count down from 2**shift-1 to 0
             If(self.shift_counter == 0,
-                Case(self.shift,
-                    { i: self.shift_counter.eq((1 << i) - 1) for i in range(2**len(self.shift)) } | { "default": self.shift_counter.eq(0) }
-                )
+                self.shift_counter.eq((1 << (self.shift + 1)) - 1)
             ).Else(
                 self.shift_counter.eq(self.shift_counter - 1)
             ),


artiq/frontend/artiq_sinara_tester.py
```diff
@@ -983,7 +994,7 @@ def test_songbirds(self):
             print("...done")
             print("{} output active. Frequencies: {} MHz.".format(
                 card_name,
-                ", ".join([str(10*i + card_n) for i in range(1, 5)]))
+                ", ".join([str(10*i + card_n) for i in range(1, len(card_dev["dds"])+1)]))
```
MorganTL: NIT: move the +1 inside the for loop `[str(10*(i + 1) + card_n) for i in range(len(card_dev["dds"]))]`

MorganTL (commented):
What's the limiting factor of the number of DDS other than fpga resources? Address length limit the DDS number to 16 or 20?

PR #2901, opened 2025-10-28
eem: Clarify Urukul SYNC_IN and SYNC_OUT

atseml:
Differentiate Urukul's DDS_RESET/SYNC_IN on EEM0[7] from SYNC_IN on EEM1[1] by renaming the latter to SYNC_OUT in the gateware. The now SYNC_OUT signal has always been an Urukul output. The converse also applies to SYNC_IN, so apply a correction to a related comment as well.
SYNC_OUT is currently unused.

See https://git.m-labs.hk/M-Labs/urukul-pld/commit/a6da8916cd52fecc4bc1976722bee0dd34e60b1a.


occheung (approved):
LGTM.

PR #2921, opened 2025-12-05
efc: Avoid repeated `afe_adc_spi`s in Shuttler IO

atseml:
Having repeated `afe_adc_spi` entries for disparate Shuttler AFE HW revs caused the ADC SPI communication of Shuttler AFE v1.3 to be incoherent, and it seemed to behave the same regardless of the specified Shuttler AFE HW rev.

### DIFF
--- a/artiq/gateware/targets/efc.py
+++ b/artiq/gateware/targets/efc.py
@@ -71,7 +71,7 @@ ltc2000_spi = [
     )
 ]
 
-shuttler_io = [
+shuttler_io_common = [
     ('dac_spi', 0,
         Subsignal('clk', Pins('fmc0:HB16_N')),
         Subsignal('mosi', Pins('fmc0:HB06_CC_N')),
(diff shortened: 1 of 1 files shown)


sbourdeauducq:
I see another bug: the AFE I/O is appended to the global list every time ``EfcShuttler`` is instantiated, so the global list accumulates them.
https://github.com/m-labs/artiq/blob/4aec72d1c1f14dfe4c4309c8af841b66a9af8e5c/artiq/gateware/targets/efc.py#L277
I suggest:
(1) capitalizing the names of global variables to clearly mark them as such
(2) either:
```python
shuttler_io = SHUTTLER_IO + afe_adc_io
```
or, simply:
```python
platform.add_extension(afe_adc_io)
```

This bug is silent right now since the class is only instantiated once per Python process, but it's certainly not best practices and could cause confusion in future use cases.

atseml:
> (2) either:
> 
> ```
> shuttler_io = SHUTTLER_IO + afe_adc_io
> ```
> 
> or, simply:
> 
> ```
> platform.add_extension(afe_adc_io)
> ```

These will require wrapping `afe_adc_io` in a list to work. I went one step further though and just defined `shuttler_io_afe_pre_v1_3` and `shuttler_io_afe_v1_3` (deduplicated of course with `shuttler_io` -> `shuttler_io_common`) since I believe that's cleaner.

For (1) all Python global constants will be capitalized in a separate PR since that's going to be a repo-wide change.

atseml:
Force-pushes re: metadata like commit message changes and authorship.

PR #2922, opened 2025-12-09
update interval,timestamp to dashboard on checkbox

hubuke:
Add uniform_interval option to artiq dashboard
When checked, "Add channels" dialogue box would have timestamp and interval trace. These two trace can be added to the waveform viewer.

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -41,9 +41,9 @@ ARTIQ-9 (Unreleased)
    - State files are now automatically backed up upon successful loading.
    - Experiments can now use ``restart_applet`` CCB command to restart applets.
    - Interactive argument widgets now remember their layout between runs.
-   - Waveform Viewer now supports the ``uniform_interval`` option, equivalent to the ``-u`` option
-     in ``artiq_coreanalyzer``. In this mode, each event advances the vcd trace by one step, 
-     with additional ``interval`` and ``timestramp`` traces indicating the actual time. 
+   - Waveform Viewer now has a ``uniform_interval`` checkbox, equivalent to the ``-u`` option
+     in ``artiq_coreanalyzer``. In this mode, each event advances the RTIO waveform trace by one step, 
+     with additional ``interval`` and ``timestamp`` traces indicating the actual time. 
 * ``artiq_session`` has been revamped. Output now indicates the originating process. Dashboard
   is started after the controller manager. All processes are cleanly terminated after dashboard
   is closed.
--- a/artiq/dashboard/waveform.py
+++ b/artiq/dashboard/waveform.py
@@ -26,6 +26,7 @@ logger = logging.getLogger(__name__)
 
 WAVEFORM_MIN_HEIGHT = 50
 WAVEFORM_MAX_HEIGHT = 200
+PREFIXES_STR = "pnumkMG"
 
 
 class ProxyClient():
@@ -655,7 +656,7 @@ class _CursorTimeControl(QtWidgets.QLineEdit):
 
     def _setValueFromText(self, text):
         try:
-            if not text.strip().endswith(("event", "events", "s")):
+            if text.strip().endswith(tuple(PREFIXES_STR)):
                 text += self._suffix
             self._value = pg.siEval(text) * (1e12 / self._timescale)
         except:


hubuke:
unit="tick"?

SimonRenblad (changes requested):

SimonRenblad:
> unit="tick"?

Better, but may be confused with CPU ticks I guess.

artiq/dashboard/waveform.py
```diff
@@ -925,3 +959,21 @@ def notify_ddb(self, mod):
     async def stop(self):
         if self.proxy_client is not None:
             await self.proxy_client.close()
+
+    def _update_uniform_interval(self):
+        self._uniform_interval = self._uniform_interval_checkbox.isChecked()
+        channel_list = comm_analyzer.get_channel_list(self._ddb)
```
SimonRenblad: some code duplication here with `_process_ddb` that could be factored out

artiq/dashboard/waveform.py
```diff
@@ -633,10 +638,14 @@ def _onTextChange(self, text):
 
     def setDisplayValue(self, value):
         self._value = value
-        self._text = pg.siFormat(value * 1e-12 * self._timescale,
-                                 suffix="s",
-                                 allowUnicode=False,
```
SimonRenblad: Why did you remove this?

SimonRenblad (changes requested):

artiq/dashboard/waveform.py
hubuke: I am not sure what is the purpose for this?
Allowing entries for example: 100u or 300m seems nice?

hubuke: Added this back. I thought this was causing issue with siEval resolving "100u". But actually only nunbers with units will be properly resolved. So I add "s" or "event" as unit if not inputted.

artiq/dashboard/waveform.py
```diff
@@ -925,3 +961,22 @@ def notify_ddb(self, mod):
     async def stop(self):
         if self.proxy_client is not None:
             await self.proxy_client.close()
+    
+    def get_uniform_interval(self):
+        return self._uniform_interval
+    
+    def get_channel_list(self):
```
SimonRenblad: only used internally, so `_` prefix?

artiq/dashboard/waveform.py
```diff
@@ -621,26 +628,35 @@ def __init__(self, parent):
         self._text = ""
         self._value = 0
         self._timescale = 1
+        self._suffix = "s"
+        self._uniform_interval = parent.get_uniform_interval() if (parent is not None and isinstance(parent.get_uniform_interval(), bool)) else False
```
SimonRenblad: When I said to use a setter, I meant to define a setter in the `_CursorTimeControl` and call it as necessary from the parent widget. Generally you want to avoid assuming members of the parent widget if it isn't necessary. 

Easiest would be to:
1. make `uniform_interval` public in the `_CursorTimeControl` (rm leading underscore)
2. set it to `False` in the constructor
3. call `self._cursor_control.update_on_uniform_interval` in `WaveformDock`.

artiq/dashboard/waveform.py
```diff
@@ -925,3 +961,22 @@ def notify_ddb(self, mod):
     async def stop(self):
         if self.proxy_client is not None:
             await self.proxy_client.close()
+    
+    def get_uniform_interval(self):
```
SimonRenblad: see comment above

artiq/dashboard/waveform.py
```diff
@@ -621,26 +628,35 @@ def __init__(self, parent):
...
+        self._suffix = "s"
+        self._uniform_interval = parent.get_uniform_interval() if (parent is not None and isinstance(parent.get_uniform_interval(), bool)) else False
         self.setDisplayValue(0)
         self.textChanged.connect(self._onTextChange)
         self.returnPressed.connect(self._onReturnPress)
 
     def setTimescale(self, timescale):
         self._timescale = timescale
 
+    def update_on_uniform_interval(self, uniform_interval):
```
SimonRenblad: nit: `updateUniformInterval`? More consistent with other methods in class.

hubuke: I was deciding between PEP8 or backward compatible.

SimonRenblad: not really a backward compat issue? but consistency within a class is nice when possible

artiq/dashboard/waveform.py
hubuke: I assume you mean keep `uniform_interval` in `_CursorTimeControl` private but use a setter in `_CursorTimeControl` to set it? 
I don't understand what is the purpose of making that public if we have a setter?

sbourdeauducq:
Needs update of release notes.

RELEASE_NOTES.rst
```diff
@@ -41,6 +41,9 @@ ARTIQ-9 (Unreleased)
    - State files are now automatically backed up upon successful loading.
    - Experiments can now use ``restart_applet`` CCB command to restart applets.
    - Interactive argument widgets now remember their layout between runs.
+   - Waveform Viewer now supports the ``uniform_interval`` option, equivalent to the ``-u`` option
+     in ``artiq_coreanalyzer``. In this mode, each event advances the vcd trace by one step,
```
sbourdeauducq: What "vcd trace"?
The dashboard waveform viewer doesn't use VCD. Please be accurate.

RELEASE_NOTES.rst
```diff
@@ -41,6 +41,9 @@ ARTIQ-9 (Unreleased)
    - State files are now automatically backed up upon successful loading.
    - Experiments can now use ``restart_applet`` CCB command to restart applets.
    - Interactive argument widgets now remember their layout between runs.
+   - Waveform Viewer now supports the ``uniform_interval`` option, equivalent to the ``-u`` option
```
sbourdeauducq: "``uniform_interval`` option"?
In a GUI?

artiq/dashboard/waveform.py
```diff
@@ -621,26 +628,35 @@ def __init__(self, parent):
...
         self._text = pg.siFormat(value * 1e-12 * self._timescale,
-                                 suffix="s",
+                                 suffix=self._suffix,
                                  allowUnicode=False,
                                  precision=15)
         self.setText(self._text)
 
     def _setValueFromText(self, text):
         try:
+            if not text.strip().endswith(("event", "events", "s")):
```
sbourdeauducq: Looks fishy to me. Isn't there a better way of doing this?

RELEASE_NOTES.rst
```diff
@@ -41,6 +41,9 @@ ARTIQ-9 (Unreleased)
...
+     in ``artiq_coreanalyzer``. In this mode, each event advances the vcd trace by one step, 
+     with additional ``interval`` and ``timestramp`` traces indicating the actual time.
```
sbourdeauducq: timestramp

Issue #30, opened 2015-06-18
RTIO exception parameters are broken


sbourdeauducq:
Stuck at 0, e.g.
artiq.coredevice.runtime_exceptions.RTIOUnderflow: at 0 ps on channel 0, violation 0 ps

sbourdeauducq:
The bug happens when exceptions that are raised with parameters by the runtime are caught by a ``try/finally`` construct in the compiled code. The ``finally`` block re-raises the exception through the runtime's ``exception_raise``, which clears the parameters. Fixing this properly needs making the compiler aware of exception parameters - not clearing the parameters when the compiled code raises an exception is not enough, e.g.

<pre>
try:
    ...
except SomeException:
    # This does need to clear the parameters
    # (and change the ID of the exception re-raised by finally) 
    raise SomeOtherException
finally:
    ...
</pre>

This bug is manifesting itself all the time at the moment, as the whole kernel code is encapsulated in a ``try/finally`` construct to handle RTIO handovers.

whitequark:
new-py2llvm does not exhibit this bug.

sbourdeauducq:
Please close when new-py2llvm is usable with the rest of ARTIQ.

whitequark:
Closing since new compiler was merged.

Issue #2775, opened 2025-06-09
test Phaser ADC in artiq_sinara_tester


sbourdeauducq:

hubuke:
Verified ADC works after adding readout registers in the phaser gateware. Readout value off potentially because of [Fix ADC Missed 1 Bit Error #28 in Quartiq/phaser](https://github.com/quartiq/phaser/pull/28/files)

PR #2930, opened 2025-12-24
sinara_tester: add phaser adc test

hubuke:
Add test for phaser ADC channels. 

This test depends on exposed ADC register, [phaser gateware pr #30](https://github.com/quartiq/phaser/pull/30).

### DIFF
--- a/artiq/frontend/artiq_sinara_tester.py
+++ b/artiq/frontend/artiq_sinara_tester.py
@@ -8,7 +8,7 @@ import sys
 
 from artiq.experiment import *
 from artiq.coredevice.ad9910 import AD9910, SyncDataEeprom
-from artiq.coredevice.phaser import PHASER_GW_BASE, PHASER_GW_MIQRO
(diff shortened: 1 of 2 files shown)


artiq/frontend/artiq_sinara_tester.py
```diff
@@ -709,6 +709,46 @@ def phaser_led_wave(self, phasers):
...
+        phaser.init()
+        phaser.write8(PHASER_ADDR_ADC_CFG, 0b0000) # set both channel G = 1
+        delay(1*ms)
+        adc_data = phaser.read_adc_raw(ch)
+        delay(1*ms)
+        voltage = phaser.convert_adc_voltage(adc_data)
+        cb(voltage)
+
+    @kernel
+    def setup_phaser_servo(self, phaser, duc):
```
Spaqin: That's unused (and as mentioned in #2775 the servo issue is not yet resolved); don't think that belongs in this PR then

Issue #2746, opened 2025-05-06
device_db entry keys ought to be sanitised


dnadlinger:
Currently, arbitrary strings are accepted as device_db keys (device names), including e.g. multi-line strings. This should be suitably restricted.

The ARTIQ design heavily suggests using Python identifiers as device names (`setattr_device()`, etc.), but this is not in fact enforced anywhere. This e.g. breaks log forwarding via the controller manager, where the device name is used as the log source, and line breaks and colons serve as special delimiters (but are not escaped).

It would probably be fine to require the devices names to be valid Python identifiers. Alternatively – to avoid unnecessarily breaking backwards compatibility in case somebody does use non-identifier device names – perhaps we could at least restrict them to be single-line and not contain `:`?

PR #2935, opened 2025-12-31
device_db validation: key must be python identifier

hubuke:
Issue: #2746 

In `artiq/master/databases.py`, force use python identifier for device_db key entry every time load the device_db file. 
Raise error otherwise.

Break backward compatibility for users with non python identifier device_db key entries.

DOCs: add device_db validation may cause backward incompatible to release note.

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -87,6 +87,7 @@ Breaking changes:
 
 * Experimental features have been removed.
 * SU-Servo coefficient memory have been remapped. Users should re-interpret the record written by Channel.get_profile_mu().
+* ``device_db.py`` now requires all device DB keys to be valid Python identifiers ``[A-Za-z0-9_]``; no special characters (except underscore) are allowed.
 
 ARTIQ-8
 -------


artiq/master/databases.py
```diff
@@ -17,9 +17,19 @@ def device_db_from_file(filename):
...
     # (raise KeyError and not AttributeError if device_db is missing)
-    return mod.__dict__["device_db"]
-
-
+    device_db = mod.__dict__["device_db"]
+    validate_device_db(device_db)
+    return device_db
+
+def validate_device_db(device_db_dict):
+    for key in device_db_dict:
```
sbourdeauducq: Consistent naming - drop the ``_dict``

sbourdeauducq: And another nit, the codebase uses ``.keys()`` explicitly elsewhere.

sbourdeauducq:
Needs entry in release notes.

dnadlinger:
> Not sure if this is actually needed.

Some sort of fix is necessary for logging not to break if a key contains newlines or colons. This could of course also be done through appropriate escaping on the logging side, but it seems to be a corner case that may not be worth supporting.

Thinking about it, perhaps the most likely sensible use for such names may be a convention like `my_device:ch0`? This still wouldn't work with `setattr_device`, but it could conceivably be in use out there.

(What prompted the report is somebody trying to comment out part of the device db using multi-line strings (triple quotes), which then was pre-pended to the name of the next entry, which happened to be a controller, hence breaking the forwarded log.)

PR #2940, opened 2026-01-09
coremgmt: improve config ux

SimonRenblad:
1. Add informative error messages for empty `artiq_coremgmt config read/write/remove`.
2. Add examples to `artiq_coremgmt config --help` to illustrate usage of `-f/-s` options.

### DIFF
--- a/artiq/frontend/artiq_coremgmt.py
+++ b/artiq/frontend/artiq_coremgmt.py
@@ -43,7 +43,17 @@ def get_argparser():
 
     # configuration
     t_config = tools.add_parser("config",
-                                help="read and change core device configuration")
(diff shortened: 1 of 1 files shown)


artiq/frontend/artiq_coremgmt.py
```diff
@@ -142,6 +155,8 @@ def main():
                 with open(filename, "wb") as fi:
                     fi.write(value)
         if args.action == "write":
+            if len(args.string) == 0 and len(args.file == 0):
```
sbourdeauducq: ``len(args.file == 0)`` ?

Also I think you can just write ``not list`` to check for emptiness.

Issue #2924, opened 2025-12-10
llvmlite 0.46 breaks optimization


Doomsel:
When running the artiq compiler with llvmlite==0.46 (instead of the 0.45 used in the current nix environment), the python process crashes with the exception below.
To reproduce run an arbitrary experiment with `artiq_run`.  

```
Invalid optimization level for size level 2
UNREACHABLE executed at /tmp/pip-install-34zfw1k3/llvmlite_3325acc1594f407ebffc2ad8ceb9ded2/ffi/newpassmanagers.cpp:286!
Fatal Python error: Aborted

Current thread 0x00007f891f0f5b28 (most recent call first):
  File "/usr/lib/python3.12/site-packages/llvmlite/binding/ffi.py", line 212 in __call__
  File "/usr/lib/python3.12/site-packages/llvmlite/binding/newpassmanagers.py", line 585 in getModulePassManager
  File "/git/artiq/artiq/compiler/targets.py", line 120 in optimize
  File "/git/artiq/artiq/compiler/targets.py", line 151 in compile
  File "/git/artiq/artiq/compiler/targets.py", line 188 in compile_and_link
  File "/git/artiq/artiq/coredevice/core.py", line 150 in compile
  File "/git/artiq/artiq/coredevice/core.py", line 175 in run
  File "/git/artiq/artiq/language/core.py", line 54 in run_on_core
```

If i change the size optimization level [here](https://github.com/m-labs/artiq/blob/c2f17af93abae2b72cc6b5a12d9fcd2d784cdfc7/artiq/compiler/targets.py#L111) to `size_level=0` the code compiles without issue. The same goes for when changing the [speed level ](https://github.com/m-labs/artiq/blob/c2f17af93abae2b72cc6b5a12d9fcd2d784cdfc7/artiq/compiler/targets.py#L117) to `2` (or commenting out the line). 

If one sets the speed level directly in line 111:
```
pto = llvm.create_pipeline_tuning_options(speed_level=1, size_level=1)
```
It directly complains:
```
self = <llvmlite.binding.newpassmanagers.PipelineTuningOptions object at 0x7f3373ff8d70>, value = 1

    @size_level.setter
    def size_level(self, value):
        if not 0 <= value <= 2:
            raise ValueError("Optimization level for size should be 0, 1, or 2")
        if value != 0 and self.speed_level != 2:
>           raise ValueError(
                "Optimization for size should be encoded with speed level == 2")
[...]

PR #2942, opened 2026-01-12
compiler: remove llvmlite workaround for size optimization level

fsagbuya:
We don't need the [version check](https://github.com/m-labs/artiq/pull/2941) and the `llvmlite v0.45` size optimization level workaround anymore since the upstream stable nixpkgs now uses `v0.46`. See: https://github.com/NixOS/nixpkgs/commit/4d10fadf1372571963f8eda62fd40fbcec679422.

Tested with `nix build .#artiq`.

Related issue: https://github.com/m-labs/artiq/issues/2924.

### DIFF
--- a/artiq/compiler/targets.py
+++ b/artiq/compiler/targets.py
@@ -1,6 +1,6 @@
 import os, sys, tempfile, subprocess, io
 from artiq.compiler import types, ir
-from llvmlite import ir as ll, binding as llvm, __version__ as llvmlite_version
+from llvmlite import ir as ll, binding as llvm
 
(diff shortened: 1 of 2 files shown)


sbourdeauducq:
I think it doesn't hurt to keep a simple workaround like that for a while, and it would support people using other installation methods?

fsagbuya:
In `MSYS2`, the upstream llvmlite version is already `0.46` as well. This is just a nit to keep the code simpler and streamlined. Yes, it would not hurt to keep it for a while.

PR #2947, opened 2026-01-15
sinara_tester: set Almazny test frequency to at least 7GHz

hubuke:
Previous 4 GHz test was due to lack of test equipment. Might cause confusion.

Almazny theoretical minimum frequency is 6800 MHz, see: [AD5356 datasheet page 4](https://www.analog.com/media/en/technical-documentation/data-sheets/adf5356.pdf)

image

Slow down attenuation update for better capture on TinySA.

### DIFF
--- a/artiq/frontend/artiq_sinara_tester.py
+++ b/artiq/frontend/artiq_sinara_tester.py
@@ -501,7 +501,7 @@ class SinaraTester(EnvExperiment):
         print("Frequencies:")
         for card_n, channels in enumerate(chunker(self.mirnies, 4)):
             for channel_n, (channel_name, channel_dev) in enumerate(channels):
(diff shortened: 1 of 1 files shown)


Spaqin:
How does it look on tinySA?

MorganTL:
The commit message didn't mention "Slow down attenuation update for better capture on TinySA.". Split the commit into two and update the title

MorganTL:
Also what the RBW setting did you use? A larger RBW should speed up the capture speed. A smaller span helps too.

hubuke:
The default setting on TinySA takes about 1.2 per scan at 500MHz span, 850kHz RBW, 450 sweep points.

The fastest setting that I think are clear enough is about 0.8 s per scan at 350M, 850k RBW, 100 sweep points.

